PR #9, opened 2019-02-16
add_electricity: add CCGT to extendable generators

fneum:
Adds the option to include extendable CCGT generators in addition to OCGT generators.

### DIFF
--- a/scripts/add_electricity.py
+++ b/scripts/add_electricity.py
@@ -339,6 +339,7 @@ def attach_extendable_generators(n, costs, ppl):
         if suptech == 'OCGT':
             ocgt = ppl.loc[ppl.Fueltype.isin(('OCGT', 'CCGT'))].groupby('bus', as_index=False).first()
             n.madd('Generator', ocgt.index,
+                   suffix=' OCGT'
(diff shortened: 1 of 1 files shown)


scripts/add_electricity.py
```diff
@@ -339,16 +339,29 @@ def attach_extendable_generators(n, costs, ppl):
...
                    bus=ocgt['bus'],
                    carrier=tech,
                    p_nom_extendable=True,
                    p_nom=0.,
                    capital_cost=costs.at['OCGT', 'capital_cost'],
                    marginal_cost=costs.at['OCGT', 'marginal_cost'],
                    efficiency=costs.at['OCGT', 'efficiency'])
+
+        elif suptech == 'CCGT':
+            ccgt = ppl.loc[ppl.Fueltype.isin(('OCGT', 'CCGT'))].groupby('bus', as_index=False).first()
```
coroa: At some point of adding more fueltypes we will have to decide what to do with the heuristic of extending only where there have been capacities previously. I think for now its fine to be able to add OCGT and/or CCGT capacities where there were gas plants previously.

PR #14, opened 2019-02-18
add_electricity.py: Remove capital costs for existing conventional capacity

Heronimonimo:
All the existing conventional capacity has sunk capital costs. Therefore these should have capital costs of 0.

### DIFF
--- a/scripts/add_electricity.py
+++ b/scripts/add_electricity.py
@@ -227,7 +227,7 @@ def attach_conventional_generators(n, costs, ppl):
                p_nom=p['p_nom'],
                efficiency=costs.at[tech, 'efficiency'],
                marginal_cost=costs.at[tech, 'marginal_cost'],
-               capital_cost=costs.at[tech, 'capital_cost'])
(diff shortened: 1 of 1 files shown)


coroa:
`capital_cost` in PyPSA actually only refers to the cost for extending the capacity (refer to https://pypsa.org/doc/components.html#generator), so it's fine here as it was. But, we don't apply this convention in `make_summary` and `plot_network`. It would be more correct to fix those, instead.

Don't want to merge this change as-is.

We can apply it as a temporary measure, after adding a warning comment and opening a new issue to fix `make_summary` and `plot_network`!

Heronimonimo:
Ah, I was under the impression that with the current situation with CCGT-ext and OCGT-ext as extendable capacity the capital costs of conventional capacity could be put to zero here.

Should the extendability of all capacity then also be part of that big change?

Heronimonimo:
Updated the pull with the warning and opened an issue for a more definitive fix.

coroa:
Thanks

PR #15, opened 2019-02-18
Add CCGT to the weight factor for conventional carriers

Heronimonimo:


scripts/cluster_network.py
```diff
@@ -30,7 +30,7 @@ def normed(x):
     return (x/x.sum()).fillna(0.)
 
 def weighting_for_country(n, x):
-    conv_carriers = {'OCGT', 'PHS', 'hydro'}
+    conv_carriers = {'OCGT','CCGT','PHS', 'hydro'}
```
coroa: Comment from the other PR applies here as well. It would be better to make this configurable from the config file. Unfortunately here we have the additional complexity that we don't have access to `snakemake.config` when the call is coming from `simplify_network`.

PR #21, opened 2019-02-20
add_electricity.py & config.yaml: Add a simple load scaler

Heronimonimo:


coroa:
Is the term `load scaler` commonly used? I would have called it plainly `factor`.

coroa:
Can you please, in the future, follow my convention to prefix commit messages with the affected rule, if it's just one or two of them! (refer to https://github.com/PyPSA/pypsa-eur/commits/master)

Heronimonimo:
The idea was to make a category for the load in the config with a sub for scaling it. Scaler seems to me the appropriate term as it scales the load.

coroa:
Is `load: scaling_factor` as a name a good compromise :) ?

Heronimonimo:
Yeah, sure. Updated the commit to with the scaling_factor name.

PR #26, opened 2019-02-26
cluster_network.py: add focus country

Heronimonimo:


coroa:
The scheme only allows to specify a single focus country. I'd instead propose to base it on the following config:
```yaml
# Ratio of total clusters for particular countries, the remaining weight is distributed according to mean load
focus_weights:
    DE: 0.5
    NL: 0.1
```

coroa:
Your implementation has a couple of problems:
1. If `distribute_clusters` is called via `simplify_network`, it will not be able to access the `snakemake` object, so you should pass `focus_weights` as an argument through from the `if __name__ == "__main__"` block.

2. Try to be backward-compatible! If the user does not specify `focus_weights` in their config, you should not break, i.e. use something like `snakemake.config.get('focus_weights', {})` to use no weights as a fallback.

3. The `L` that's used in the optimisation problem must sum to `1` or it'll miss the `n_clusters` target. That's what `normed` is for.

Heronimonimo:
Thank you for taking the time to comment on the code!

I do get your points about backward compatibility and such. Unfortunately I just started learning to code in Python (being used to Matlab and R) by working with this model since a few weeks. So I don't think I'm able at this point in time to update the commit to include the points raised.

Please do feel free to close the pull request for now or edit this pull request to fix these points.

fneum:
closing in favor of #87

PR #27, opened 2019-02-26
config.yaml: Add focus country config handle

Heronimonimo:


coroa:
Should be part of #26 , since the PRs depend on each other!

PR #28, opened 2019-02-27
make_summary.py: Add export of marginal prices

Heronimonimo:
Marginal prices are an important feature of the electricity system, it is very interesting to export them.


fneum:
Since this is not calculating any additional information, I would prefer if for the analysis of marginal prices we leave it to inspecting `n.buses_t.marginal_price` directly from the network object; i.e.:
```python
import pypsa
n = pypsa.Network("path/to/network.nc")
n.buses_t.marginal_price`
```

PR #44, opened 2019-06-05
environment.yaml: install pypsa and dependencies through conda-forge

fneum:


environment.yaml
```diff
@@ -6,17 +6,7 @@ dependencies:
...
   - pip
 
-  # Dependencies of PyPSA
-  - six
-  - numpy
-  - pyomo
-  - scipy
-  - pandas>=0.22.0
-  - matplotlib
-  - memory_profiler
```
coroa: `memory_profiler` is not a PyPSA dependency. Earlier mistake of mine, belongs instead to the `Dependencies of the workflow group below`.

PR #45, opened 2019-06-06
add instructions to delete tar.xz and add natura.tiff from command line

fneum:
minor change in `README.md` for installing pypsa-eur


coroa:
Looks good! Although I did not test running the shell commands.

If you want, you can push documentation-only commits also directly to the master branch.

If you have another moment you could remove the note about needing pypsa from master branch from the top of the README.

Issue #33, opened 2019-03-05
Set min/max constraint for each technology per country


Heronimonimo:
What would be the advisable way to set a minimum/maximum constraint for technologies per country?

I was thinking about using a `.csv` with the same structure as the `costs.csv`. This could than be loaded at the `add_electricity` phase. Would this be the right way?

anyexingluzk:
As far as I understand, the add_electricity rule produces the non-clustered network. I am not sure if you can set the p_min/max on the country level. I am also trying to implement the same idea as you do, and my initial plan is loading it in prepare_network phase. Let's see if it works :)

Heronimonimo:
True, it's probably something that has to be put in as an extra optimization constraint at that state like the line cost limit.

Let me know if you have something for me to test with or have a look at.

Heronimonimo:
@anyexingluzk: any news on this?

anyexingluzk:
@Heronimonimo I have implemented the minimum capacity for a certain technology per country by adding p_nom_min = value, and p_nom_extendable = True in the prepare_network.py script.

Heronimonimo:
@anyexingluzk Nice! Can you make a pull request of your changes?

Heronimonimo:
@anyexingluzk Would you mind sharing the code you used to implement this?

anyexingluzk:
@Heronimonimo The way I have implemented is quite straightforward as below:
```
def replace_pmin_by_pnom(n):
    index = (n.generators.p_nom != 0) & (n.generators.p_nom_extendable == True) 
    n.generators.loc[index, 'p_nom_min'] = n.generators.loc[index, 'p_nom']
    return n
```
Basically, assign the value of p_nom to p_nom_min for extendable generators. Regarding the value of p_nom, it depends on your model, where I use existing generators in my case. It is also possible to load a csv file which includes all the limits.

Heronimonimo:
If I understand your code correctly it sets the nominal power of a generator as the minimum installed capacity. Does this also work for wind and solar?

anyexingluzk:
It works if the p_nom_extendable is True.

Heronimonimo:
Ah. And how would I be able to define the p_nom value for wind/solar per country?

(Sorry for al these questions, I'm quite new to PyPSA and Python)

anyexingluzk:
[...]

PR #46, opened 2019-06-17
solve_network.py & agg_p_nom_minmax.csv: add minmax constraint per tech

Heronimonimo:
Solution to #33 as discussed there. With much thanks to @coroa for doing many code suggestions!


scripts/solve_network.py
```diff
@@ -145,7 +161,10 @@ def run_lopf(n, allow_warning_status=False, fix_ext_lines=False):
         free_output_series_dataframes(n)
 
         pypsa.opf.network_lopf_build_model(n, formulation=solve_opts['formulation'])
+        
+        add_country_carrier_generation_constraints(n, opts)
         add_opts_constraints(n, opts)
```
fneum: I think, adding country carrier constraints should be optional and should be activated through a keywords in `snakemake.wildcards.opts`. For example 'CCL'?

```python
if 'CCL' in opts:
  add_country_carrier_generation_constraints(n, opts)
```

Another alternative is to move `add_country_carrier_generation_constraints(n, opts)` into the `add_opts_constraints(n, opts)`.

scripts/solve_network.py
```diff
@@ -77,6 +77,22 @@ def bau_mincapacities_rule(model, carrier):
         ext_gens_i = n.generators.index[n.generators.carrier.isin(conv_techs) & n.generators.p_nom_extendable]
         n.model.safe_peakdemand = pypsa.opt.Constraint(expr=sum(n.model.generator_p_nom[gen] for gen in ext_gens_i) >= peakdemand - exist_conv_caps)
 
+def add_country_carrier_generation_constraints(n, opts=None):
+    agg_p_nom_minmax = pd.read_csv("data/agg_p_nom_minmax.csv", index_col=list(range(2)))
```
fneum: Two remarks:
* This will fail if `data/agg_p_nom_minmax.csv` does not exist. There should be either
(a) a check that the file exists, and if not ensure that some warning is given and a default behaviour is implemented (e.g. limits are 0 and full potential of a country?) or
(b) you should raise a descriptive error message if the file could not be loaded.
* It might also make sense to specify the path to `agg_p_nom_minmax.csv` in the `config.yaml` to store multiple different limits in `./data`.

```python
agg_p_nom_limits = snakemake.config['electricity']['agg_p_nom_limits']
if os.path.isfile(agg_p_nom_limits):
  agg_p_nom_minmax = pd.read_csv(agg_p_nom_limits, index_col=list(range(2)))
else:
  raise FileNotFoundError("Need to specify the path to a .csv file containing aggregate capacity limits per country in config['electricity']['agg_p_nom_limit'].")
```

fneum: If you follow this path, can you also amend the `config.yaml`?

```yaml
electricity:
```

https://github.com/PyPSA/pypsa-eur/blob/master/config.yaml#L27-L37

Heronimonimo: I included your changes, thanks!

scripts/solve_network.py
Heronimonimo: Not entirely shure how to add this in. Please feel free to change my code!

scripts/solve_network.py
fneum: You're quick. I just thought this would be a better error message:
```
FileNotFoundError("{} not found. \
    Need to specify the path to an existing .csv file \
    containing aggregate capacity limits per country \
    in config['electricity']['agg_p_nom_limit'].".format(agg_p_nom_limits))
```

Heronimonimo: > You're quick. I just thought this would be a better error message:
> 
> ```python
> FileNotFoundError("{} not found. \
>     Need to specify the path to an existing .csv file \
>     containing aggregate capacity limits per country \
>     in config['electricity']['agg_p_nom_limit'].".format(agg_p_nom_limits))
> ```

In don't see any difference between this error message and the previous one you wrote.

coroa:
Fixes #33.

scripts/solve_network.py
coroa: I think I second both of Fabian's suggestions, make the inclusion of the constraints conditional on the CCL or some other short acronym, which means you can easily run it with and without the per-country capacity limits:

- `elec_s_12_lv1.25_Co2L-3H-CCL.nc` is then with the per-country-per-carrier constraints, while
- `elec_s_12_lv1.25_Co2L-3H.nc` does not have them

Heronimonimo: Sounds like a good plan to make it opt-in with the config handle.

Heronimonimo:
Just made a small fix to the Constraint.Skip. 

For me this pull request works nicely locally and is ready to merge into master.

PR #47, opened 2019-06-18
.gitignore: don't overwrite config.yaml

Heronimonimo:
It's quite annoying that everytime you pull the repo the local config gets overwritten.

Another option would be to split the config into multiple parts and .gitignore only certain parts.


coroa:
Hmm ... we could transition to having a config.example.yaml under version control and tell users to regularly merge with their used config.yaml by hand.

Heronimonimo:
Yes, good idea. Just made a copy named config.example.yaml in the patch. Ready to merge?

coroa:
Not yet, such an intrusive change needs documentation in installation instructions in the README that one has to copy `config.example.yaml` to `config.yaml` before use.

Heronimonimo:
> Not yet, such an intrusive change needs documentation in installation instructions in the README that one has to copy `config.example.yaml` to `config.yaml` before use.

Done! Added a little text in the paragraph about script overview.

fneum:
1. Would move the description to the installation section.
2. Corrected `config.example.yanl` to `config.example.yaml`
3. If `config.yaml` is in `.gitignore` it should to my understanding not be in the repository.

Recent changes (this branch is behind the master branch) should be added to the config.example.yaml!

coroa:
@fneum : Thanks for the changes, and you are right, we should remove `config.yaml` from the repository, which is quite a drastic change for the unsuspecting user.

I fear we have to switch to a proper scheme with release tags and release notes, soon. Let me ponder on that for a bit.

fneum:
closed via #65

PR #49, opened 2019-06-18
Add CO2 limit in `opts` wildcard

fneum:
This PR allows the user to specify a **relative emissions reduction target** in the `opts` wildcard.

Currently, `snakemake networks/elec_s_100_Co2L-3H.nc` will set the emissions limit based on the value given in `snakemake.config['electricity']['co2limit']`.

The new feature is that `snakemake networks/elec_s_100_Co2L0.05-3H.nc` will set the emissions limit based on a base emissions level specified in `snakemake.config['electricity']['co2base']` and the factor `0.05` given in the wildcard. All of `Co2L`, `Co2L0` and `Co2L0.05` are supported.

The **advantage** is that it is now easier to run with different emission reduction targets.

An **extension** to this could be allowing specifying absolute *and* relative emission reduction targets. E.g. also something akin to `Co2La7.7e+7` and `Co2Lr0.05`. But this looks a bit confusing.


coroa:
I'm in favour and I agree with the implementation, if you assure me it works fine for you, i'll merge. BUT: we really have to start pushing the documentation along. At this speed, even I might loose track sooner or later.

Heronimonimo:
Good feature. Would it be an idea to add a co2base.csv giving 1990 electricity sector CO2 emission numbers per country?

coroa:
@Heronimonimo I'm not opposed. Feel free.

Heronimonimo:
> @Heronimonimo I'm not opposed. Feel free.

Yes, made one you can find here: https://github.com/Heronimonimo/pypsa-eur/blob/master/data/co2base.csv

To get to these values I took:
- carbon intensity of heat & electricity from https://www.eea.europa.eu/data-and-maps/daviz/sds/co2-emission-intensity-from-electricity-generation-2/@@view
- electrical energy from http://appsso.eurostat.ec.europa.eu/nui/show.do?dataset=nrg_105a&lang=en

Not the best proxy. But it gives atleast something to work with.

fneum:
@coroa Agree that documentation is a very pressing issue.
Let me run a test before merging which I will post here to demonstrate it is working.

@Heronimonimo this .csv is great to integrate with `snakemake.config['countries']` so it automatically calculates the `co2base` depending on the selected countries.

coroa:
The sources should be documented in the CSV.

coroa:
I thought about your sources @Heronimonimo and I don't think, that's a particularly robust approach. 

It'd already be better to use the latest values (2012) from [EDGAR for category 1A1 (power industry)](https://edgar.jrc.ec.europa.eu/overview.php?v=432_GHG&SECURE=123) and maybe it makes sense to ping @nworbmot about a good per country sectoral co2 emissions data source.

Heronimonimo:
> I thought about your sources @Heronimonimo and I don't think, that's a particularly robust approach.
> 
> It'd already be better to use the values from [EDGAR for category 1A1 (power industry)](https://edgar.jrc.ec.europa.eu/overview.php?v=432_GHG&SECURE=123) (for 1990, of course) and maybe it makes sense to ping @nworbmot about a good per country sectoral co2 emissions data source.

Your totally right. The source you are giving seems much better!

scripts/prepare_network.py
```diff
@@ -16,10 +16,16 @@
 import pypsa
 from add_electricity import load_costs, update_transmission_costs
 
-def add_co2limit(n, Nyears=1.):
+def add_co2limit(n, Nyears=1., factor=None):
+
+    if factor:
```
fneum:
```suggestion
    if factor is not None:
```

PR #51, opened 2019-06-18
Add `extra_functionality` and `extra_postprocessing` to solve_network()

fneum:
This PR generalizes some parts of the `solve_networks()`.

For example the script `solve_operations_network.py` uses the `solve_network()` function of `solve_network.py` by

```python
from solve_network import patch_pyomo_tmpdir, solve_network, prepare_network
```

Another script may require slightly different functionality for `solve_network()`, but only marginally.
The idea is to pass some `extra_functionality` to modify the pyomo model before solving as well as some `extra_postprocessing` that requires the pyomo model analogous to `pypsa.Network.lopf()`.

Not sure if this is worth adding in `solve_network.py` or whether it would be better to have some code duplication and write `solve_network()` anew for each script requiring slightly different functionality.

Another feature is to skip the iteration process overall through `skip_iterating`.

@coroa Agree that documentation is a more pressing issue. This is not urgent, but wanted to put it on the table :).


scripts/solve_network.py
```diff
@@ -182,6 +189,9 @@ def run_lopf(n, allow_warning_status=False, fix_ext_lines=False):
             add_lv_constraint(n)
             add_lc_constraint(n)
 
+        if extra_functionality is not None:
+            extra_functionality(n, *extra_functionality_args)
```
FabianHofmann: Is there a specific reason for passing `extra_functionality_args`? The pypsa `extra_functionality` takes per default only `n` and `snapshots` . Perhaps keeping those would make sense and then there can be more keyword arguments `extra_functionality_args`, so 
```
extra_functionality(n, snapshots,  *extra_functionality_args)
```

fneum:
```suggestion
            extra_functionality(n, n.snapshots, *extra_functionality_args)
```

Issue #50, opened 2019-06-18
Adapt plotting scripts for `cartopy`


Heronimonimo:
When making a plot I get the error:
```
INFO:pypsa.io:Imported network elec_s_32_lc1.0_Ep-3H-CCL.nc has buses, carriers, generators, lines, links, loads, storage_units
WARNING:pypsa.plot:argument `basemap` is deprecated, use `geomap` instead.
Traceback (most recent call last):
  File "/home/pypsa/pypsa-eur/.snakemake/scripts/tmpjr9460ee.plot_network.py", line 118, in <module>
    ax=ax)
  File "/home/.conda/envs/pypsa-eur/lib/python3.7/site-packages/pypsa/plot.py", line 181, in plot
    'The passed axis is not a GeoAxesSubplot. You can '
AssertionError: The passed axis is not a GeoAxesSubplot. You can create one with:
import cartopy.crs as ccrs
fig, ax = plt.subplots(subplot_kw={"projection":ccrs.PlateCarree()})
```
This started happening after updating my conda. Maybe due to recent changes in pypsa?

fneum:
In PyPSA v0.14.0 we changed from `basemap` to `cartopy` for plotting country shapes and choosing projections (https://pypsa.org/doc/release_notes.html). This change is not yet reflected in `PyPSA-Eur`. 

All plots of a map should create axes like that
```python
fig, ax = plt.subplots(subplot_kw={"projection":ccrs.PlateCarree()})
```
Particularly the `subplot_kw` argument is important to specify the projection.

Heronimonimo:
I just replaced the `fig, ax = ...` line by:
```
import cartopy.crs as ccrs
fig, ax = plt.subplots(subplot_kw={"projection":ccrs.PlateCarree()})
```
That seems to fix part of the problem. But now I run into two problems. 

Firstly 'OCGT-ext' and 'CCGT-ext' are not included in the new tech-colors. This can be fixed by either adding them to the list or adding some code to remove the '-ext' extension while plotting.

Secondly:
```
Traceback (most recent call last):
  File "/home/pypsa/pypsa-eur/.snakemake/scripts/tmprhs9nf08.plot_network.py", line 127, in <module>
    ax=ax)
  File "/home/.conda/envs/pypsa-eur/lib/python3.7/site-packages/pypsa/plot.py", line 173, in plot
    if use_cartopy and geomap:
UnboundLocalError: local variable 'use_cartopy' referenced before assignment
```
[...]

PR #52, opened 2019-06-18
plot_network.py: fix for cartopy issue in pypsa/plot.py

Heronimonimo:
Quick fix for #50 as discussed in while waiting for definitive solution in PyPSA/PyPSA#75.

### DIFF
--- a/scripts/plot_network.py
+++ b/scripts/plot_network.py
@@ -112,13 +112,14 @@ n.plot(line_widths=pd.concat(line_widths_exp)/linewidth_factor,
        bus_sizes=bus_sizes/bus_size_factor,
        bus_colors=tech_colors,
        boundaries=map_boundaries,
+       geomap=True,
        ax=ax)
 n.plot(line_widths=pd.concat(line_widths_cur)/linewidth_factor,
        line_colors=pd.concat(line_colors_with_alpha),
        bus_sizes=0,
        bus_colors=tech_colors,
        boundaries=map_boundaries,
-       geomap=False,
+       geomap=True,
        ax=ax)
 ax.set_aspect('equal')
 ax.axis('off')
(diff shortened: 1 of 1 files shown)


Heronimonimo:
One thing to make shure is that changes also work for those running a pypsa version from conda-forge.

fneum:
This should actually work for the current conda-forge version.

I made a few minor changes to your PR.
* One of the plots should not plot the country outlines.
* plt.subplots() still requires the original figsize argument.

Heronimonimo:
> This should actually work for the current conda-forge version of PyPSA if I am not mistaken.
> 
> I made a few minor changes to your PR.
> 
> * One of the plots should not plot the country outlines.
> * plt.subplots() still requires the original figsize argument.

I'm using the conda-forge version and unfortunately it doesn't work for me, still getting:
`UnboundLocalError: local variable 'use_cartopy' referenced before assignment`

In my case `geomap=True` is needed in the `n.plot` to fix it.

FabianHofmann:
>I'm using the conda-forge version and unfortunately after the changes you made it doesn't work for me, getting:
>UnboundLocalError: local variable 'use_cartopy' referenced before assignment
>
>In my case geomap=True is needed in the n.plot to fix it.

Yes, I think so too. The geomap=True is necessary to not run into the (purely) pypsa bug. As I see right now, the only disadvantage is slightly thicker borders when plotting two times with geomap=True

fneum:
Ah, ok. Sorry for the misunderstanding. We should just keep in mind then to turn `geomap=False` in the second `n.plot()` once we have a new PyPSA release.

coroa:
Will merge after the PR solves #50 completely. Ie, it handles plotting of technologies with carriers like `OCGT-ext`. For reference, these technology "sub"-types were included too allow fixed and extendable powerplants at the same time. I think it would be best to fall back to using the `OCGT` only `tech_colors` in such a case. I think they are only possible for conventional generators.

FabianHofmann:
close as outdated

PR #68, opened 2019-08-13
Add an example to the documentation of pypsa-eur, for now: just rst, …

eb5194:
…data missing. Data (an additional folder called example) will be added when the new version of atlite is online, such that we can provide a smaller package (one day or a few days instead of a whole  month). By then, example.rst should also be adapted to this change. Folder examples will contain the era5-cutout, the corresponding config.yaml, the solved network (elec_s_10_lvopt_Co2L-3H.nc) and an exemplary notebook pypsa-eur-example.ipynb which just reads the provided network and does a very brief analysis.


fneum:
Work on this will continue in the `documentation` branch.

PR #69, opened 2019-08-17
allow usage of own powerplants for selected country

eb5194:
in build_powerplants.py add the feature to add own carriers (add_my_carriers) and restrict to a specific build year (restrict_buildyear), which both can be configured in config.yaml (my_carriers_for_countries, my_carriers_switch, restrict_buildyear).

### DIFF
--- a/scripts/build_powerplants.py
+++ b/scripts/build_powerplants.py
@@ -18,30 +18,27 @@ def country_alpha_2(name):
         cntry = pyc.countries.get(official_name=name)
     return cntry.alpha_2
 
-def add_my_carriers(ppl):
-    switch = snakemake.config['electricity']['my_carriers_switch']
-    if switch == 'replace-all' or switch == 'replace-selection' or switch == 'add':
-        countries_dict = snakemake.config['countries_dict'] # dictionary, eg. GB: United Kindgom
-        for country in snakemake.config['electricity']['my_carriers_for_countries']:
-            dirname = os.path.abspath(os.path.join(os.path.dirname(__file__),".."))
-            add_ppls = pd.read_csv(dirname + "/resources/powerplants_" + country + ".csv", index_col=0)
-            if switch == 'replace-all' or switch == 'replace-selection':
-                to_drop = ppl[ppl.Country == countries_dict[country]]
-                if switch == 'replace-selection':
-                    to_drop = to_drop[to_drop.Fueltype.isin(add_ppls.groupby('Fueltype').mean().index)]
-                ppl = ppl.drop(to_drop.index)
-            ppl = ppl.append(add_ppls, sort='False')
-    else:
-        logger.warning('my_carriers_switch is invalid keyword, try one of [add, replace-all, replace-selection]. powerplants remain unchanged.')
-    return ppl
+def add_custom_carriers(ppl):
+    switch = snakemake.config['electricity']['custom_powerplants']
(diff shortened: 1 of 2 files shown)


FabianHofmann:
Hey! I think the setup it nice and should work well. Just as counterproposal:

There is *one* optional, additional dataset, called custom_powerplants.csv in the resource folder. This is added to the standard powerplant list if in config.py the default

```yaml
custom_powerplants: false
``` 

is either set to `true` or to a [pandas.query string](https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.DataFrame.query.html). The latter enables you to filter the additional data to the settings you want, *e.g.*:

```yaml
custom_powerplants: Country in ['DE', 'FR'] & YearCommisioned < 2050
``` 

The same filter option could be enabled for the standard powerplant list, *e.g.*

```yaml
powerplants_filter: Country not in ['DE', 'FR'] & YearCommisioned < 2050
``` 

(The powerplants_filter should come first in the config.yaml, though. )

This could strongly reduce the code such as :

```python
def add_custom_carriers(ppl):
    custom_ppl_query = snakemake.config['electricity']['custom_powerplants'']
    if not custom_ppl_query:
        return ppl
    # import custom_ppl here
    if isinstance(custom_ppl_query, str):
        custom_ppl.query(custom_ppl_query, inplace=True)
    return ppl.append(cusom_ppl)
```

And of course the filter for the standard ppl, not as a function but in the 'main' code.

coroa:
Couple of comments:
- `custom` is a better name, than `my`
- input data which is not generated, has to live in `data`: so put it powerplants in `data/custom_powerplants.csv`
- @FabianHofmann's proposal to take advantage of the pandas query syntax sounds great to me.
- The use of such a query string means of course that you have to preprocess the csv file and clean the YearCommissioned column to a single integer.

scripts/build_powerplants.py
```diff
@@ -18,6 +18,31 @@ def country_alpha_2(name):
...
+                    to_drop = to_drop[to_drop.Fueltype.isin(add_ppls.groupby('Fueltype').mean().index)]
+                ppl = ppl.drop(to_drop.index)
+            ppl = ppl.append(add_ppls, sort='False')
+    else:
+        logger.warning('my_carriers_switch is invalid keyword, try one of [add, replace-all, replace-selection]. powerplants remain unchanged.')
+    return ppl
+
+def restrict_buildyear(ppl):
+    year = snakemake.config['electricity']['restrict_buildyear']
+    search_pattern = [str(int(year)+x) for x in range(1,2050-int(year))]
```
coroa: Took me a while to reason about, what you are searching for, instead use:
```python
search_pattern = [str(x) for x in range(int(year),2050)]
```
or even
```python
search_pattern = map(str, range(int(year), 2050))
```

eb5194:
Hi both, thanks for your input. I made few changes from both your proposals, yet I would like to keep the separation - filtering the year is a different issue than replacing or adding custom powerplants.

eb5194:
>     * The use of such a query string means of course that you have to preprocess the csv file and clean the YearCommissioned column to a single integer.

I did a histogram to see how many of the data is wrong processed by just casting to int. It's less than 1%, so I won't do any preprocessing as there are already so many NaN values, this will not have such a huge impact

FabianHofmann:
cl

PR #76, opened 2019-10-24
Update make_summary for pandas-0.25 compatability.

euronion:
The pandas Index union operation does no longer produce a MultiIndex but a Index tuple in pandas-0.25.2.
This causes the script to fail. (The problem does not exist with pandas-0.24.0.)

Fix by explicitly checking for this case (only the first loop iteration).

### DIFF
--- a/scripts/make_summary.py
+++ b/scripts/make_summary.py
@@ -38,7 +38,11 @@ def calculate_costs(n,label,costs):
         capital_costs = c.df.capital_cost*c.df[opt_name.get(c.name,"p") + "_nom_opt"]
         capital_costs_grouped = capital_costs.groupby(c.df.carrier).sum()
 
-        costs = costs.reindex(costs.index|pd.MultiIndex.from_product([[c.list_name],["capital"],capital_costs_grouped.index]))
+        new_index = costs.index|pd.MultiIndex.from_product([[c.list_name],["capital"],capital_costs_grouped.index])
+        if isinstance(new_index, pd.Index):
+            new_index = pd.MultiIndex.from_tuples(new_index)
+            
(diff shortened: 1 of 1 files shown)


scripts/make_summary.py
```diff
@@ -37,10 +37,14 @@ def calculate_costs(n,label,costs):
     for c in n.iterate_components(n.branch_components|n.controllable_one_port_components^{"Load"}):
         capital_costs = c.df.capital_cost*c.df[opt_name.get(c.name,"p") + "_nom_opt"]
         capital_costs_grouped = capital_costs.groupby(c.df.carrier).sum()
+        
+        new_index = costs.index|pd.MultiIndex.from_product([[c.list_name],["capital"],capital_costs_grouped.index])
+        if isinstance(new_index, pd.Index):
+            new_index = pd.MultiIndex.from_tuples(new_index)
+            
+        costs = costs.reindex(new_index)
```
FabianHofmann: wouldn't it make more sense to make a small function which does that? Then you don't have to repeat the code all time 

```
def extend_multiindex(ds, index):
    new_index = ds.index|index
    if isinstance(new_index, pd.Index):
        new_index = pd.MultiIndex.from_tuples(new_index)
    return ds.reindex(new_index)
```
could make the proper reindexing. Then it just be 

```
extend_i = pd.MultiIndex.from_product([[c.list_name],["capital"],
                                               capital_costs_grouped.index])
costs = extend_multiindex(costs, extend_i)
```

euronion: Sure. It just started as a single point problem and turned out to be quickly solveable w/ copy-pasta.

I refractored and also removed some duplicate actions when recreating the same index for value assignment.

PR #83, opened 2019-10-28
update powerplants.csv

### LINKED ISSUE #61: Update powerplants.csv from new powerplantmatching release
https://github.com/FRESNA/powerplantmatching/releases/tag/v0.4.1

In this process, switch to conda packages for atlite, pypsa and powerplantmatching.

fneum:
closes #61 

Obtained from
```
import powerplantmatching as pm
pm.powerplants(from_url=True)
```

## Question

@FabianHofmann Would it make sense to put `pm.powerplants(from_url=True)` in `build_powerplants.py` and delete `powerplants.csv` from the PyPSA-Eur repository?


FabianHofmann:
yes, totally! 

I think the only reason that Jonas put it there was that installing powerplantmatching properly seemed to be an issue. However, it shouldn't be any problem using the [pip installation](https://pypi.org/project/powerplantmatching/). The conda installation is still pending...

fneum:
Let's leave it in for now, but if `config.enable.powerplantmatching: true` will download and update from powerplantmatching

PR #84, opened 2019-10-31
Update powerplants according to new powerplantmatching version

FabianHofmann:
This PR inlcudes a revision of `build_powerplants.py` and `add_electricity.py`

Conclusion: 

* clean code, break very long lines and use powerplantmatching (ppm) functions (since it is in the environment now) for specific conversions
* exploit all important information of ppm data. This counts especially for hydro data, where information like max_hours and clear distinction between ror, phs and hydro dams can be derived from the [JRC data](https://github.com/energy-modelling-toolkit/hydro-power-database) 
* include functionality proposed in #69 (even though strongly reduced, I hope @eb5194 is fine with it (due to flexibility `query` functions are the right choice here))

Some technical remarks:

* `load_powerplants` does not take the network as an argument anymore, as the filtering of nodes (as for what it was used before) is already taking place in `build_powerplants.py`
* `load_powerplants` now converts column names according to `pypsa.generators` columns. Before, this was done on the run in `attach_conventional_generators`, but since it is also necessary for `attach_extendable_generators`, I guess, it makes more sense in this way
* `attach_hydro` now splits the `ppl` frame into `ror`, `phs` and `hydro` pieces where only respective generators are included. These are then directly `madd`ed to the network (if not empty). There is no inflow for phs anymore as this information is not statet by the [JRC data](https://github.com/energy-modelling-toolkit/hydro-power-database) 
* if hydro dams are attached in `attach_hydro` and `hydro_max_hours == 'energy_capacity_totals_by_country'`, then *only*  missing values of `max_hours` are scaled up. Other hydro dams with non-NA `max_hours` value are untouched.


scripts/build_powerplants.py
```diff
@@ -35,71 +36,68 @@
...
-        ppl_b = ppl_country == cntry
-        if not ppl_b.any():
-            cntries_without_ppl.append(cntry)
-            continue
+    substation_i = n.buses.query('substation_lv').index
+    kdtree = KDTree(n.buses.loc[substation_i, ['x','y']].values)
 
-        kdtree = KDTree(n.buses.loc[substation_lv_i, ['x','y']].values)
-        ppl.loc[ppl_b, 'bus'] = substation_lv_i[kdtree.query(ppl.loc[ppl_b, ['lon','lat']].values)[1]]
+    ppl['bus'] = substation_i[kdtree.query(ppl[['lon','lat']].values)[1]]
```
coroa: I tried here to not mistakenly assign a powerplant to a neighbouring bus in another country.

Can you output a warning of how many and maybe which powerplants, are now assigned to the "wrong" country!

scripts/add_electricity.py
```diff
@@ -327,100 +337,99 @@ def attach_hydro(n, costs, ppl):
...
+             .rename(index=lambda s: str(s) + ' hydro')
+    ror = ppl.query('technology == "Run-Of-River"')
+    phs = ppl.query('technology == "Pumped Storage"')
+    hydro = ppl.query('technology == "Reservoir"')
 
     country = ppl['bus'].map(n.buses.country).rename("country")
 
-    if ppl.has_inflow.any():
-        dist_key = ppl.loc[ppl.has_inflow, 'p_nom'].groupby(country).transform(normed)
+    inflow_idx = ror.index ^ hydro.index
```
coroa:
```suggestion
    inflow_idx = ror.index | hydro.index
```
`^` seems to be an `xor`, ie include it if it in exactly one of the two indices. in principle here one would want rather a full `or`, ie if its in one of the two. in this special case they are equivalent since their "schnitt" (`&`) is empty.

scripts/add_electricity.py
```diff
@@ -327,100 +337,99 @@ def attach_hydro(n, costs, ppl):
...
+            hydro_capacities = pd.read_csv(snakemake.input.hydro_capacities,
+                                           comment="#", na_values='-', index_col=0)
+            estim_hydro_max_hours = hydro_capacities['E_store[TWh]'] * 1e3 / \
+                                    hydro_capacities['p_nom_discharge[GW]']
+
+            missing_countries = (pd.Index(hydro['country'].unique())
+                                 .difference(estim_hydro_max_hours.dropna().index))
+            if not missing_countries.empty:
+                logger.warning("Assuming max_hours=6 for hydro reservoirs in the countries: {}"
+                               .format(", ".join(missing_countries)))
```
coroa: Do we still need these amazingly crude estimations, how well does just using the JRC data work out? 

`snakemake.input.hydro_capacities` contain hand researched energy storage capacities from Emil. Maybe check in his thesis for the full references: `/home/vres/data/literature/MasterThesis/150729_MasterThesis_EmilEriksen.pdf`

scripts/build_powerplants.py
```diff
@@ -35,71 +36,68 @@
...
+    return ppl.append(add_ppls, sort=False)
 
-def country_alpha_2(name):
-    try:
-        cntry = pyc.countries.get(name=name)
-    except KeyError:
-        cntry = None
-    if cntry is None:
-        cntry = pyc.countries.get(official_name=name)
-    return cntry.alpha_2
```
coroa: Does it make sense to migrate the fallback to `official_name` also to `pm.utils`?

scripts/build_powerplants.py
FabianHofmann: yes, good point. I did not think about it. I think I'll reverse that.

scripts/add_electricity.py
FabianHofmann: ah yes, thanks for clarifying the difference.

scripts/add_electricity.py
FabianHofmann: I think we have to rely on them a bit longer. So far only roughly a third of all hydro dams in JRC have stated storage capacities. But then on the other hand, the question is, is it worth it to modify the option 
```
hydro_max_hours == 'estimate_by_large_installations'
```
so that non-NA `max_hours` values are untouched?

scripts/build_powerplants.py
FabianHofmann: yes, done in https://github.com/FRESNA/powerplantmatching/commit/411c91a5f836ae54dedcb8430da777a13dbbc6eb

scripts/add_electricity.py
coroa: Yes, i'd think so!

eb5194:
Hi Fabian, sure I'm fine with reducing the code and your adaption (which I also tried to realise in a later commit in #64 , see https://github.com/PyPSA/pypsa-eur/pull/69/commits/ffef74e9c54fb25f0216eaab4b2e0b27f01d9ae1). Yet the functionality now is a bit different if I understand correctly? I suggested to allow the user to add a type of carrier in a specific country **or** replace all of them (instead of only replacing as in your version above). Also I cannot find the restriction to a specific build year? That was only 5 lines of code or so

fneum:
Let's make the build year restriction a separate PR.

FabianHofmann:
hey @eb5194, the query function allows you to do all three points

1. Adding all powerplants from custom

        powerplants_filter: false
        custom_powerplants: true

2. Replacing powerplants in e.g. Germany by custom data

        powerplants_filter: Country not in ['Germany']
        custom_powerplants: true

    or 

        powerplants_filter: Country not in ['Germany']
        custom_powerplants: Country in ['Germany']


3. Adding additional built year constraint

        powerplants_filter: Country not in ['Germany'] and YearCommissioned <= 2015
        custom_powerplants: YearCommissioned <= 2015

Does that cover what you mean?

eb5194:
Yes it does. Thanks for the explanation - I imagined query's functionality less powerful, then. Awesome!

FabianHofmann:
ok cool

PR #87, opened 2019-11-06
cluster: add focus weights (in favor of #26)

fneum:
This PR offers a different implementation for `focus_weights` which allows to set a proportion of the total number of clusters for particular countries. The rest (countries not listed) is distributed according to mean load. Defaults to no focus weights.

In the ``config.yaml`` add for instance:

```yaml
focus_weights:
    DE: 0.9
    FR: 0.1
```

which results in
focus_weights

The sum of focus weights must be less than 1.

### DIFF
--- a/scripts/cluster_network.py
+++ b/scripts/cluster_network.py
@@ -166,7 +166,7 @@ def distribute_clusters(n, n_clusters, focus_weights=None, solver_name=None):
 
         total_focus = sum(list(focus_weights.values()))
 
-        assert total_focus <= 1.0, "The sum of focus weights must be less than 1."
+        assert total_focus <= 1.0, "The sum of focus weights must be less than or equal to 1."
         
         for country, weight in focus_weights.items():
             L[country] = weight
--- a/scripts/cluster_network.py
+++ b/scripts/cluster_network.py
@@ -7,6 +7,8 @@ Relevant Settings
 
 .. code:: yaml
 
+    focus_weights:
+
     renewable: (keys)
         {technology}:
             potential:
@@ -20,7 +22,7 @@ Relevant Settings
 
 .. seealso:: 
     Documentation of the configuration file ``config.yaml`` at
-    :ref:`renewable_cf`, :ref:`solving_cf`, :ref:`lines_cf`
+    :ref:`toplevel_cf`, :ref:`renewable_cf`, :ref:`solving_cf`, :ref:`lines_cf`
 
 Inputs
 ------
@@ -146,7 +148,7 @@ def plot_weighting(n, country, country_shape=None):
 
 # # Determining the number of clusters per country
 
-def distribute_clusters(n, n_clusters, solver_name=None):
+def distribute_clusters(n, n_clusters, focus_weights=None, solver_name=None):
     if solver_name is None:
         solver_name = snakemake.config['solving']['solver']['name']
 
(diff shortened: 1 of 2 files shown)


scripts/cluster_network.py
```diff
@@ -160,6 +162,22 @@ def distribute_clusters(n, n_clusters, solver_name=None):
...
         "Number of clusters must be {} <= n_clusters <= {} for this selection of countries.".format(len(N), N.sum())
 
+    if focus_weights is not None:
+
+        total_focus = sum(list(focus_weights.values()))
+
+        assert total_focus <= 1.0, "The sum of focus weights must be less than 1."
+        
+        for country, weight in focus_weights.items():
+            L[country] = weight
```
FabianHofmann:
```suggestion
            L[country] = weight / len(L[country])
```
Like this you don't have to worry about how many subnetworks lie in the country and L always sums up to zero. So the assertion check below is not necessary

scripts/cluster_network.py
```diff
@@ -160,6 +162,22 @@ def distribute_clusters(n, n_clusters, solver_name=None):
...
+
+        remainder = [c not in focus_weights.keys() for c in L.index.get_level_values('country')]
+        L[remainder] = L.loc[remainder].pipe(normed) * (1 - total_focus)
+
+        logger.warning('Using custom focus weights for determining number of clusters.')
+
+    assert L.sum() == 1.0, "Country weights L must sum up to 1.0 when distributing clusters."
```
FabianHofmann:
```suggestion
```

scripts/cluster_network.py
```diff
@@ -160,6 +162,22 @@ def distribute_clusters(n, n_clusters, solver_name=None):
     assert n_clusters >= len(N) and n_clusters <= N.sum(), \
         "Number of clusters must be {} <= n_clusters <= {} for this selection of countries.".format(len(N), N.sum())
 
+    if focus_weights is not None:
+
+        total_focus = sum(list(focus_weights.values()))
+
+        assert total_focus <= 1.0, "The sum of focus weights must be less than 1."
```
FabianHofmann:
```suggestion
        assert total_focus <= 1.0, "The sum of focus weights must be less or equal 1."
```

scripts/cluster_network.py
fneum: No, the idea is for the focus weight is to specify the share of clusters to be allocated to that country (e.g. DE: 0.5 should result in 50% of the nodes allocated to Germany and the remainder is distributed proportional to mean load for all the other countries which are not listed). Your suggestion would distort this.

fneum:
Thanks for reviewing :+1:

scripts/cluster_network.py
FabianHofmann: Okay, quick question: 
What if Denmark has 2 synchronous zones and we want to give Denmark 40% of all nodes. Then `L["DK"]` would include two entries. When we give them both a value of 0.4, the total of Denmark would have a weight of 80%. By dividing by `len(L["DK"])`, we guarantee that the resulting total weight of denmark is 40%. 

Am I missing a point here?

fneum: Ah, I was confused because of the assertion check which should still be done, but you are right.

fneum: changed in https://github.com/PyPSA/pypsa-eur/commit/7df2b76f9daeb8d2775b833edf486c770e46c682

PR #94, opened 2019-11-08
Powerplants

FabianHofmann:
enable filtering for decommissioning years, in `configf.yaml` use e.g.

    powerplants_filter: YearDecommissioning >= 2030

for discarding powerplants which will be decommissioned until 2030. But be aware, the decommissioning dates are strongly relying on heuristics!


fneum:
Merci!

PR #96, opened 2019-11-12
own databundle

JanFrederickUnnewehr:
add and option to use own databundles in `config.yaml`

```
enable:
  prepare_links_p_nom: false
  build_cutout: false
  build_natura_raster: false
  retrieve_databundle: enable # enable: to retrieve databundle from zenodo; false: to use own databundle
```

### DIFF
--- a/Snakefile
+++ b/Snakefile
@@ -24,7 +24,7 @@ rule solve_all_elec_networks:
         expand("results/networks/elec_s{simpl}_{clusters}_l{ll}_{opts}.nc",
                **config['scenario'])
 
-if config['enable'].get('prepare_links_p_nom', false):
+if config['enable'].get('prepare_links_p_nom', False):
     rule prepare_links_p_nom:
         output: 'data/links_p_nom.csv'
         threads: 1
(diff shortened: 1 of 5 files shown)


Snakefile
```diff
@@ -39,12 +39,14 @@ datafiles = ['ch_cantons.csv', 'je-e-21.03.02.xls',
...
             'nama_10r_3gdp.tsv.gz', 'time_series_60min_singleindex_filtered.csv', 
             'corine/g250_clc06_V18_5.tif']
+
 if not config['tutorial']:
     datafiles.extend(["natura/Natura2000_end2015.shp", "GEBCO_2014_2D.nc"])
 
-rule retrieve_databundle:
-    output:  expand('data/bundle/{file}', file=datafiles)
-    script: 'scripts/retrieve_databundle.py'
+if config['enable']['retrieve_databundle']:
```
```suggestion
if config['enable'].get('retrieve_databundle', true):
```

fneum:
I am not opposed. A few remarks, however, before I merge:

- I would stick to `true/false` here. I.e. `enable: retrieve_databundle: true`.

- As you might have seen the continuous integration test failed. This is due to the fact that your implementation makes specifying `enable: retrieve_databundle: ` obligatory, but you did not amend the additional setting in the other `config*.yaml` files. Please add the setting to `config.tutorial.yaml` and to `config.test1.yaml`. 

- A default for `enable: retrieve_databundle: ` would be nice in the `Snakefile`: 

```python
if config['enable'].get('retrieve_databundle', true):`
```

- It would also be great if you could provide a small note in the documentation at [`doc/configtables/toplevel.csv`](https://github.com/PyPSA/pypsa-eur/blob/master/doc/configtables/toplevel.csv)

JanFrederickUnnewehr:
Hey Fabian, thanks for the implementation. You're just faster than me. Next time I have to hurry. ;-)

PR #97, opened 2019-11-13
battery+H2: option for stores and links instead of storage_units

fneum:
**Goal:** add more flexible variant of storage expansion without a fixed ratio between power and energy rating.

I have the feeling that it would actually make sense to move the functions `attach_storage(...)` and `attach_stores(...)` to a rule after simplification and clustering. Not sure yet.


FabianHofmann:
@fneum I'm having a look at this now ;) It definitely makes sense to move the attachment of (especially) stores and storageunits behind the clustering. I don't think the clustering can handle the links-storebus-link nexus in an efficient manner. I'd suggest to add a rule `add_extra_stores` which comes after simplifying. Any other thoughts?

FabianHofmann:
I moved the attachement of stores/storageunits in add_extra_stores (no doc done yet). As far as I see one has to place and extra attribute to the output to trigger `snakemake`. So it would be 

    snakemake networks/elec_s_37_es.nc 

`es` for extra stores (?). But I'm not sure if that is really necessary haven't completely figured out how snakemake does this...

Snakefile
```diff
@@ -219,6 +219,19 @@ rule cluster_network:
     # group: 'build_pypsa_networks'
     script: "scripts/cluster_network.py"
 
+
+rule add_extra_stores:
+    input:
+        network='networks/{network}_s{simpl}_{clusters}.nc',
+        tech_costs=COSTS,
+    output: 'networks/{network}_s{simpl}_{clusters}.nc'
+    benchmark: "benchmarks/add_extra_stores/{network}_s{simpl}_{clusters}_f"
```
```suggestion
    benchmark: "benchmarks/add_extra_stores/{network}_s{simpl}_{clusters}"
```
typo?

Snakefile
```diff
@@ -219,6 +219,19 @@ rule cluster_network:
     # group: 'build_pypsa_networks'
     script: "scripts/cluster_network.py"
 
+
+rule add_extra_stores:
```
```suggestion
rule add_additional_components:
```

```suggestion
rule add_extra_components:
```
I was thinking to make/name this rule more general to add any energy system components that should/cannot be added before clustering. E.g. another extension that would fit well in this rule is to build expansion options for a generic hydrogen network.

Snakefile
```diff
@@ -219,6 +219,19 @@ rule cluster_network:
...
+    input:
+        network='networks/{network}_s{simpl}_{clusters}.nc',
+        tech_costs=COSTS,
+    output: 'networks/{network}_s{simpl}_{clusters}.nc'
```
fneum: Yes, I think it is necessary to reflect this rule in the output file names. I like `_ec` (for extra components) or `C` (for Components). I would avoid `_ac` (for additional components) due to ambiguity with AC power flow.
```suggestion
    output: 'networks/{network}_s{simpl}_{clusters}_ec.nc'
```

scripts/add_electricity.py
```diff
@@ -595,13 +521,11 @@ def add_nice_carrier_names(n):
     attach_load(n)
 
     update_transmission_costs(n, costs)
-    attach_conventional_generators(n, costs, ppl)
 
+    attach_conventional_generators(n, costs, ppl)
     attach_wind_and_solar(n, costs)
-    if 'hydro' in snakemake.config['renewable']:
-        attach_hydro(n, costs, ppl)
+    attach_hydro(n, costs, ppl)
```
fneum: Yes, I think hydro should stay in `add_electricity`.

scripts/add_extra_stores.py
```diff
@@ -0,0 +1,219 @@
...
+- today's transmission topology and transfer capacities (optionally including lines which are under construction according to the config settings ``lines: under_construction`` and ``links: under_construction``),
+- today's thermal and hydro power generation capacities (for the technologies listed in the config setting ``electricity: conventional_carriers``), and
+- today's load time-series (upsampled in a top-down approach according to population and gross domestic product)
+
+It further adds extendable ``generators`` and ``storage_units`` with **zero** capacity for
+
+- photovoltaic, onshore and AC- as well as DC-connected offshore wind installations with today's locational, hourly wind and solar capacity factors (but **no** current capacities),
+- long-term hydrogen and short-term battery storage units (if listed in the config setting ``electricity: extendable_carriers``), and
+- additional open- and combined-cycle gas turbines (if ``OCGT`` and/or ``CCGT`` is listed in the config setting ``electricity: extendable_carriers``)
+"""
```
fneum: Ok, this is an open TODO marker.

scripts/add_extra_stores.py
```diff
@@ -0,0 +1,219 @@
...
+import logging
+import pandas as pd
+import pypsa
+from add_electricity import load_costs
+
+idx = pd.IndexSlice
+logger = logging.getLogger(__name__)
+
+
+def normed(s): return s/s.sum()
```
fneum: import from `add_electricity` or move to `_helpers.py`

scripts/add_extra_stores.py
```diff
@@ -0,0 +1,219 @@
...
+
+def _add_missing_carriers_from_costs(n, costs, carriers):
```
fneum: import from `add_electricity` or move to `_helpers.py`

scripts/add_extra_stores.py
```diff
@@ -0,0 +1,219 @@
...
+               p_nom_extendable=True)
+
+        n.madd("Link", b_buses_i + " discharger",
+               bus0=b_buses_i,
+               bus1=buses_i,
+               efficiency=costs.at['battery inverter','efficiency'],
+               capital_cost=costs.at['battery inverter', 'capital_cost'],
+               p_nom_extendable=True)
+
+def add_nice_carrier_names(n):
```
fneum: import from `add_electricity` or move to `_helpers.py`

scripts/add_extra_stores.py
```diff
@@ -0,0 +1,219 @@
...
+
+if __name__ == "__main__":
+    # Detect running outside of snakemake and mock snakemake for testing
+    if 'snakemake' not in globals():
+        from vresutils.snakemake import MockSnakemake, Dict
+
+        snakemake = MockSnakemake(output=['networks/elec_s_5.nc'])
+        snakemake.input = snakemake.expand(
+            Dict(network='networks/elec_s_5.nc',
+                 tech_costs='data/costs.csv'))
```
fneum:
```python
        snakemake = MockSnakemake(output=['networks/elec_s_5_ec.nc'])
        snakemake.input = snakemake.expand(
            Dict(network='networks/elec_s_5_ec.nc',
                 tech_costs='data/costs.csv'))

File: scripts/add_extra_stores.py [hunk 10]
```diff
@@ -0,0 +1,219 @@
...
+
+    logging.basicConfig(level=snakemake.config['logging_level'])
+
+    n = pypsa.Network(snakemake.input.network)
+    Nyears = n.snapshot_weightings.sum()/8760.
```
fneum: I don't think we need this here.

fneum:
> I moved the attachement of stores/storageunits in add_extra_stores (no doc done yet). As far as I see one has to place and extra attribute to the output to trigger snakemake. So it would be
``snakemake networks/elec_s_37_es.nc ``
es for extra stores (?). But I'm not sure if that is really necessary haven't completely figured out how snakemake does this...

I would keep it more general, e.g. `ec` for extra components and we should definitely add a suffix to the output filename and carry it over to all downstream rules.

fneum:
Test passed bc rule extra_stores not triggered by Snakefile:

```bash
$ snakemake solve_all_elec_networks
Building DAG of jobs...
Using shell: /bin/bash
Provided cores: 1 (use --cores to define parallelism)
Rules claiming more threads will be scaled down.
Job counts:
	count	jobs
	1	add_electricity
	1	base_network
	1	build_bus_regions
	1	build_powerplants
	4	build_renewable_profiles
	1	build_shapes
	1	cluster_network
	1	prepare_network
	1	retrieve_cutout
	1	retrieve_databundle
	1	retrieve_natura_raster
	1	simplify_network
	1	solve_all_elec_networks
	1	solve_network
	17
```

Snakefile
FabianHofmann: yes, more or less (that was my original suffix)

scripts/add_extra_stores.py
FabianHofmann: You mean the `Nyears`? It is used below for the cost assumptions

FabianHofmann:
> I would keep it more general, e.g. `ec` for extra components and we should definitely add a suffix to the output filename and carry it over to all downstream rules (even if it may not be strictly necessary).

Yes! Also renamed script `add_extra_stores` &rarr; `add_extra_components`

scripts/add_extra_stores.py
fneum: yes, I was absolutely wrong here.

fneum:
Just tested up to including `solve_network` with different settings:
```yaml
    StorageUnit: [battery]
    Store: [H2]
```
```yaml
    StorageUnit: [battery, H2]
    Store: []
```
```yaml
    StorageUnit: []
    Store: [H2, battery]
```
No errors and contents of the corresponding dataframes e.g. costs look correct.

The only thing I am unsure is whether to add a carrier to the links for e.g. electrolysis and fuel cells; for instance `AC-H2` or `H2-AC`.

fneum:
Actually `n.lines.carrier` also doesn't exist. So I think we can leave out adding carriers to the links. What do you think @FabianHofmann

FabianHofmann:
Great, I'm just finishing the doc string

> The only thing I am unsure is whether to add a carrier to the links for e.g. electrolysis and fuel cells; for instance `AC-H2` or `H2-AC`.

mmh, any idea how the pypsa-eur-sec does that? If so then we should definitely do it as well

fneum:
PyPSA-Eur-Sec adds carriers `H2 Store`, `H2 Electrolysis`, `H2 Fuel Cell`, `battery`, `battery charger`, `battery_discharger`. We could do the same.

FabianHofmann:
> Actually `n.lines.carrier` also doesn't exist. So I think we can leave out adding carriers to the links. What do you think @FabianHofmann

Just had a look at pypsa-eur-sec and there both directions have names. 
* Battery: "battery discharger" and "battery charger" 
* Hydrogen: "H2 Electrolysis" and "H2 Fuel Cell"

I'd say let's do the same (perhaps same capital letter convention)

FabianHofmann:
oh just so you wrote the same :) should reload more often

fneum:
BTW, I think it is the right decision to add this `add_extra_components` rule also regarding the re-integration of the PyPSA-Eur-Sec offspring at some point. I'd stick to lower case for more consistency.

FabianHofmann:
> BTW, I think it is the right decision to add this `add_extra_components` rule also regarding the re-integration of the PyPSA-Eur-Sec offspring at some point. I'd stick to lower case for more consistency.

totally agree

FabianHofmann:
@fneum could you double check the docstring and properly integrate it in the doc? I think this is a small thing for you

scripts/add_extra_components.py
```diff
@@ -0,0 +1,167 @@
...
+        dicountrate:
+        emission_prices:
+
+    electricity:
+        max_hours:
+        marginal_cost:
+        capital_cost:
+        extendable_carriers:
+            StorageUnit:
+            Stores:
```
fneum:
```suggestion
            Store:
```

scripts/add_extra_components.py
```diff
@@ -0,0 +1,167 @@
...
+Outputs
+-------
+
+- ``networks/{network}_s{simpl}_{clusters}_ec.nc``:
+
+
+Description
+-----------
+
+The rule :mod:`add_extra_components` attaches additional artificial components, specified by the configuration file, to the clustered and simplyfied network. Therefore it converts networks of kind `networks/{network}_s{simpl}_{clusters}.nc`` to ``networks/{network}_s{simpl}_{clusters}_ec.nc``, which in contrast to the previous contain
```
fneum:
```suggestion
The rule :mod:`add_extra_components` attaches additional components to the clustered and simplified network. These can be configured in the `config.yaml` at `electricity: extendable_carriers: `. It further processes `networks/{network}_s{simpl}_{clusters}.nc`` to build ``networks/{network}_s{simpl}_{clusters}_ec.nc``, which in contrast to the former contain
```

FabianHofmann:
great!

PR #102, opened 2019-11-20
Add logging to logfiles to all snakemake workflow scripts.

euronion:
Until now only a few scripts actually wrote their logging information to the `log/` directory.
This PR unifies the behaviour and adds writing to a logfile in `log/` for all scripts in the snakemake workflow.

### DIFF
--- a/Snakefile
+++ b/Snakefile
@@ -363,7 +363,7 @@ def input_plot_p_nom_max(wildcards):
 rule plot_p_nom_max:
     input: input_plot_p_nom_max
     output: "results/plots/{network}_s{simpl}_cum_p_nom_max_{clusters}_{technology}_{country}.{ext}"
-    log: "logs/plot_p_nom_max/{network}_s{simpl}_{clusters}_ec_l{ll}_{opts}_{country}.log"
+    log: "logs/plot_p_nom_max/{network}_s{simpl}_{clusters}_{technology}_{country}_{ext}.log"
     script: "scripts/plot_p_nom_max.py"
 
 rule build_country_flh:
--- a/Snakefile
+++ b/Snakefile
@@ -347,7 +347,7 @@ def input_make_summary(w):
 rule make_summary:
     input: input_make_summary
     output: directory("results/summaries/{network}_s{simpl}_{clusters}_ec_l{ll}_{opts}_{country}")
-    log: 'logs/make_summary/{network}_s{simpl}_{clusters}_ec_l{ll}_{opts}_{country}.log",
+    log: "logs/make_summary/{network}_s{simpl}_{clusters}_ec_l{ll}_{opts}_{country}.log",
     script: "scripts/make_summary.py"
 
 rule plot_summary:
--- a/Snakefile
+++ b/Snakefile
@@ -326,7 +326,7 @@ rule plot_network:
     output:
         only_map="results/plots/{network}_s{simpl}_{clusters}_ec_l{ll}_{opts}_{attr}.{ext}",
         ext="results/plots/{network}_s{simpl}_{clusters}_ec_l{ll}_{opts}_{attr}_ext.{ext}"
-    log: "logs/plot_network/{network}_s{simpl}_{clusters}_ec_l{ll}_{opts}.log"
+    log: "logs/plot_network/{network}_s{simpl}_{clusters}_ec_l{ll}_{opts}_{attr}_{ext}.log"
     script: "scripts/plot_network.py"
 
 def input_make_summary(w):
@@ -353,7 +353,7 @@ rule make_summary:
 rule plot_summary:
     input: "results/summaries/{network}_s{simpl}_{clusters}_ec_l{ll}_{opts}_{country}"
     output: "results/plots/summary_{summary}_{network}_s{simpl}_{clusters}_ec_l{ll}_{opts}_{country}.{ext}"
-    log: "logs/plot_summary/{network}_s{simpl}_{clusters}_ec_l{ll}_{opts}_{country}.log"
+    log: "logs/plot_summary/{summary}_{network}_s{simpl}_{clusters}_ec_l{ll}_{opts}_{country}_{ext}.log"
     script: "scripts/plot_summary.py"
 
 def input_plot_p_nom_max(wildcards):
--- a/Snakefile
+++ b/Snakefile
@@ -48,6 +48,7 @@ if not config.get('tutorial', False):
 if config['enable'].get('retrieve_databundle', True):
     rule retrieve_databundle:
         output:  expand('data/bundle/{file}', file=datafiles)
+        log: "logs/retrieve_databundle.log"
         script: 'scripts/retrieve_databundle.py'
 
 rule build_powerplants:
--- a/Snakefile
+++ b/Snakefile
@@ -53,6 +53,7 @@ rule build_powerplants:
         base_network="networks/base.nc",
         custom_powerplants="data/custom_powerplants.csv"
     output: "resources/powerplants.csv"
+    log: "logs/build_powerplants.log"
     threads: 1
     resources: mem=500
     # group: 'nonfeedin_preparation'
@@ -72,6 +73,7 @@ rule base_network:
         offshore_shapes='resources/offshore_shapes.geojson',
         europe_shape='resources/europe_shape.geojson'
     output: "networks/base.nc"
+    log: "logs/base_network.log"
     benchmark: "benchmarks/base_network"
     threads: 1
     resources: mem=500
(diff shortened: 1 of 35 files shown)


Snakefile
```diff
@@ -311,6 +326,7 @@ rule plot_network:
     output:
         only_map="results/plots/{network}_s{simpl}_{clusters}_ec_l{ll}_{opts}_{attr}.{ext}",
         ext="results/plots/{network}_s{simpl}_{clusters}_ec_l{ll}_{opts}_{attr}_ext.{ext}"
+    log: "logs/plot_network/{network}_s{simpl}_{clusters}_ec_l{ll}_{opts}.log"
```
```suggestion
    log: "logs/plot_network/{network}_s{simpl}_{clusters}_ec_l{ll}_{opts}_{attr}_{ext}.log"
```
due to occuring
```
SyntaxError:

Not all output, log and benchmark files of rule plot_network
contain the same wildcards. This is crucial though, in order
to avoid that two or more jobs write to the same file.
```

Snakefile
```diff
@@ -331,11 +347,13 @@ def input_make_summary(w):
 rule make_summary:
     input: input_make_summary
     output: directory("results/summaries/{network}_s{simpl}_{clusters}_ec_l{ll}_{opts}_{country}")
+    log: "logs/make_summary/{network}_s{simpl}_{clusters}_ec_l{ll}_{opts}_{country}.log",
     script: "scripts/make_summary.py"
 
 rule plot_summary:
     input: "results/summaries/{network}_s{simpl}_{clusters}_ec_l{ll}_{opts}_{country}"
     output: "results/plots/summary_{summary}_{network}_s{simpl}_{clusters}_ec_l{ll}_{opts}_{country}.{ext}"
+    log: "logs/plot_summary/{network}_s{simpl}_{clusters}_ec_l{ll}_{opts}_{country}.log"
```
```suggestion
    log: "logs/plot_summary/{summary}_{network}_s{simpl}_{clusters}_ec_l{ll}_{opts}_{country}_{ext}.log"
```

Snakefile
```diff
@@ -345,6 +363,7 @@ def input_plot_p_nom_max(wildcards):
 rule plot_p_nom_max:
     input: input_plot_p_nom_max
     output: "results/plots/{network}_s{simpl}_cum_p_nom_max_{clusters}_{technology}_{country}.{ext}"
+    log: "logs/plot_p_nom_max/{network}_s{simpl}_{clusters}_ec_l{ll}_{opts}_{country}.log"
```
```suggestion
    log: "logs/plot_p_nom_max/{network}_s{simpl}_{clusters}_{technology}_{country}_{ext}.log"
```

fneum:
I think you forgot 2 code suggestions.

euronion:
Thanks!

Strange. I was sure I had included all of them. (CI just takes too long to wait for it to finish.)

fneum:
note to self: travis ci only builds if PR is mergeable: https://docs.travis-ci.com/user/pull-requests/#my-pull-request-isnt-being-built

euronion:
> note to self: travis ci only builds if PR is mergeable: https://docs.travis-ci.com/user/pull-requests/#my-pull-request-isnt-being-built

Very handy information, thx! I was wondering myself as well, why travis was playing lazy.

euronion:
Ok, I hope I fixed it.

## Cause
The problem was that to access the `snakemake.log` one as with `snakemake.input` has to use `snakemake.log[0]` even if it is just one element.

## Additional features

* The output from logging is printed to STDOUT/STDERR *and* the log file specified by `snakemake.log` (instead of only redirecting them to the log file through the `filename` keyword).
* I added `snakemake.config['logging_format']` to `config.default.yaml`, which allows the user to specify the logging message format

## Caveats

* All files with `from _helpers import progress_retrieve` are not using the log file.

There is some collission, presumably with the `progressbar` module, causing the redirect
to file not to work. Instead, only empty log files are created.
Affected are the three scripts `retrieve_natura_raster.py`,  `retrieve_databundle.py` and `retrieve_cutout.py`.

* I didn't test the effects on `plot_network.py`, which also imports the `_helpers` module and might be affected by the same problem.

Since the `progressbar` module is only needed in one function, I moved the import statement there as import on demand. Hope this excludes the `plot_network.py` script from this weird behaviour.

scripts/retrieve_databundle.py
```diff
@@ -28,22 +28,36 @@
...
     else:
         url = "https://zenodo.org/record/3517935/files/pypsa-eur-data-bundle.tar.xz"
 
     tarball_fn = "./bundle.tar.xz"
 
+    logger.info(f"Downloading databundle from '{url}'.")
     progress_retrieve(url, tarball_fn)
 
-    tarfile.open(tarball_fn).extractall('./data')
+    to_fn = "./data"
```
FabianHofmann: A question about this here: Why doesn't the path break in windows? I thought path were separated by '/'. Perhaps saver should be to rather using "data"?

euronion: Yep, that's not a good style there; it still works due to [this tricky Python feature](https://docs.python.org/3/library/os.html#os.altsep), but is not very stable.

I'll quickly fix it.

FabianHofmann:
It seems that the comment I wrote yesterday wasn't submitted, so again:

Why not putting a function in `_helpers` such as 
```
def set_basic_config(logging, snakemake, skip_handlers=False):
    kwargs = {'format': snakemake.config['logging_format'],  'level': snakemake.config['logging_level']}
    if not skip_handlers:
        kwargs.update({'handlers': [logging.FileHandler(snakemake.log[0]), logging.StreamHandler()]})
    logging.basicConfig(kwargs)
                        
```
this could save a lot of space in the scripts

euronion:
@fneum The issue is caused by the `powerplantmatching` module, I already created a PR to fix it (https://github.com/FRESNA/powerplantmatching/pull/15).

All the scripts suffering from the problem either directly import `powerplantmatching` or import it indirectly via `add_electricity`.

A **non-viable** workaround is to move `import logging` and the `logging.basicConfig(...)` before the `powerplantmatching` import. But this causes significant code refractoring.

I suggest we leave this part as is until the new ppm version fixes it.

Good suggestion @FabianHofmann . I'll include it later this afternoon.

fneum:
Ok, so wait to merge until https://github.com/FRESNA/powerplantmatching/pull/15 included in powerplantmatching release? Would there be any changes remaining in PyPSA-Eur then?

FabianHofmann:
@fneum I'd say so, pypi version is already out. conda might take a couple of minutes

fneum:
Ok, logs are complete now until `solve_network` and excluding `retrieve*` (which is ok). We can merge once
- `- powerplantmatching>=0.4.3` in `environment.*`
- @euronion included the function `set_basic_config` in `scripts/_helpers.py`

FabianHofmann:
nice! the conda feedstock is updated, but still not available for installing, guess it takes some more minutes

coroa:
> ```python
> def set_basic_config(*logging*, snakemake, skip_handlers=False):
>     kwargs = {'format': snakemake.config['logging_format'],  'level': snakemake.config['logging_level']}
>     if not skip_handlers:
>         kwargs.update({'handlers': [logging.FileHandler(snakemake.log[0]), logging.StreamHandler()]})
>     logging.basicConfig(kwargs)
>                         
> ```

@euronion The  `logging` package should not be a parameter

euronion:
Ok, I hope I covered everything.

Building on @FabianHofmann suggestion I added some smartness to the function, e.g. if no 'log' parameter is provided in the `Snakefile`, a filename is constructed using the rule's name.

I also changed the `config.yaml` entry from
```
logging_level: xyz
logging_format: xyz
```

to 
```
logging:
  level: xyz
  format: xyz
```

and ensured backwards compatability via the function.

An update of the documentation is also included.

euronion:
Just saw @coroa's comment.

> @euronion The `logging` package should not be a parameter

Would you suggest to `import logging` in the `_helper.py`?
I am not sure what happens to the configuration of the root logger then.

scripts/_helpers.py
```diff
@@ -4,13 +4,48 @@
 
 import pypsa
 
-from add_electricity import load_costs, update_transmission_costs
+def configure_logging(logging, snakemake, skip_handlers=False):
```
coroa: `logging` is a top-level module. There is no reason for it to be passed as an argument.

coroa:
> Just saw @coroa's comment.
> 
> > @euronion The `logging` package should not be a parameter
> 
> Would you suggest to `import logging` in the `_helper.py`?
> I am not sure what happens to the configuration of the root logger then.

Yes, import is fine. `logging.basicConfig` always sets the root logger. Doesn't matter where you call it from.

euronion:
Ok, thanks for reassuring on the behaviour of `logging.basicConfig` @coroa .
All changes made.

AFAICT it runs w/o problems locally (only tested for all scripts until the solving starts).

Could you do a final check on the log files written @fneum ?

PR #106, opened 2019-11-28
Add H2 pipelines in `add_extra_components`

fneum:
- adds h2 pipelines alongside clustered network connections
- h2 storage has to be modelled as store!
- partly inspired by recent commit in pypsa-eur-sec
- cost assumptions included

This is a first, simplified issue. Future tasks for hydrogen networks:
- **storage capacity:** is modelling gas network as links sufficient or should it be some link-store-link combination or storageunit with 2 buses?
- **bidirectional flow:** is operation in both directions easily possible? or do we need 2 links with each `p_min_pu=0`? 
- **recompression:** is it necessary to consider energy demand for recompression? or marginal cost of recompression?


fneum:
Uff, that was not what I wanted. Sorry. I attempted to delete the `databundle.tar.xz`; will try with another branch...

FabianHofmann:
no worry, seems the new branch includes everything necessary

PR #107, opened 2019-12-01
Introduce mocksnakemake which acutally parses Snakefile

FabianHofmann:
The new function `_helpers.mocksnakemake` (i'm open for another name) creates a `snakemake` object which mimics the actual snakemake object produced by snakemake-workflow. Therefore you are able to run all scripts within a (I)python terminal (or just by calling `python <script-name>`) which makes debugging and creating scripts much easier. 

It parses the `Snakefile` and sets all path for input, output, log etc. The function itself is a bit hacky but stable for all scripts. 

Additionally it creates directories for log and output files in case they are not existent.

# Additional Fixes in

* `plot_p_nom_max`
* `plot_network`
* `build_country_flh`
* `make_summary` for tutorial network


config.default.yaml
```diff
@@ -223,7 +223,9 @@ plotting:
   storage_techs: ["hydro+PHS", "battery", "H2"]
   load_carriers: ["AC load"]
   AC_carriers: ["AC line", "AC transformer"]
-  link_carriers: ["DC line", "Converter AC-DC"] 
+  link_carriers: ["DC line", "Converter AC-DC"]
+  heat_links: []
+  heat_generators: []
```
fneum: These were removed a while ago.

config.tutorial.yaml
```diff
@@ -201,7 +201,9 @@ plotting:
   storage_techs: ["hydro+PHS", "battery", "H2"]
   load_carriers: ["AC load"]
   AC_carriers: ["AC line", "AC transformer"]
-  link_carriers: ["DC line", "Converter AC-DC"] 
+  link_carriers: ["DC line", "Converter AC-DC"]
+  heat_links: []
+  heat_generators: []
```
fneum: These were removed a while ago.

data/links_p_nom.csv
```diff
@@ -1,20 +1,18 @@
...
-SACOI 2,"France- Lucciana 42°31′40″N 9°26′59″E﻿ / ﻿42.52778°N 9.44972°E"," Codrongianos- Italy 40°39′7″N 8°42′48″E﻿ / ﻿40.65194°N 8.71333°E",422(118/304),200,300.0,1992,Thyr,Multiterminal scheme Supplier: Alstom,,9.449722222222222,42.52777777777778,8.679351,40.65765
+SACOI 2,Italy - Suvereto 43°3′10″N 10°41′42″E﻿ / ﻿43.05278°N 10.69500°E ( before 1992: Italy - San Dalmazio 43°15′43″N 10°55′05″E﻿ / ﻿43.26194°N 10.91806°E),"France- Lucciana 42°31′40″N 9°26′59″E﻿ / ﻿42.52778°N 9.44972°E ; Codrongianos, Italy 40°39′7″N 8°42′48″E﻿ / ﻿40.65194°N 8.71333°E",422(118/304),200,300.0,1992,Thyr,Multiterminal scheme Supplier: Alstom,,10.695,43.05277777777778,9.449722222222222,42.52777777777778
 Skagerrak 3,Denmark - Tjele 56°28′44″N 9°34′1″E﻿ / ﻿56.47889°N 9.56694°E,Norway - Kristiansand 58°15′36″N 7°53′55″E﻿ / ﻿58.26000°N 7.89861°E,230(130/100),350,440.0,1993,Thyr,Supplier: Nexans ABB,[34],9.566944444444445,56.47888888888889,7.898611111111111,58.26
```
fneum: Were all of these intentional changes?

Snakefile
```diff
@@ -140,6 +140,7 @@ else:
 if config['enable'].get('build_natura_raster', False):        
     rule build_natura_raster:
         input: 
+            cutout_dir = "cutouts",
```
fneum: Why is this necessary?

scripts/build_shapes.py
```diff
@@ -79,6 +79,8 @@
 from shapely.ops import cascaded_union
 
 import pycountry as pyc
+from _helpers import mocksnakemake
```
fneum: Duplicated import?

scripts/prepare_network.py
```diff
@@ -180,13 +180,9 @@ def average_every_nhours(n, offset):
 if __name__ == "__main__":
     # Detect running outside of snakemake and mock snakemake for testing
```
fneum: Delete? Small thing, but this is not yet consistent.

scripts/solve_network.py
```diff
@@ -92,6 +92,7 @@
 pypsa.pf.logger.setLevel(logging.WARNING)
 
 from vresutils.benchmark import memory_logger
+from _helpers import mocksnakemake
```
fneum: Is there a reason why you don't import it in the `if 'snakemake' not in globals():` block as elsewhere?

test/config.test1.yaml
```diff
@@ -201,7 +201,9 @@ plotting:
   storage_techs: ["hydro+PHS", "battery", "H2"]
   load_carriers: ["AC load"]
   AC_carriers: ["AC line", "AC transformer"]
-  link_carriers: ["DC line", "Converter AC-DC"] 
+  link_carriers: ["DC line", "Converter AC-DC"]
+  heat_links: []
+  heat_generators: []
```
fneum: These were removed a while ago.

config.default.yaml
fneum: Probably we should remove
https://github.com/PyPSA/pypsa-eur/blob/036d44418da5db156741249cc7e2323788fe508c/scripts/_helpers.py#L67-L68 then.

scripts/_helpers.py
```diff
@@ -150,3 +151,88 @@ def dlProgress(count, blockSize, totalSize):
 
     urllib.request.urlretrieve(url, file, reporthook=dlProgress)
 
+
+def mocksnakemake(rulename, **wildcards):
```
euronion:
```suggestion
def mock_snakemake(rulename, **wildcards):
```

Name suggestion. Calling it that way includes an imperative.

config.default.yaml
FabianHofmann: yes, makes totally sense

data/links_p_nom.csv
FabianHofmann: More or less, these came when rerunning the `prepare_links_p_nom` script, this the updated version of the csv file. But perhaps it is better to have only code changes for this PR and update the data later

scripts/_helpers.py
```diff
@@ -150,3 +151,88 @@ def dlProgress(count, blockSize, totalSize):
...
+            return
+        # for mildy hacky input functions (like make_summary):
+        if len(smfiles) == 1 and callable(smfiles[0]):
+            new_input_files = smfiles[0](wc)
+            # overwrite smfiles with extracted files from function
+            smfiles = sm.io.InputFiles()
+            for index, p in enumerate(new_input_files):
+                smfiles.insert(index, p)
+        # now iterate over each item and make path an absolut path
+        files = sm.io.InputFiles()
```
euronion:
```suggestion
        files = getattr(sm.io, type(smfiles))()
```

To keep the distinction between `InputFiles`, `OutputFiles` and others.

scripts/_helpers.py
```diff
@@ -150,3 +151,88 @@ def dlProgress(count, blockSize, totalSize):
...
+            else:
+                files.insert(index, p)
+            if key is not None:
+                files.add_name(key)
+        return files
+
+    Input = make_io_accessable(rule.input)
+    Output = make_io_accessable(rule.output)
+    if not rule.log:
+        rule.log.insert(0, f"logs/{rule.name}.log")
```
euronion:
```suggestion
```

Since you only want to mock the `snakemake` object, I would not include additional assumptions in here.
Especially for debugging, additions to the `snakemake` object here will make it more complicated.

scripts/_helpers.py
```diff
@@ -150,3 +151,88 @@ def dlProgress(count, blockSize, totalSize):
...
+    Log = make_io_accessable(rule.log)
+    # create log dir if not existent
```
euronion: As above, I suggest to remove any actions outstide the `snakemake` object mocking and remove the whole (optional) folder creation loop.

Snakefile
FabianHofmann: because the cutout directory was hard coded in `build_natura_raster` before. Only like this the path can be set correctly by `mocksnakemake`

scripts/build_shapes.py
FabianHofmann: yes!

scripts/build_natura_raster.py
```diff
@@ -43,17 +43,22 @@
 import atlite
 import geokit as gk
 
+
 def determine_cutout_xXyY(cutout_name):
-    cutout = atlite.Cutout(cutout_name, cutout_dir="cutouts")
+    cutout = atlite.Cutout(cutout_name, cutout_dir=snakemake.input.cutout_dir)
```
coroa: Like this one.

scripts/build_natura_raster.py
```diff
@@ -43,17 +43,22 @@
...
 
+
 if __name__ == "__main__":
+    if 'snakemake' not in globals():
+        from _helpers import mocksnakemake
+        snakemake = mocksnakemake('build_natura_raster') #has to be enabled
     configure_logging(snakemake)
 
-    cutout_names = np.unique([res['cutout'] for res in snakemake.config['renewable'].values()])
+    cutout_names = {res['cutout'] for res in snakemake.config['renewable'].values()}
```
coroa: Or this one?

scripts/build_country_flh.py
```diff
@@ -198,7 +174,7 @@ def build_aggregate(flh, countries, areamatrix, breaks, p_area, fn):
     # Use GLAES to compute available potentials and the transition matrix
     paths = dict(snakemake.input)
 
-    init_globals(bounds, dx, dy, config, paths)
+    init_globals(bounds.xXyY, dx, dy, config, paths)
```
coroa: Or this one?

scripts/prepare_network.py
FabianHofmann: yep

scripts/solve_network.py
FabianHofmann: no actually not, you are right one should keep it consistent having it always in the `if` block

scripts/make_summary.py
```diff
@@ -471,6 +471,15 @@ def to_csv(dfs):
 
 
 if __name__ == "__main__":
+    if 'snakemake' not in globals():
+        from _helpers import mocksnakemake
+        snakemake = mocksnakemake('make_summary', network='elec', simpl='',
+                           clusters='5', ll='copt', opts='Co2L-24H', country='all')
+        network_dir = os.path.join('..', 'results', 'networks')
```
euronion:
```suggestion
        network_dir = os.path.join('results', 'networks')
```

Comparing with the original code, the `'..'` seem misplaced.

scripts/plot_network.py
```diff
@@ -17,7 +17,8 @@
 
 import logging
 logger = logging.getLogger(__name__)
-from _helpers import load_network, aggregate_p, aggregate_costs, configure_logging
+from _helpers import (load_network, aggregate_p, aggregate_costs,
+                      configure_logging, mocksnakemake)
```
euronion:
```suggestion
                      configure_logging)
```

to stay consistent with the other files.

scripts/plot_network.py
```diff
@@ -236,30 +240,21 @@ def split_costs(n):
...
-        snakemake = Dict()
-        snakemake = MockSnakemake(
-            path='..',
-            wildcards=dict(network='elec', simpl='', clusters='90', lv='1.25', opts='Co2L-3H', attr='p_nom', ext="pdf"),
-            input=dict(network="results/networks/{network}_s{simpl}_{clusters}_lv{lv}_{opts}.nc",
-                    tech_costs="data/costs.csv"),
-            output=dict(only_map="results/plots/{network}_s{simpl}_{clusters}_lv{lv}_{opts}_{attr}.{ext}",
-                        ext="results/plots/{network}_s{simpl}_{clusters}_lv{lv}_{opts}_{attr}_ext.{ext}")
-        )
+        snakemake = mocksnakemake('plot_network', network='elec', simpl='',
```
euronion:
```suggestion
        from _helpers import mocksnakemake
        snakemake = mocksnakemake('plot_network', network='elec', simpl='',
```

to stay consistent with the other files

scripts/retrieve_databundle.py
```diff
@@ -36,7 +36,6 @@
 import tarfile
 
 if __name__ == "__main__":
-
     configure_logging(snakemake) # TODO Make logging compatible with progressbar (see PR #102)
```
euronion:
```suggestion
    # Detect running outside of snakemake and mock snakemake for testing
    if 'snakemake' not in globals():
        from _helpers import mocksnakemake
        snakemake = mocksnakemake('retrieve_databundle')
    configure_logging(snakemake) # TODO Make logging compatible with progressbar (see PR #102)
```

Snakefile
coroa: Agreed, shouldn't be necessary.

scripts/_helpers.py
```diff
@@ -150,3 +151,88 @@ def dlProgress(count, blockSize, totalSize):
...
+        keyword arguments fixing the wildcards. Only necessary if wildcards are
+        needed.
+    """
+    import snakemake as sm
+    import os
+    import logging
+    from pypsa.descriptors import Dict
+    from os.path import abspath
+
+    os.chdir('..')
```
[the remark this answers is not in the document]
FabianHofmann: yes, good idea!

scripts/_helpers.py
FabianHofmann: yes, that's perhaps better, I'll do that in a seperate commit

scripts/make_summary.py
FabianHofmann: no it is actually not, this effect comes again from having paths hard coded in the script. The `mock_snakemake` function expects to be executed in the scripts folder. That's why all paths have to be rearranged in the function. And since this path here doesn't come from the `Snakefile`, it has to be adapted like this

scripts/_helpers.py
FabianHofmann: so you'd suggest having a argument as `forward_mkdir` which if set to true enables directory creation?

scripts/build_natura_raster.py
FabianHofmann: You're right, this came on the run while going through the scripts.

data/links_p_nom.csv
fneum: yes, should be part of a separate PR.

scripts/build_country_flh.py
FabianHofmann: as well as this, but it is necessary for running the script. I can however create an issue for that

scripts/_helpers.py
```diff
@@ -150,3 +148,89 @@ def dlProgress(count, blockSize, totalSize):
...
+        needed.
+    """
+    import snakemake as sm
+    import os
+    import logging
+    from pypsa.descriptors import Dict
+    from os.path import abspath
+
+    old_wd = os.getcwd()
+    os.chdir(base_dir)
```
FabianHofmann: @euronion the `base_dir` is not defined. A good suggestion for determining it (rather then setting it to '..')

scripts/_helpers.py
euronion: No. I would remove the lines for optionally creating the `log` dir altogether.
If `Snakemake` does not do it, then neither should the `mock_snackmake`.

scripts/make_summary.py
euronion: I see. Makes sense now.

scripts/_helpers.py
euronion: I think passing it as a function argument is reasonable:

```
    def mocksnakemake(rulename, base_dir="..", **wildcards):
```

scripts/_helpers.py
FabianHofmann: I'm wondering, `Snakemake` actually creates missing directories in case there not there. For debugging it is really convenient to create those dirs as well.

euronion: Ah, you are right! Doing a 180, yes creating the directories should happen in this case!

If there are no objections against using `pathlib` and we are assuming Python > 3.4 (we are, aren't we?) then you can also simply write:

```
from pathlib import Path
for file in list(Log) + list(Output):
        Path(file).parent.mkdir(parents=True, exist_ok=True)

Comment (FabianHofmann) [reply scripts/_helpers.py]: yes, I like that!

File: Snakefile (hunk 4 above)
Comment (FabianHofmann) [reply Snakefile]: is reverted now

File: data/links_p_nom.csv (hunk 3 above)
Comment (FabianHofmann) [reply data/links_p_nom.csv]: is reverted now

File: scripts/_helpers.py (hunk 21 above)
Comment (FabianHofmann) [reply scripts/_helpers.py]: `base_dir` is now set to  `Path(__file__).parent.parent`. You agree with that?

File: scripts/_helpers.py [hunk 22]
```diff
@@ -25,11 +27,14 @@ def configure_logging(snakemake, skip_handlers=False):
     kwargs.setdefault("level", "INFO")
 
     if skip_handlers is False:
+        fallback_path = Path(__file__).parent.parent.joinpath('logs', f"{snakemake.rule}.log")
+        logfile = snakemake.log.get('python', snakemake.log[0] if snakemake.log
+                                    else fallback_path)
```
FabianHofmann: @euronion you agree with that? for debugging in scripts folder, it is necessary to set the path independent of the cwd

scripts/_helpers.py
euronion: No, this should not work the way you want it to, as you cannot got beyond your anchor, in this case `__name__`, see [the pathlib doc](https://docs.python.org/3/library/pathlib.html#pathlib.PurePath.parent).

```
p = Path(__file__)  # "_helpers"
p.parent            # "."
p.parent.parent     # "."
```

You can either hard-code `base_dir = ".."`, which I think is os-portable and the easier to understand version.
Or you have to resolve the absolute path first:

```
p = Path(__file__) # "_helpers"
p = p.absolute()   # "/home/user/pypsa-eur/scripts/_helpers"
p.parent           # "/home/user/pypsa-eur/scripts"
p.parent.parent    # "/home/user/pypsa-eur"
```

scripts/_helpers.py
euronion: No, I can not agree with this change.

* First: See the comment above, the second `.parent` in your code has not effect.
* Second: I do not fully understand your comment, as it still depending on the `cwd` (but using it as a relative anchor).
* Also: If you really use the parents parent `..`, then it breaks the original behaviour for the case of neither `snakemake.log.python`  nor `snakemake.log` being defined. Instead of `logs/xyz.log` it would use `../logs/xyz.log` as a path.

If you are doing debugging and encounter this problem, you could still `skip_handlers` and add them manually to the `logging` root logger.

scripts/_helpers.py
coroa: @FabianHofmann : The missing key information is that, if you execute a script as `python relative/path/test.py`, the `__file__` variable in `test.py` will only hold `relative/path/test.py` and not its absolute path, like it is if you `import test` or use `run` in ipython.

I think it makes sense to use `__file__` to locate the correct directory, but am split between
1. `Path(__file__).resolve().parent.parent`, and
2. `Path(__file__).parent / ".."`

scripts/_helpers.py
```diff
@@ -150,3 +152,81 @@ def dlProgress(count, blockSize, totalSize):
...
+            if isinstance(p, list):
+                if len(p) == 1:
+                    p = p[0]
+            if isinstance(p, str):
+                files.insert(index, sm.io.apply_wildcards(abspath(p), wc))
+            else:
+                files.insert(index, p)
+            if key is not None:
+                files.add_name(key)
+        return files
```
coroa: Seeing now, the difficulty with which you are trying to patch all available paths to work with whatever path one started from. I fear that this code is prone to break soon, as you noticed with the log file fallbacks for example.

I think it's better to take a radically different path and make clear, that if you want to run a script from a python interpreter one has to start the interpreter from the repository root, where the Snakefile is located, and simplify all of this, drastically.

scripts/_helpers.py
euronion: I am a bit confused now, on my system `resolve()` is insufficient:

```
In [1]: import pathlib

In [2]: p = pathlib.Path(__name__)

In [3]: p
Out[3]: WindowsPath('__main__')

In [4]: p.resolve()
Out[4]: WindowsPath('__main__')

In [5]: p.absolute()
Out[5]: WindowsPath('C:/Users/__main__')
```

If we want to determine the parent of the current working directory, the solution should be
```
pathlib.Path.cwd().parent
```

If we want to locate the great-parent directory of the scripts, then semantically I would suggest to go with @coroa 
```
Path(__file__).aboslute().parent.parent
```

FabianHofmann: ah okay, thank you both for clarifying! I tested it now with to 

```
Path(__file__).parent.joinpath('..', 'logs', f"{snakemake.rule}.log")
```
which works well in for both `mocksnakemake` and normal `snakemake`

coroa: @euronion It's about `__file__` not `__name__`, which is not defined in the interpreter only, well in a file. On an existing file `resolve()` should work.

scripts/_helpers.py
FabianHofmann: @coroa you are right, about the simplification. However, I would like to keep it like that for now, as I have barely time this and next week, and going through the snakemake code takes a lot of time. I'll set an (self-referenced) issue for that as soon as the PR is through.

scripts/_helpers.py
FabianHofmann: I tested the current version, for `snakemake.log.python` and `snakemake.log` being undefined, and it works for both `Snakemake` and `mock_snakemake`. I don't care about the specific implementation, both  

* `Path(__file__).resolve().parent.parent.joinpath('logs', f"{snakemake.rule}.log")`
* `Path(__file__).parent.joinpath('..', 'logs', f"{snakemake.rule}.log")`

I'd go for the second as it is a bit shorter

scripts/_helpers.py
coroa: @FabianHofmann I think you misunderstood. I would propose to throw the towel and reduce everything down to the following and be done with it. You're trying to crawl down a rabbit hole.

```python
def mock_snakemake(rulename, **wildcards):
    """
    This function is expected to be executed in the 'scripts'-directory within
    the snakemake project. It returns a snakemake.script.Snakemake object,
    based on the Snakefile of the upper directory, such that includes
    mostly all necessary information (input, output, log etc.).

    If a rule has wildcards, you have to specify them in **wildcards.

    Parameters
    ----------
    rulename: str
        name of the rule for which the snakemake object should be generated
    **wildcards:
        keyword arguments fixing the wildcards. Only necessary if wildcards are
        needed.
    """
    import snakemake as sm
    import os
    from pypsa.descriptors import Dict
    from os.path import abspath

    for p in sm.SNAKEFILE_CHOICES:
        if os.path.exists(p):
            snakefile = p
            break
    workflow = sm.Workflow(snakefile)
    workflow.include(snakefile)
    rule = workflow.get_rule(rulename)
    wc = Dict(wildcards)

    # create log and output dir if not existent
    for path in chain(rule.log.values(), rule.output.values()):
        Path(path).parent.mkdir(parents=True, exist_ok=True)

    snakemake = sm.script.Snakemake(input=rule.input, output=rule.output,
                                    params=rule.params, wildcards=wc, threads=None,
                                    resources=rule.resources, log=rule.log,
                                    config=workflow.config, rulename=rule.name,
                                    bench_iteration=None)

    return snakemake
```

FabianHofmann: But this is not working, I wished it was as easy. The problem is that the wildcards are not applied here (which made the whole thing a bit tricky). 
I am sure there is an easier way to do this than I did, but it requires going through the snakemake code and look how they create the `snakemake` object.

FabianHofmann: initializing the `snakemake` class is literately just passing objects (see [here](https://github.com/snakemake/snakemake/blob/5bb0ed743c8850532a8d4c485105fefc9cf33bc5/snakemake/script.py#L167))

FabianHofmann: I think I have it:

```python
def mock_snakemake(rulename, **wildcards):
    """
    This function is expected to be executed from the 'scripts'-directory of the
    the snakemake project. It returns a snakemake.script.Snakemake object,
    based on the Snakefile.

    If a rule has wildcards, you have to specify them in **wildcards.

    Parameters
    ----------
    rulename: str
        name of the rule for which the snakemake object should be generated
    **wildcards:
        keyword arguments fixing the wildcards. Only necessary if wildcards are
        needed.
    """
    import snakemake as sm
    import os
    from pypsa.descriptors import Dict
    from snakemake.script import Snakemake
    import logging

    base_dir = Path(__file__).parent.joinpath('..').absolute()
    if Path(os.getcwd()).absolute() != base_dir:
        logging.info(f'Changing directory to repository root {base_dir}')
        os.chdir(base_dir)

    for p in sm.SNAKEFILE_CHOICES:
        if os.path.exists(p):
            snakefile = p
            break
    workflow = sm.Workflow(snakefile)
    workflow.include(snakefile)
    workflow.global_resources = {}
    rule = workflow.get_rule(rulename)
    dag = sm.dag.DAG(workflow, rules=[rule])
    wc = Dict(wildcards)
    job = sm.jobs.Job(rule, dag, wc)
    snakemake = Snakemake(job.input, job.output, job.params, job.wildcards,
                          job.threads, job.resources, job.log,
                          job.dag.workflow.config, job.rule.name, None,)

    # create log and output dir if not existent
    for path in list(rule.log) + list(rule.output):
        Path(path).parent.mkdir(parents=True, exist_ok=True)
    return snakemake
```
does the thing. I know there is a change of the directory in the function itself which is nasty. But it has to be done there as all the relative imports within the scripts would not work otherwise, like (`from _helpers import configure_logging` or `from add_electricity import ...`)

euronion: Ok, so using this function will (in some cases) change my current working directory without changing it back? I am curious to see how this holds up in the real world.

Minor correction of my previous statement, based on @coroa 's comment:
It should be `.resolve()`, not `.absolute()` (which is neither tested nor officially documented). So this
```
base_dir = Path(__file__).parent.joinpath('..').absolute()
    if Path(os.getcwd()).absolute() != base_dir:
```
should really be this:
```
base_dir = Path(__file__).resolve().parents[1]
    if Path.cwd().resolve() != base_dir:
```

FabianHofmann: great, thanks for the suggestion. I pushed the new function, and see where it gets... It really takes  effort to understand `Snakemake` properly. It works well for all scripts, only exception is that the logging sometime behaves weird. Thankful if someone can test it.

coroa: Instead of changing the working directory, it would also be fine to fail with an error from which directory one has to run the code!

FabianHofmann: true, but then (as mentioned above) the relative module imports won't work...

coroa: Ah sorry, I didn't get that. Why is that? Can the relative imports be written differently?

I don't immediately see how, though.

scripts/retrieve_natura_raster.py
```diff
@@ -36,15 +36,17 @@
     if 'snakemake' not in globals():
         from _helpers import mock_snakemake
         snakemake = mock_snakemake('retrieve_natura_raster')
+        to_fn = Path("..", "resources/natura.tiff")
+    else:
+        to_fn = Path("resources/natura.tiff")
```
coroa: What's wrong about `snakemake.output[0]` instead?

FabianHofmann:
Sooo, now I think this should be fine now. Some updates on the function: 

1. the mocking has to be run from the `scripts` directory (assertion included). So being in the `pypsa-eur/scripts` this has to work

         python add_electricity 

2. the function **temporarly** changes the directory to the repo root `pypsa-eur`. This is necesarry for retrieving and setting all necessary `Snakemake` options. 
3. returned the `snakemake` object includes only **absolute paths**, so that the paths for io-files current stay the same independent  of the working directory

scripts/retrieve_natura_raster.py
FabianHofmann: You're too fast :)  See below

FabianHofmann: sorry didn't read correctly, you are right

scripts/_helpers.py
```diff
@@ -150,3 +152,58 @@ def dlProgress(count, blockSize, totalSize):
...
+    """
+    import snakemake as sm
+    import os
+    from pypsa.descriptors import Dict
+    from snakemake.script import Snakemake
+
+    script_dir = Path(__file__).parent.resolve()
+    msg = ('mock_snakemake has to be run from the repository scripts '
+           f'directory {script_dir}')
+    assert Path.cwd().resolve() == script_dir, msg
```
coroa:
```suggestion:-2+1
    assert Path.cwd().resolve() == script_dir, \
        f'mock_snakemake has to be run from the repository scripts directory {script_dir}'
```

coroa: Nitpick

scripts/_helpers.py
```diff
@@ -150,3 +152,58 @@ def dlProgress(count, blockSize, totalSize):
...
+    workflow.global_resources = {}
+    rule = workflow.get_rule(rulename)
+    dag = sm.dag.DAG(workflow, rules=[rule])
+    wc = Dict(wildcards)
+    job = sm.jobs.Job(rule, dag, wc)
+
+    def make_accessable(*ios):
+        for io in ios:
+            for i in range(len(io)):
+                io[i] = os.path.abspath(io[i])
```
coroa: Awesome improvement compared to before!

coroa: Note that, `os.path.join("..", os.path.relpath(io[i]))` pbb works equally well w/o absolute paths

scripts/_helpers.py
FabianHofmann: I resolve this conversation, since it's outdated. Just as a side note:
I tried a lot with importing the modules differently, one approach was to import them as 
```python 
from scripts._helpers import ...
```
But this did not work for the normal Snakemake workflow as the repo root, `pypsa-eur`, is not in the paths (sys.path). Trying to add `pypsa-eur` to sys.path in the beginning of the Snakefile does not affect the execution of a rule, as a different thread process is opened for this.

scripts/plot_p_nom_max.py
```diff
@@ -42,31 +43,17 @@ def cum_p_nom_max(net, tech, country=None):
...
+        snakemake = mock_snakemake('plot_p_nom_max', network='elec', simpl='',
+                                  technology='solar', ext='pdf', clusters= '5,full',
+                                  country= 'all')
     configure_logging(snakemake)
 
     plot_kwds = dict(drawstyle="steps-post")
 
     clusters = snakemake.wildcards.clusters.split(',')
-    techs = snakemake.params.techs
+    techs = snakemake.config['renewable'].keys()
```
coroa:
```suggestion
    techs = snakemake.wildcards.technology
```

scripts/retrieve_cutout.py
```diff
@@ -45,6 +45,12 @@
 from _helpers import progress_retrieve, configure_logging
 
 if __name__ == "__main__":
+    if 'snakemake' not in globals():
+        from _helpers import mock_snakemake
+        snakemake = mock_snakemake('retrieve_cutout')
+        prepath = '..'
```
coroa:
```suggestion
        rootpath = '..'
```

scripts/retrieve_cutout.py
```diff
@@ -45,6 +45,12 @@
...
+    else:
+        prepath = '.'
```
coroa:
```suggestion
        rootpath = '.'
```

scripts/retrieve_cutout.py
```diff
@@ -54,13 +60,13 @@
         url = "https://zenodo.org/record/3517949/files/pypsa-eur-cutouts.tar.xz"
 
     # Save location
-    tarball_fn = Path("./cutouts.tar.xz")
+    tarball_fn = Path(f"{prepath}/cutouts.tar.xz")
```
coroa:
```suggestion
    tarball_fn = Path(f"{rootpath}/cutouts.tar.xz")
```

scripts/retrieve_cutout.py
```diff
@@ -54,13 +60,13 @@
...
 
     logger.info(f"Downloading cutouts from '{url}'.")
     progress_retrieve(url, tarball_fn)
 
     logger.info(f"Extracting cutouts.")
-    tarfile.open(tarball_fn).extractall()
+    tarfile.open(tarball_fn).extractall(path = prepath)
```
coroa:
```suggestion
    tarfile.open(tarball_fn).extractall(path=rootpath)
```

scripts/retrieve_databundle.py
```diff
@@ -36,7 +36,13 @@
 import tarfile
 
 if __name__ == "__main__":
-
+    # Detect running outside of snakemake and mock snakemake for testing
+    if 'snakemake' not in globals():
+        from _helpers import mock_snakemake
+        snakemake = mock_snakemake('retrieve_databundle')
+        pathprefix = '..'
```
coroa:
```suggestion
        rootpath = '..'
```

scripts/retrieve_databundle.py
```diff
@@ -36,7 +36,13 @@
...
+    else:
+        pathprefix = '.'
```
coroa:
```suggestion
        rootpath = '.'
```

scripts/retrieve_databundle.py
```diff
@@ -45,15 +51,15 @@
         url = "https://zenodo.org/record/3517935/files/pypsa-eur-data-bundle.tar.xz"
 
     # Save locations
-    tarball_fn = Path("./bundle.tar.xz")
-    to_fn = Path("./data")
+    tarball_fn = Path(f"{pathprefix}/bundle.tar.xz")
```
coroa:
```suggestion
    tarball_fn = Path(f"{rootpath}/bundle.tar.xz")
```

scripts/retrieve_databundle.py
```diff
@@ -45,15 +51,15 @@
...
+    to_fn = Path(f"{pathprefix}/data")
```
coroa:
```suggestion
    to_fn = Path(f"{rootpath}/data")
```

scripts/retrieve_natura_raster.py
```diff
@@ -33,16 +33,14 @@
...
     url = "https://zenodo.org/record/3518215/files/natura.tiff"
 
     logger.info(f"Downloading natura raster from '{url}'.")
-    progress_retrieve(url, to_fn)
-    
-    logger.info(f"Natura raster available as '{to_fn}'.")
\ No newline at end of file
+    progress_retrieve(url, snakemake.output[0])
+
+    logger.info(f"Natura raster available as '{snakemake.output[0]}'.")
```
coroa:
```suggestion
    logger.info(f"Natura raster available as '{snakemake.output[0]}'.")
```

coroa: Return at end of file

scripts/_helpers.py
FabianHofmann: nitpick included

scripts/_helpers.py
coroa: Let's say maybe. Not probably.

scripts/plot_p_nom_max.py
FabianHofmann: this seems to be conflicting with the code, which wants techs to be a list. However then the question rises why the technology wildcard is there in the first place... I see two options: 
1. keeping it as it was and delete the technology wildcard or
2. change the code to plot only one technology which is defined by the wildcard

FabianHofmann: If you don't mind I'll keep it as it is and raise a small issue for later

coroa: True.

I had used it w/o the technology wildcard but with a techs [param](https://snakemake.readthedocs.io/en/stable/snakefiles/rules.html#non-file-parameters-for-rules), only that I forgot to add it to the `Snakefile` as I was only using it with my mocksnakemake.

I think it would be good if the snakemake.output matches with what is actually created. This is only possible with the wildcard, thus:
```suggestion
    techs = [snakemake.wildcards.tech]
```

But let's call it `tech` as we did for the `build_renewable_profile` rule. So changes for `Snakefile` are also needed.

scripts/plot_p_nom_max.py
```diff
@@ -75,14 +62,15 @@ def cum_p_nom_max(net, tech, country=None):
...
     fig, axes = plt.subplots(1, len(techs))
 
-    for cluster in clusters:
-        net = pypsa.Network(getattr(snakemake.input, cluster))
+    for j, cluster in enumerate(clusters):
+        net = pypsa.Network(snakemake.input[j])
 
         for i, tech in enumerate(techs):
-            cum_p_nom_max(net, tech, country).plot(x="p_max_pu", y="c_p_nom_max", label=cluster, ax=axes[0][i], **plot_kwds)
+            cum_p_nom_max(net, tech, country).plot(x="p_max_pu", y="p_nom_max",
```
coroa:
```suggestion
            cum_p_nom_max(net, tech, country).plot(x="p_max_pu", y="cum_p_nom_max",
```

scripts/plot_p_nom_max.py
coroa: Ah no, due to the clusters list, this is not possible anyway :/

scripts/plot_p_nom_max.py
```diff
@@ -75,14 +62,15 @@ def cum_p_nom_max(net, tech, country=None):
...
+                         label=cluster, ax=axes[i], **plot_kwds)
```
coroa: It should be `axes[0][i]` shouldn't it? It's about the `i`th column.

scripts/plot_p_nom_max.py
```diff
@@ -75,14 +62,15 @@ def cum_p_nom_max(net, tech, country=None):
...
 
     for i, tech in enumerate(techs):
-        ax = axes[0][i]
+        ax = axes[i]
```
coroa: Same.

scripts/plot_p_nom_max.py
FabianHofmann: yep, this makes sense

scripts/plot_p_nom_max.py
FabianHofmann: ah I see what you mean, but then we should include the subplot rows. before the lines of different cluster sizes (and one carrier) were plotted in the same subplot

scripts/plot_p_nom_max.py
coroa: Can you add this before merging? And maybe commit with a `[fixes #112]` tag.

FabianHofmann: yes

scripts/plot_p_nom_max.py
coroa: There's just a single row:
https://github.com/PyPSA/pypsa-eur/pull/107/files/1966cccec6824c86ca46095f230b32094ea1542d#diff-720aa7a680646ab82b6f89980ddb841aL76

coroa: It was more or less the idea to compare how the maximum available capacity changes for different cluster sizes. For this it makes sense to have the different cluster sizes in the same panel and a different panel for each tech (completely different amnts). I don't have an example on my working computer unfortunately.

FabianHofmann: https://github.com/PyPSA/pypsa-eur/issues/112#issuecomment-563300768
This is what comes out now, this is also what you mean or?

FabianHofmann: > There's just a single row:
> https://github.com/PyPSA/pypsa-eur/pull/107/files/1966cccec6824c86ca46095f230b32094ea1542d#diff-720aa7a680646ab82b6f89980ddb841aL76

yes, it might be due to different pyplot version but for me axes[1] targets (in this case) at the first row and second column as axes is a one-dimensional list

coroa: Ah okay, earlier there was a squeeze argument that defaulted to False maybe it's true now. I think its not cumulative in the plots.

coroa: Did you add https://github.com/PyPSA/pypsa-eur/pull/107#discussion_r355507291 v?

coroa:
This is good to go after renaming the `technology` wildcard in `plot_p_nom_max` to `techs`!

Great work!

FabianHofmann:
thanks! I'll merge this as soon as the tests have past

coroa:
With all the there and back commits. Do you mind to squash it into one humonguous squashed commit?

FabianHofmann:
> With all the there and back commits. Do you mind to squash it into one humonguous squashed commit?

no, squashing is totally fine

PR #110, opened 2019-12-02
import and bug fix for custom ppl

FabianHofmann:
As @coroa figured out, `add_custom_powerplants` needs a type stable import for the `bus` column (and a second bug was detected)


FabianHofmann:
sorry let me do this again.. there are old commit which shouldn't be here

PR #127, opened 2020-02-13
environment: add gurobipy

fneum:


fneum:
failing test was due to connection timeout in the windows test.

PR #128, opened 2020-02-13
helpers: importing networks with override_components

### LINKED ISSUE #122: override_components in config.yaml

Following discussions/questions on the mailing list `Components disappear in snakefile execution`:

On keeping `override_components` throughout the workflow:
> Unfortunately not, however we are planning to make the imports more flexible in future versions. Perhaps we could even enable to add components globally through the snakefile... I think about it. 
For now I only see the two options: (1.)  adjusting every import in the scripts in pypsa-eur or (2.) changing your components.csv in your pypsa/pypsa folder. I would say the first option is the 'saver' one as you only modify one package. @FabianHofmann 

I thought of a way to maintain overwritten components through the `config.yaml`. This is just a sketch:

1. Add the definition of additional components to `config.yaml`, e.g.:

```yaml
override_components:
  ShadowPrice:
    component: ["shadow_prices","Shadow price for a global constraint.",np.nan]
    attributes:
      name: ["string","n/a","n/a","Unique name","Input (required)"]
      value: ["float","n/a",0.,"shadow value","Output"]
```

2. Write a function in `scripts/_helpers.py` to load `override_components` and `override_component_attrs` in the correct format:

```python
import pypsa
import pandas as pd
from pypsa.descriptors import Dict

def read_override_components(components):
    override_components = pypsa.components.components.copy()
    override_component_attrs = Dict({k : v.copy() for k,v in pypsa.components.component_attrs.items()})
    for k, v in components.items():
        override_components.loc[k] = v['component']
        override_component_attrs[k] = pd.DataFrame(columns = ["type","unit","default","description","status"])
        for attr, val in v['attributes'].items():
            override_component_attrs[k].loc[attr] = val
        
    return override_components, override_component_attrs
```

3. In all `scripts/*.py` where a network is imported, add

```python
x, y = read_override_components(snakemake.config['override_components'])
network = pypsa.Network(snakemake.input[0], override_components=x, override_component_attrs=y)
```

fneum:
Closes #122.

See discussion in #122.

Adds a new function `_helpers.load_network()` that loads pypsa.Network with overridden components as specified in `snakemake.config['override_components']`. Docstrings provided.
This is more a precursor for future merges with PyPSA-Eur-Sec, which I added already now due to a recent specific request on the PyPSA mailing list. It's not used in the existing rules, but contributes to extensibility of PyPSA-Eur downstream of the workflow. I think this is fair to add it but not (yet) use it, as I currently do not see a use case where you would want to add extra components in the workflow earlier than maybe `add_extra_components`. This is also why I did not add it in the documentation of `config.yaml`. Convince me if you think otherwise.

## Checklist

### DIFF
--- a/scripts/plot_network.py
+++ b/scripts/plot_network.py
@@ -172,7 +172,7 @@ def plot_map(n, ax=None, attribute='p_nom', opts={}):
 
     return fig
 
-#n = load_network(snakemake.input.network, opts, combine_hydro_ps=False)
+#n = load_network_for_plots(snakemake.input.network, opts, combine_hydro_ps=False)
 
 
 def plot_total_energy_pie(n, ax=None):
@@ -261,7 +261,7 @@ if __name__ == "__main__":
     map_figsize = opts['map']['figsize']
     map_boundaries = opts['map']['boundaries']
 
-    n = load_network(snakemake.input.network, snakemake.input.tech_costs, snakemake.config)
+    n = load_network_for_plot(snakemake.input.network, snakemake.input.tech_costs, snakemake.config)
 
     scenario_opts = snakemake.wildcards.opts.split('-')
 
--- a/scripts/plot_network.py
+++ b/scripts/plot_network.py
@@ -261,7 +261,7 @@ if __name__ == "__main__":
     map_figsize = opts['map']['figsize']
     map_boundaries = opts['map']['boundaries']
 
-    n = load_network_for_plot(snakemake.input.network, snakemake.input.tech_costs, snakemake.config)
+    n = load_network_for_plots(snakemake.input.network, snakemake.input.tech_costs, snakemake.config)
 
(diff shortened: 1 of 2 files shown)


scripts/plot_network.py
```diff
@@ -261,7 +261,7 @@ def split_costs(n):
     map_figsize = opts['map']['figsize']
     map_boundaries = opts['map']['boundaries']
 
-    n = load_network(snakemake.input.network, snakemake.input.tech_costs, snakemake.config)
+    n = load_network_for_plot(snakemake.input.network, snakemake.input.tech_costs, snakemake.config)
```
FabianHofmann: This should probably be ...plots?
```suggestion
    n = load_network_for_plots(snakemake.input.network, snakemake.input.tech_costs, snakemake.config)
```

PR #138, opened 2020-03-13
New components

ulicious:
The adjustment will import carriers and components, which have no impact on the electricity network itself and, therefore, can be added in the very end of the snakemake process: add_extra_components

**Important:**
1.) Components and sub-types of this component have to be declared in the config file (see code bellow)

2.) A new folder (pypsa-eur/data/new_components/) was added. This folder contains:
- new_components_list.csv file: list of all components, which should be added. The components and sub-types, mentioned in the config file, will be added.
- COMPONENT.csv: Lists all attributes of the component (similar to e.g. generators.csv)
- COMPONENT_facilities.csv: Lists all the existing facilities of the component type
- COMPONENT_profiles (optional): Lists the underlying time series of the output of the component

3.) Such components are not added permanently to the network. They need to be imported in every snakemake rule, if adjustments like aggregation or solving should be applied on these new components. This is currently solved with the load_network method in the __helper__ script, but could also be done by importing the add_new_components method from the add_extra_components rule and use the same code (line 340 to 358) of this rule in other rules -> Maybe adding this method to the __helpers__ script?

4.) Furthermore, the snakefile needs to adjusted to import the underlying geojson file (see code bellow)

- Adding the add_carrier, add_new_components and attach_facilities methods to import components, which have no impact on electricity network
- Adjusting the import of the network to consider new components

## Checklist

- [ ] Newly introduced dependencies are added to `environment.yaml` and `environment.docs.yaml`.


ulicious:
One mistake in the description:

"This is currently solved with the load_network method in the helper script, but could also be done by importing the add_new_components method from the add_extra_components rule and use the same code (line 340 to 358) of this rule in other rules"

Due to changes in the code, it's not line 340 to 358, but line 373 to 388.

FabianHofmann:
hey @ulicious, we just had a small discussion about this feature and came to the conclusion that use case of it is not general enough, also it requires additional files that users don't have. Hope you don't mind that I close it

PR #149, opened 2020-03-25
cluster: skip recalculation of link parameters if empty

fneum:


fneum:
Yields KeyError for `underwater_fraction` because it is not added to the columns (no default column)

FabianHofmann:
> Yields KeyError for `underwater_fraction` because it is not added to the columns (no default column)

Alright, as to expect

PR #150, opened 2020-04-03
Renewable profiles option using all available areas

JanFrederickUnnewehr:
The user can use this functionality to decide whether he wants to use all available area for the potential calculation or not.
As a result, two options are now available to reduce the available potential for areas with very low potential. This can be used for example to generated CF for renewable generation units that are already existing (a system that represents today's capacity's) as reported  #103 
This can results in the problem, that renewable generators are connected to buses that do not have a CF time series, because no potential was determined at the bus. 

Option 1 (line 292)-> new implementation
```
if config.get('keep_all_available_areas', False) != True: # check config file whether user wants to use all availbe weather cells for renewable profile and potential generation
        potmatrix.data[potmatrix.data < 1.] = 0 # ignore weather cells where only less than 1 MW can be installed
```
This option works directly on the resulting potential areas and not on the assigned pypsa grid potentials.

Option 2 (line 359)-> already existing option
 ```
# select only buses with some capacity and minimal capacity factor
    ds = ds.sel(bus=((ds['profile'].mean('time') > config.get('min_p_max_pu', 0.)) &
                     (ds['p_nom_max'] > config.get('min_p_nom_max', 0.))))
```
This option works on the resulting / assigned pypsa grid potentials and CF

Should we add both options to the config and documentation files?

## Checklist

~~- [ ] Newly introduced dependencies are added to `environment.yaml` and `environment.docs.yaml`.~~

### DIFF
--- a/scripts/build_renewable_profiles.py
+++ b/scripts/build_renewable_profiles.py
@@ -289,7 +289,8 @@ if __name__ == '__main__':
 
     potentials = config['capacity_per_sqkm'] * vlanduse._cutout_cell_areas(cutout)
     potmatrix = matrix * spdiag(potentials.ravel())
-    potmatrix.data[potmatrix.data < 1.] = 0 # ignore weather cells where only less than 1 MW can be installed
+    if config.get('keep_all_available_areas', False) != True:
+        potmatrix.data[potmatrix.data < 1.] = 0 # ignore weather cells where only less than 1 MW can be installed
     potmatrix.eliminate_zeros()
 
     resource = config['resource']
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: scripts/build_renewable_profiles.py [hunk 1]
```diff
@@ -289,7 +289,8 @@ def calculate_potential(gid, save_map=None):
 
     potentials = config['capacity_per_sqkm'] * vlanduse._cutout_cell_areas(cutout)
     potmatrix = matrix * spdiag(potentials.ravel())
-    potmatrix.data[potmatrix.data < 1.] = 0 # ignore weather cells where only less than 1 MW can be installed
+    if config.get('keep_all_available_areas', False) != True: # check config file whether user wants to use all availbe weather cells for renewable profile and potential generation
```
coroa, line 292:
```suggestion
    if not config.get('keep_all_available_areas', False):
```

JanFrederickUnnewehr:
@coroa, cleaned code for a better readability
```
 if not config.get('keep_all_available_areas', False): # check config file whether user wants to use all availbe weather cells for renewable profile and potential generation
        potmatrix.data[potmatrix.data < 1.] = 0 # ignore weather cells where only less than 1 MW can be installed
```

PR #166, opened 2020-06-27
Options for regional equity and autarky

fneum:

Add "distributional production equity" and "autarky" options to `{opts}` wildcard:

Require each country or node to on average produce a minimal share of its total consumption itself. Example: `EQ0.5c` demands each country to produce on average at least 50% of its consumption; `EQ0.5` demands each node to produce on average at least 50% of its consumption.

Moreover, the option `ATK` removes all lines and links requiring each node to be autarkic.
`ATKc` removes all cross-border transmission.

Also can now set a global capacity limit for individual lines/links in `config["links"]["p_nom_max"]` and `config["lines"]["s_nom_max"]` (this still needs to be added to docs).

## Checklist


scripts/solve_network.py
```diff
@@ -167,6 +168,28 @@ def add_CCL_constraints(n, config):
...
+    load = n.snapshot_weightings @ \
+           n.loads_t.p_set.groupby(lgrouper, axis=1).sum()
+    inflow = n.storage_units_t.inflow.groupby(sgrouper, axis=1).sum().sum()
+    inflow = inflow.reindex(load.index).fillna(0.)
+    rhs = scaling * level * ( load - inflow )
+    lhs = linexpr((n.snapshot_weightings * scaling,
+                   get_var(n, "Generator", "p").T)
+                 ).T.groupby(ggrouper, axis=1).apply(join_exprs)
+    define_constraints(n, lhs, ">=", rhs, "equity", "min")
+
```
FabianHofmann, line 191: Nice syntax here! 

However here is a problem with the (load - inflow) part. Note you still have the spillage variable, i.e. not all of the inflow is converted into power production. I see two options, either add the negative spillage variable to the lhs (then the inflow should be outside of the parantheses), or add the positive power production of storage units to the lhs. 

Here the equations

$p_{generators}^{c} + p_{storage}^d = \chi d^c$

where c is the group, $\chi$ the level and $p^c$ the total power production per group. With 

$p_{storage}^c = p_{inflow}^c - p_{spillage}^c$

you see what I mean.

scripts/prepare_network.py
```diff
@@ -224,4 +224,9 @@ def average_every_nhours(n, offset):
     elif ll_type == 'c':
         set_line_cost_limit(n, factor, Nyears)
 
+    ln_config = snakemake.config["lines"]
+    lk_config = snakemake.config["links"]
+    n.lines.s_nom_max = ln_config.get("s_nom_max,", np.inf)
+    n.links.loc[n.links.carrier=='DC', "p_nom_max"] = lk_config.get("p_nom_max", np.inf)
+
```
FabianHofmann, line 231: Don't you think it makes more sense to **clip** the maximum capacities to the upper limit. Here the limits which might be lower than `ln_config['s_nom_max']` for example are overwritten.

scripts/solve_network.py
FabianHofmann: Sorry, it does not render the latex equations...

fneum: > Note you still have the spillage variable

Ah, yes! I'll see how to implement it.

> Sorry, it does not render the latex equations...

I wonder why Github does not manage to add this feature...

scripts/prepare_network.py
fneum: True, did not think about it, because we never had limits on those.

scripts/solve_network.py
fneum: Two options then:
1) weight * (generation - spill) + inflow >= level * weight * load
2) generation + storage discharge - storage charge >= level * load
I have a slight preference for 1) because fewer variables. @FabianHofmann what do you think?

scripts/solve_network.py
```diff
@@ -167,6 +168,33 @@ def add_CCL_constraints(n, config):
...
+        sgrouper = n.storage_units.bus.map(n.buses.country)
+    else:
+        ggrouper = n.generators.bus
+        lgrouper = n.loads.bus
+        sgrouper = n.storage_units.bus
+    load = n.snapshot_weightings @ \
+           n.loads_t.p_set.groupby(lgrouper, axis=1).sum()
+    inflow = n.storage_units_t.inflow.groupby(sgrouper, axis=1).sum().sum()
+    inflow = inflow.reindex(load.index).fillna(0.)
+    rhs = scaling * ( level * load - inflow )
```
FabianHofmann, line 186: Last thing here: I think the inflow should also be weighted with snapshot weightings or?

fneum: True, somehow I thought it was MWh already.

Issue #67, opened 2019-08-12
Update data in the data bundle to most up-to-date versions


fneum:
For some of the data in the data bundle newer versions are available ; e.g. for

- `data/bundle/EIA_hydro_generation_2000_2014.csv` up to 2016
- `data/bundle/NUTS_2013_60M_SH/data/NUTS_RG_60M_2013.shp` for 2016
- `data/bundle/nama_10r_3popgdp.tsv.gz` up to 2017
- `data/bundle/je-e-21.03.02.xls` up to 2017
- `data/bundle/corine/g250_clc06_V18_5.tif` with CLC 2018 https://land.copernicus.eu/pan-european/corine-land-cover/clc2018
- `data/bundle/GEBCO_2014_2D.nc` for 2019
- `data/bundle/time_series_60min_singleindex_filtered.csv` possibly
- entsoegridkit #57 #55 
- powerplantmatching #61 
-

coroa:
`pietzker2014.xlsx` should pbb be name `pietzcker2014.xlsx` and is Supplementary data 2 from https://www.sciencedirect.com/science/article/pii/S0306261914008149

coroa:
Updating `data/bundle/time_series_60min_singleindex_filtered.csv` will probably make some work, since the load data format changed somewhen in 2016/17 with overlapping data, which is kept in a separate column. So one will have to come up with a rule to combine the two columns to one consistent load time series.

coroa:
For updating the GIS files, `corine` and `GEBCO`, it must be verified explicitly that they still agree with specified coordinate reference systems in https://github.com/PyPSA/pypsa-eur/blob/7a18740cfb565b5d83d04387f3a7b35c28fa56cc/scripts/build_renewable_profiles.py#L36-L40 .

coroa:
New versions of the data bundle should be put as a separate zenodo file (see also #60)

fneum:
to be combined with xref #342

fneum:
mostly done over time, e.g. #1027 #936 #924

PR #170, opened 2020-07-10
Update load processing; part of #67

JanFrederickUnnewehr:
Is part of #67 

The following update concerns the processing of load data in pypsa-eur.
Added a rule (`build_load_data`) that downloads the latest load data from the OPSD website.
The resulting data are cleaned and gaps filled based on manual filling methods. Before and after filling gaps the rule proved information about the gaps (length, frequency) by the function `nan_statistics(df)`.

## Checklist

- [X] I tested my contribution locally and it seems to work fine.
- [X] Newly introduced dependencies are added to `environment.yaml` and `environment.docs.yaml`.
- [X] Changes in configuration options are added in all of `config.default.yaml`, `config.tutorial.yaml`, and `test/config.test1.yaml`.

Is there a way to automate the generation of the file `doc/configuration.rst`.
Lets discuss first the proposed changes before I will add it to the `doc/release_notes.rst`.

### DIFF
--- a/scripts/add_electricity.py
+++ b/scripts/add_electricity.py
@@ -211,7 +211,7 @@ def attach_load(n):
                 .filter(items=snakemake.config['countries']))
     
     # Scalling data according to scalling factor in config.yaml
-    logger.info(f"Load data scalled with scalling factior {snakemake.config['load']['scaling_factor']}.")
+    logger.info(f"Load data scalled with scalling factor {snakemake.config['load']['scaling_factor']}.")
     opsd_load = opsd_load * snakemake.config.get('load', {}).get('scaling_factor', 1.0)
 
     # Convert to naive UTC (has to be explicit since pandas 0.24)
--- a/scripts/add_electricity.py
+++ b/scripts/add_electricity.py
@@ -26,9 +26,9 @@ Relevant Settings
         extendable_carriers:
             Generator:
         estimate_renewable_capacities_from_capacity_stats:
-
+            
     load:
-        scaling_factor:
+        scaling_factor:           
 
     renewable: (keys)
         hydro:
@@ -53,14 +53,9 @@ Inputs
         :scale: 34 %
 
 - ``data/geth2015_hydro_capacities.csv``: alternative to capacities above; NOT CURRENTLY USED!
-- ``data/bundle/time_series_60min_singleindex_filtered.csv``: Hourly per-country load profiles since 2010 from the `ENTSO-E statistical database <https://www.entsoe.eu/data/power-stats/hourly_load/>`_
 
-    .. image:: ../img/load-box.png
-        :scale: 33 %
 
-    .. image:: ../img/load-ts.png
-        :scale: 33 %
-
+- ``resources/load.csv`` Hourly per-country load profiles.
 - ``resources/regions_onshore.geojson``: confer :ref:`busregions`
 - ``resources/nuts3_shapes.geojson``: confer :ref:`shapes`
 - ``resources/powerplants.csv``: confer :ref:`powerplants`
@@ -91,7 +86,6 @@ It further adds extendable ``generators`` with **zero** capacity for
 """
 
 from vresutils.costdata import annuity
-from vresutils.load import timeseries_opsd
 from vresutils import transfer as vtransfer
 
 import logging
@@ -203,6 +197,17 @@ def load_powerplants(ppl_fn=None):
             .rename(columns=str.lower).drop(columns=['efficiency'])
             .replace({'carrier': carrier_dict}))
 
+def load_opsd_loaddata(load_fn=None, countries=None):
+    if load_fn is None:
+        load_fn = snakemake.input.load
+    
+    if countries is None:
+        countries = snakemake.config['countries']
+    
+    load = pd.read_csv(load_fn, index_col=0, parse_dates=True)
+    load = load.filter(items=countries)
+    
+    return (load)
 
 # =============================================================================
 # Attach components
@@ -214,9 +219,11 @@ def attach_load(n):
     substation_lv_i = n.buses.index[n.buses['substation_lv']]
     regions = (gpd.read_file(snakemake.input.regions).set_index('name')
                .reindex(substation_lv_i))
-    opsd_load = (timeseries_opsd(slice(*n.snapshots[[0,-1]].year.astype(str)),
-                                 snakemake.input.opsd_load) *
-                 snakemake.config.get('load', {}).get('scaling_factor', 1.0))
+    opsd_load = load_opsd_loaddata(load_fn=snakemake.input.load, countries=snakemake.config['countries'])
+    
+    # Scalling data according to scalling factor in config.yaml
+    logger.info(f"Load data scalled with scalling factior {snakemake.config['load']['scaling_factor']}.")
+    opsd_load = opsd_load * snakemake.config.get('load', {}).get('scaling_factor', 1.0)
 
     # Convert to naive UTC (has to be explicit since pandas 0.24)
     opsd_load.index = opsd_load.index.tz_localize(None)
--- a/scripts/add_electricity.py
+++ b/scripts/add_electricity.py
@@ -197,18 +197,6 @@ def load_powerplants(ppl_fn=None):
             .rename(columns=str.lower).drop(columns=['efficiency'])
             .replace({'carrier': carrier_dict}))
 
-def load_opsd_loaddata(load_fn=None, countries=None):
-    if load_fn is None:
-        load_fn = snakemake.input.load
-    
-    if countries is None:
-        countries = snakemake.config['countries']
-    
-    load = pd.read_csv(load_fn, index_col=0, parse_dates=True)
-    load = load.filter(items=countries)
-    
-    return (load)
-
 # =============================================================================
 # Attach components
 # =============================================================================
--- a/scripts/add_electricity.py
+++ b/scripts/add_electricity.py
@@ -55,7 +55,7 @@ Inputs
 - ``data/geth2015_hydro_capacities.csv``: alternative to capacities above; NOT CURRENTLY USED!
 
 
-- ``resources/load.csv`` Hourly per-country load profiles.
+- ``resources/opsd_load.csv`` Hourly per-country load profiles.
 - ``resources/regions_onshore.geojson``: confer :ref:`busregions`
 - ``resources/nuts3_shapes.geojson``: confer :ref:`shapes`
 - ``resources/powerplants.csv``: confer :ref:`powerplants`
--- a/scripts/build_load_data.py
+++ b/scripts/build_load_data.py
@@ -183,10 +183,7 @@ def manual_adjustment(load, source="ENTSOE_power_statistics"):
     
         load.loc[start:stop, 'EE'] = load.loc[start-w:stop-w, 'EE'].values
             
-        # to fill a 3 hour gap in the night
-        load['EE'] = load['EE'].interpolate()
-        
-        
+    
         # To fill the gaps in FR from start to stop,
         # we copy the same period from one week before into it
         start = pd.Timestamp('2018-08-12 07:00')
@@ -194,9 +191,7 @@ def manual_adjustment(load, source="ENTSOE_power_statistics"):
         w = pd.Timedelta(weeks=1)
     
         load.loc[start:stop, 'FR'] = load.loc[start-w:stop-w, 'FR'].values
-            
-        # to fill a two 3 hour gaps in the night
-        load['FR'] = load['FR'].interpolate()
+
         
         # To fill the first gaps in LT from start to stop,
         # we copy the same period from the next sunnday into it
@@ -258,7 +253,6 @@ def manual_adjustment(load, source="ENTSOE_power_statistics"):
         
         load.loc[start:stop, 'LU'] = load.loc[start+w:stop+w, 'LU'].values
         
-        load['LU'] = load['LU'].interpolate()
     
         # To fill the gaps in MK from start to stop,
         # we copy the same period from one week before into it
@@ -368,7 +362,6 @@ def manual_adjustment(load, source="ENTSOE_power_statistics"):
         load.loc[date_1, 'MK'] = load.loc[date_1+w, 'MK']
         load.loc[date_2, 'MK'] = load.loc[date_2+w, 'MK']
         
-        load['MK'] = load['MK'].interpolate()
       
         # Kosovo (KV) and Albania (AL) do not exist in the data set
         # Kosovo (KV) gets the same load curve as Serbia (RS)
--- a/scripts/build_load_data.py
+++ b/scripts/build_load_data.py
@@ -13,7 +13,6 @@ Relevant Settings
     load:
         url:
         source:
-        scaling_factor:
 
 .. seealso::
     Documentation of the configuration file ``config.yaml`` at
@@ -34,7 +33,6 @@ Description
 
 The configuration options ``load: source:`` can be used to control which load data should be used in the model "ENTSOE_transparency" or "ENTSOE_power_statistics".
 
-The configuration options ``load: scaling_factor:`` can be used to scal the completly load data by a specify factor.
 
 """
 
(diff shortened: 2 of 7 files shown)


scripts/add_electricity.py
```diff
@@ -214,9 +219,11 @@ def attach_load(n):
     substation_lv_i = n.buses.index[n.buses['substation_lv']]
     regions = (gpd.read_file(snakemake.input.regions).set_index('name')
                .reindex(substation_lv_i))
-    opsd_load = (timeseries_opsd(slice(*n.snapshots[[0,-1]].year.astype(str)),
-                                 snakemake.input.opsd_load) *
-                 snakemake.config.get('load', {}).get('scaling_factor', 1.0))
+    opsd_load = load_opsd_loaddata(load_fn=snakemake.input.load, countries=snakemake.config['countries'])
```
FabianHofmann, line 222:
```suggestion
    opsd_load = (pd.read_csv(snakemake.input.load, index_col=0, parse_dates=True)
                .filter(items=snakemake.config['countries']))
```
I think it's fine to take a shortcut here without defining the function above.

scripts/build_load_data.py
```diff
@@ -0,0 +1,442 @@
...
+    load : pd.DataFrame
+        Load time-series with UTC timestamps x ISO-2 countries
+    """
+        
+    if countries is None:
+        countries = snakemake.config['countries']
+     
+    if source == 'ENTSOE_transparency':
+        load = (pd.read_csv(fn, index_col=0, parse_dates=True)
+                .loc[:, lambda df: df.columns.to_series().str.endswith('_load_actual_entsoe_transparency')]
```
FabianHofmann, line 75:
```suggestion
                .filter(like='_load_actual_entsoe_transparency')
```
or with regex ($ for end of line)
`.filter(regex='_load_actual_entsoe_transparency$')`

scripts/build_load_data.py
```diff
@@ -0,0 +1,442 @@
...
+                .rename(columns=lambda s: s[:-len('_load_actual_entsoe_transparency')])
+                .dropna(how="all", axis=0))
+        
+    elif source == 'ENTSOE_power_statistics':
+        load = (pd.read_csv(fn, index_col=0, parse_dates=True)
+            .loc[:, lambda df: df.columns.to_series().str.endswith('_load_actual_entsoe_power_statistics')]
+            .rename(columns=lambda s: s[:-len('_load_actual_entsoe_power_statistics')])
+            .dropna(how="all", axis=0))
+    else:
+        logger.warning("Please proviede correct source name for load data")
```
FabianHofmann, line 85:
```suggestion
        raise NotImplementedError(f"Data for source `{source}` not available.")
```

scripts/build_load_data.py
```diff
@@ -0,0 +1,442 @@
...
+    
+    
+    if 'GB_UKM' in load.columns:
+        load.rename(columns={'GB_UKM' : 'GB'}, inplace=True)
+    
+    load = load.filter(items=countries)
```
FabianHofmann, line 91:
```suggestion
   load = load.rename(columns={'GB_UKM' : 'GB'}).filter(items=countries)
```

scripts/build_load_data.py
```diff
@@ -0,0 +1,442 @@
...
+
+    if years is not None:
+        load = load.loc[years]
+        
+    
+    return load
+
```
FabianHofmann, line 98:
```suggestion
def consecutive_nans(ds):
    return (ds.isnull().astype(int)
            .groupby(ds.notnull().astype(int).cumsum())
            .transform('sum'))

def fill_large_gaps(ds, gapsize=3):
    """Fill up large gaps with load data from the previous week."""
    week_shift = pd.Series(ds.values, ds.index + pd.Timedelta('1w'))
    return ds.where(consecutive_nans(ds) < gapsize, week_shift.reindex_like(ds))
```

Instead of all the manual corrections below, it's easier to rely on some heuristics. With those two functions one can simply call 
```
df = df.apply(fill_large_gaps).interpolate()
```
for the interpolation. The copying of load data for KV and AL has still to be done. But for the latter an `if` statement would be helpful which checks whether they do not exist (who knows, next versions of opsd might introduce them)

scripts/build_load_data.py
```diff
@@ -0,0 +1,442 @@
...
+    logger.info(f"Downloading load data from '{url}'.")
+    
+    progress_retrieve(url, to_fn)
+
+    logger.info(f"Raw load data available at '{to_fn}'.")
+    
+    opsd_load = (load_timeseries_opsd(years = slice(*pd.date_range(freq='y', **snakemake.config['snapshots'])[[0,-1]].year.astype(str)),
+                                 fn=to_fn,
+                                 countries = snakemake.config['countries'],
+                                 source = snakemake.config['load']['source']))
```
FabianHofmann, line 421:
```suggestion
    opsd_load = (load_timeseries_opsd(years = slice(*pd.date_range(freq='y', **snakemake.config['snapshots'])[[0,-1]].year.astype(str)),
                                 fn=url,
                                 countries = snakemake.config['countries'],
                                 source = snakemake.config['load']['source']))
```
Directly reading from url is fine, as we do not really need the raw data

scripts/add_electricity.py
JanFrederickUnnewehr: That's a good idea. So, we can save a few lines of code.

scripts/build_load_data.py
JanFrederickUnnewehr: OK, I have also thought that it might be useful to separate the loading and processing of load data. Then we could control the automatic loading of the data like in the data bundle. Maybe this is a bit too much work just for the load data. Maybe it makes sense to integrate everything into the load data bundle function.

FabianHofmann: I think it's fine to do it here. Its good to have as many things as possible outside of the databundle. It causes extra work to maintain it properly. The more is retrieved automatically, the better.

eb5194:
maybe we could add a switch if we want to manually fill the gaps as in the old timeseries_opsd? See PR in FRESNA https://github.com/FRESNA/vresutils/pull/14/files ("manual_alterations"), as there's many gaps lasting for a couple of hours only, but others as well for many months. Maybe we could add a switch in the config manual_load_alterations (or similar): True/False? Then it's up to the user

JanFrederickUnnewehr:
Moin, check out the new version of `build_load_data `. I have tried to include all suggestions in the code. Did I forget anything? We could leave the big manual adjustment part as an example but also shorten it, as it is not complete. See 'GB'

eb5194:
you copy often from start to stop, maybe for that you can use my function as in vresutils:

`def copy_timeslice(load, cntry, start, stop, delta):
    start = pd.Timestamp(start)
    stop = pd.Timestamp(stop)
    if start in load.index and stop in load.index:
        load.loc[start:stop, cntry] = load.loc[start-delta:stop-delta, cntry].values
    return load`

and then use it as, for example, `copy_timeslice(load, 'GR', '2015-08-11 21:00', '2015-08-15 20:00', pd.Timedelta(weeks=1))`. Saves many many lines of code and makes it more readable?

also, if you interpolate, you can add a limit of how many hours should be interpolated at max. Before, we used 4 hours.
load[interpolate_countries] = load[interpolate_countries].interpolate(limit=4).

You update Kosovo and Albania twice, line 142+
`load['KV'] = load['RS'] * (4.8 / 27.)
load['AL'] = load['MK'] * (4.1 / 7.4)`
and then in 370+
load['KV'] = load['RS'] * (5. / 33.)
load['AL'] = load['MK'] * (6.0 / 7.0)`
but with different factors. which one is correct?

you can try to use my suggestions from the PR in vresutils I mentioned before, I also discussed them a while ago with @coroa. But your "alterations" seem more complete. Did you take into account holidays, weekends, etc.?

Now we have a code duplicate once in vresutils and here... not sure what to do with it. Is someone else using vresutils? If its just PyPSA-Eur, then we can delete it there and keep only this version otherwise we have two different sources people could use and they are not exactly the same, so produce different results. Super bad for debugging later on

scripts/build_load_data.py
```diff
@@ -0,0 +1,430 @@
...
+    # adjust gaps manuel
+    if snakemake.config['load']['adjust_gaps_manuel']:
+        logger.info(f"Load data are adjusted manual.")
+        opsd_load = manual_adjustment(load=opsd_load, source=snakemake.config['load']['source'])
+    
+    # check the number and lenght of gaps after adjustment and interpolating
+    nan_stats = nan_statistics(opsd_load)
+    
+    if nan_stats.consecutive.max() > gap_filling_threshold:        
+        logger.warning(f'Load data contains gaps after manuel adjustment. Modify manual_adjustment() function!')
```
FabianHofmann, line 427: I think we're almost there. I would restructure this part a bit (sorry for many iterations). 

1. manual corrections if enabled
2. interpolate gaps < gap_size_interpolated (with info) if nan exist
3. fill by weekly shift (with a warning) if nan exist
4. raise a error if nan still exist

Like this we are prioritizing the manual corrections and do the heuristic sanitizing afterwards

FabianHofmann:
> you copy often from start to stop, maybe for that you can use my function as in vresutils:
> 
> ```
> def copy_timeslice(load, cntry, start, stop, delta):
>     start = pd.Timestamp(start)
>     stop = pd.Timestamp(stop)
>     if start in load.index and stop in load.index:
>         load.loc[start:stop, cntry] = load.loc[start-delta:stop-delta, cntry].values
>     return load
> ```
> 
> and then use it as, for example, `copy_timeslice(load, 'GR', '2015-08-11 21:00', '2015-08-15 20:00', pd.Timedelta(weeks=1))`. Saves many many lines of code and makes it more readable?
> 
> also, if you interpolate, you can add a limit of how many hours should be interpolated at max. Before, we used 4 hours.
> `load[interpolate_countries] = load[interpolate_countries].interpolate(limit=4)`.
> 
> You update Kosovo and Albania twice, line 142+
> 
> ```
> load['KV'] = load['RS'] * (4.8 / 27.)
> load['AL'] = load['MK'] * (4.1 / 7.4)
> ```
> 
> and then in 370+
> 
> ```
> load['KV'] = load['RS'] * (5. / 33.)
> load['AL'] = load['MK'] * (6.0 / 7.0)
> ```
> 
> but with different factors. which one is correct?
> 
> you can try to use my suggestions from the PR in vresutils I mentioned before, I also discussed them a while ago with @coroa. But your "alterations" seem more complete. Did you take into account holidays, weekends, etc.?
> 
> Now we have a code duplicate once in vresutils and here... not sure what to do with it. Is someone else using vresutils? If its just PyPSA-Eur, then we can delete it there and keep only this version otherwise we have two different sources people could use and they are not exactly the same, so produce different results. Super bad for debugging later on

scripts/add_electricity.py
```diff
@@ -203,6 +197,17 @@ def load_powerplants(ppl_fn=None):
...
+    if load_fn is None:
+        load_fn = snakemake.input.load
+    
+    if countries is None:
+        countries = snakemake.config['countries']
+    
+    load = pd.read_csv(load_fn, index_col=0, parse_dates=True)
+    load = load.filter(items=countries)
+    
+    return (load)
```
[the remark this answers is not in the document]
JanFrederickUnnewehr: Yes, right!

scripts/build_load_data.py
JanFrederickUnnewehr: Should we leave the warning at the beginning that the data set contains gaps? In line 411 & 412
The new order is a good idea!

JanFrederickUnnewehr:
> 
> 
> you copy often from start to stop, maybe for that you can use my function as in vresutils:
> 
> ```
> def copy_timeslice(load, cntry, start, stop, delta):
>     start = pd.Timestamp(start)
>     stop = pd.Timestamp(stop)
>     if start in load.index and stop in load.index:
>         load.loc[start:stop, cntry] = load.loc[start-delta:stop-delta, cntry].values
>     return load
> ```
> 
> and then use it as, for example, `copy_timeslice(load, 'GR', '2015-08-11 21:00', '2015-08-15 20:00', pd.Timedelta(weeks=1))`. Saves many many lines of code and makes it more readable?

That is a good idea. I will integrate in my code.

> also, if you interpolate, you can add a limit of how many hours should be interpolated at max. Before, we used 4 hours.
> `load[interpolate_countries] = load[interpolate_countries].interpolate(limit=4)`.

Limit is now implemented as `gap_filling_threshold`

> You update Kosovo and Albania twice, line 142+
> 
> ```
> load['KV'] = load['RS'] * (4.8 / 27.)
> load['AL'] = load['MK'] * (4.1 / 7.4)
> ```
> 
> and then in 370+
> 
> ```
> load['KV'] = load['RS'] * (5. / 33.)
> load['AL'] = load['MK'] * (6.0 / 7.0)
> ```
> 
> but with different factors. which one is correct?

The correction are for two different years. The first one only for years before 2016 the secound one for year 2018.
"scale parameter selected by energy consumption ratio from IEA Data browser for the year 2017"

> you can try to use my suggestions from the PR in vresutils I mentioned before, I also discussed them a while ago with @coroa. But your "alterations" seem more complete. Did you take into account holidays, weekends, etc.?

Holidays are not included. Any idea, how to integrate it in an automated way?

 
> Now we have a code duplicate once in vresutils and here... not sure what to do with it. Is someone else using vresutils? If its just PyPSA-Eur, then we can delete it there and keep only this version otherwise we have two different sources people could use and they are not exactly the same, so produce different results. Super bad for debugging later on

When the changes are transferred to the master, the previous used functions "vresutils.load" are simply no longer imported. Is that what you mean?

eb5194:
> The correction are for two different years. The first one only for years before 2016 the secound one for year 2018.
> "scale parameter selected by energy consumption ratio from IEA Data browser for the year 2017"

Will that work, if you run i.e. 2013-2018? Wouldn't then years 2013-2018 be adapted with the 2018 value?

> Holidays are not included. Any idea, how to integrate it in an automated way?

No, sorry. Unfortunately I did it brute force. But maybe, if you add an "holidays" array, and then check if the gap you want to fill is in range of this array, the timedelta increases by one. Also, a weekend days should not be filled with a working day... would that work?
 
> When the changes are transferred to the master, the previous used functions "vresutils.load" are simply no longer imported. Is that what you mean?

I think we should remove one of them, otherwise it's just more code to be maintained, which basically should be 1:1 the same and that's just confusing

JanFrederickUnnewehr:
> Will that work, if you run i.e. 2013-2018? Wouldn't then years 2013-2017 be adapted with the 2018 value?

Is it intended that the simulation period of one calculation (one model run) is longer than one year? If so, the function `manual_adjustment()` must be modified. The old code allows more years but only one data source ("ENTSOE_power_statistics") and the manual adjustment is not complete for all countries and years. This code currently only allows one year at a time but both data sources ("ENTSOE_transparency" or "ENTSOE_power_statistics").
"ENTSOE_power_statistics" is only available until mid of 2019, as far as I know  

> > Holidays are not included. Any idea, how to integrate it in an automated way?
I would ignore that for the moment.  

 "vresutils.load"
Surely we should have only one version of load processing at the end. Let's see what @coroa has to say about this.

JanFrederickUnnewehr:
After a fruitful discussion with @FabianHofmann we agreed on the following procedure:
-When downloading the data, they are automatically filtered for the simulation period (snapshots from config)
-missing countries are "added" in the manual adjustment function
-the scaling parameters are not automatically adjusted by the selected simulation year
-the scaling factor for the manual "adding" of countries must be changed manually by the user in the code
-in the manual adjustment function, a distinction is still made between the two sources in order to be able to include "ENTSOE_transparency" data for future load time series

Code structure:
1) manual adjustment of the data (if the user sets adjust_gaps_manual: true)
2) interpolate the data (considering the gap_filling_threshold: 3 #hours)
3) for larger gaps (max = 7 days) copy the previous week
4) test if there are still gaps, if yes stop with warning and point to manual adjustment function

FabianHofmann:
Hey @JanFrederickUnnewehr,  I'm closing this in the favour of #211 . If you still want to integrate the power statistics time series there should be a good way. But we can discuss it in the other PR or via mail/phone.

PR #171, opened 2020-07-10
fix emission prices

FabianHofmann:

I'm not sure if the previous setup was intentional, but regarding that different generators might have different efficiencies and the emissions are carrier-specific only, it makes more sense assume that the price in the config sets net emission price.

## Checklist

- [x] ~Code and workflow changes are sufficiently documented.~
- [x] ~Newly introduced dependencies are added to `environment.yaml` and `environment.docs.yaml`.~
- [x] ~Changes in configuration options are added in all of `config.default.yaml`, `config.tutorial.yaml`, and `test/config.test1.yaml`.~
- [x] ~Changes in configuration options are also documented in `doc/configtables/*.csv` and line references are adjusted in `doc/configuration.rst` and `doc/tutorial.rst`.~

### DIFF
--- a/config.default.yaml
+++ b/config.default.yaml
@@ -179,7 +179,7 @@ costs:
     hydro: 0.
     H2: 0.
     battery: 0.
-  emission_prices: # only used with the option Ep
+  emission_prices: # in currency per tonne emission, only used with the option Ep
     co2: 0.
 
 solving:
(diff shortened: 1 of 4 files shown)


fneum (commented):
Yes! Let's change that. But needs to be documented in the configuration options and the opts wildcard, as well as a brief release note before merging.

PR #172, opened 2020-07-20
add compatibility with pyomo 5.7.0

fneum:


fneum:
grazie @FabianHofmann

PR #181, opened 2020-08-23
Fix two issues with new 2020 ENTSO-E grid extract

fneum:

There were two issues with the 2020 ENTSO-E grid extract from https://www.entsoe.eu/data/map/ preventing clustering down to 37 nodes (39 was the minimum because additional subnetworks not connected to AC network or cross-country HVDC links were formed).

**Skagerrak HVDC links between DK and NO** were partially allocated to the wrong terminal bus in NO and included a "self-link".

For **HVDC connections mainland IT, Sardinia and Corsica** a node was wrongly created at the intersection between two actually separate connections.

Note 22 Aug 2020

## Checklist

~- [ ] Code and workflow changes are sufficiently documented.~
~- [ ] Changes in configuration options are added in all of `config.default.yaml`, `config.tutorial.yaml`, and `test/config.test1.yaml`.~
~- [ ] Changes in configuration options are also documented in `doc/configtables/*.csv` and line references are adjusted in `doc/configuration.rst` and `doc/tutorial.rst`.~

### DIFF
--- a/scripts/base_network.py
+++ b/scripts/base_network.py
@@ -275,6 +275,13 @@ def _apply_parameter_corrections(n):
                 inds = r.index.intersection(df.index)
                 df.loc[inds, attr] = r[inds].astype(df[attr].dtype)
 
+def _apply_sardinia_corsica_link_corrections(n):
+    for lk_keep, lk_remove in [("12998", "12997"),("5627", "5628")]:
+        n.links.at[lk_keep, "bus1"] = n.links.at[lk_remove, "bus0"]
+        n.links.at[lk_keep, "length"] += n.links.at[lk_remove, "length"]
+        n.remove("Link", lk_remove)
+    n.remove("Bus", "7276")
+
 def _set_electrical_parameters_lines(lines):
     v_noms = snakemake.config['electricity']['voltages']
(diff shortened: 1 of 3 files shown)


scripts/base_network.py
```diff
@@ -548,6 +555,8 @@ def base_network():
 
     _apply_parameter_corrections(n)
 
+    _apply_sardinia_corsica_link_corrections(n)
+
```
FabianHofmann, line 559:
```suggestion
```
Now, `n = _remove_unconnected_components(n)` has to be placed below `_apply_parameter_corrections(n)` and bus '7276' is removed

FabianHofmann (commented):
Great that you figured out the problems! I don't love the hard coded bit though. I made a suggestion how we can avoid the hard-coded ` _apply_sardinia_corsica_link_corrections(n)` function.

fneum:
Ah, this trick is very neat! I also thought the hardcoded bit was not the nicest, but couldn't think of this alternative.

FabianHofmann (approved):
Nice! I think as soon you double-checked if the assignment is correct, let's merge it.

fneum:
Behaviour is the same as in the previous hardcoded version.

PR #182, opened 2020-08-25
Take account of varying store/dispatch efficiencies of H2 storage rather than only roundtrip

pz-max:
Closes # (if applicable).

I think the previous code version, in the config.yaml "Storage_Unit" case, was the correct approach for the battery storage but lead to a false round-trip efficiency for H2. For instance, the input efficiency of the fuel cell efficiency will be square rooted (0.58^0.5), as well as the electrolysor (0.8^0.5) resulting in a false roundtrip efficiency of 68% rather than 46%.

## Checklist

Required changes in `add_electricity.py`, `add_extra_components` and `costs.csv`

- [x] Newly introduced dependencies are added to `environment.yaml` and `environment.docs.yaml`.


pz-max:
Unrelated to the above changes but relevant for the storage_unit case, in the add_extra_components.py line 80 and 81 the following is given:
               `efficiency_store=costs.at[carrier, 'efficiency'],`
               `efficiency_dispatch=costs.at[carrier, 'efficiency'],`  
Both efficiency_store and efficiency_dispatch **take the same efficiency** which is defined at add_electricity as roundtrip efficiency. But should I not rather input the real data for store and dispatch efficiency instead of the roundtrip efficiency? (Think the solver takes at the moment just the dispatch RT efficiency as input value)

fneum:
Thanks @pz-max for bringing up this issue. There are definitely aspects to discuss.

As you already noted, the current implementation is correct for battery storage. Your changes just switch from noting the round-trip efficiency of the inverters in `data/costs.csv` to noting the single-trip efficiency. This is a good idea, as it unifies how efficiencies are represented in the `data/costs.csv`.

As for the H2 storage, also the round-trip efficiency seems to be correct already in the current implementation. Correct me if I am wrong!:

https://github.com/PyPSA/pypsa-eur/blob/9331f1ec46ceafd8aeffbbb655e3ec9acee3acdb/scripts/add_electricity.py#L167-L172

The function returns `(efficiency_electrolysis ** 0.5) * (efficiency_fuelcell ** 0.5)` which is the same as `(efficiency_electrolysis * efficiency_fuelcell) ** 0.5`.

https://github.com/PyPSA/pypsa-eur/blob/9331f1ec46ceafd8aeffbbb655e3ec9acee3acdb/scripts/add_extra_components.py#L73-L83

This returned value is used for both `efficiency_store` and `efficiency_dispatch` resulting in the right roundtrip efficiency `efficiency_electrolysis * efficiency_fuelcell`. 

BUT: The issue is rather that `efficiency_store=efficiency_dispatch`, which is incorrect (for batteries it happens to be correct). It distorts the amount of energy capacity needed. I don't know what the reasoning was for this implementation, but I think taking account of different efficiencies of fuel cell and electrolysis in the StorageUnit component is something that should be fixed. I'll make a proposal.

pz-max:
@fneum thanks for the reply and for the clarification. 
I was not aware of the second part with "add_extra_components.py" and its meaning.
Though, the current design leads to misleading `cost.at ` data which is then a part of the `n.storage_unit` dataframe. 
In the  `n.storage_unit` dataframe, the store and dispatch efficiency shows actually for both:  `RT ** 0.5` rather than just `RT`. 
In the best case of course it would be if store efficiency in `n.storage_unit` dataframe shows the real store efficiency input rather then the RT.

Yes, `efficiency_store=efficiency_dispatch` seems to be false. I think we have output **rather than only the roundtrip `efficiency`** in the `add_electricity.py` script, **two variables `efficiency_store` and `efficiency_dispatch`.**

pz-max:
If you think the approach with two variables could work then I could propose a bit code as well

fneum:
> the store and dispatch efficiency shows actually for both: RT ** 0.5 rather than just RT
Yes, thus both together will yield the round-trip efficiency `RT ** 0.5 * RT ** 0.5 = RT`.

Sure, go ahead by further modifying this branch if you want. The battery modifications already look fine.

The following could work:

```py
def costs_for_storage(store, link1, link2=None, max_hours=1.):
        capital_cost = link1['capital_cost'] + max_hours * store['capital_cost']
        efficiency_dispatch = link1['efficiency']
        if link2 is not None:
            capital_cost += link2['capital_cost']
            efficiency_store = link2['efficiency']
        else:
            efficiency_store = link1['efficiency']
        return pd.Series(dict(capital_cost=capital_cost,
                              marginal_cost=0.,
                              efficiency_store=efficiency_store,
                              efficiency_dispatch=efficiency_dispatch,
                              co2_emissions=0.))
```

and then in `attach_storageunits()` take

```py
               efficiency_store=costs.at[carrier, 'efficiency_store'],
               efficiency_dispatch=costs.at[carrier, 'efficiency_dispatch'],
```

pz-max:
> _storageunits

Great. Exaclty what I had in mind.
I will test that these days and report on the results.

pz-max:
Hey @fneum ,
The proposed design lead to an error. Error screenshot below:
Screenshot from 2020-10-08 16-52-48

Just read an [ARTICLE](https://link.springer.com/article/10.1007/s40565-019-0548-z#Equ17) which might be related to the occured problem:
"It can be seen that the scheduling problem of Virtual Energy Storage Systems was usually solved with MILP model [5,6,7] which is time-consuming. **However, the precondition of using the LP model requires the bi-directional efficiency must be the same, which is not the real case**. To convexify the problem, a proof in [8] was given whereas we found it is only a necessary but not sufficient condition which may incur misunderstanding. Thus, a complimentary proof for [8] is given in this paper to give a general condition of using the LP model."

However, I must dig a bit deeper to understand the above. Do you have a quick idea of the problem/solution?

fneum:
closing in favour of #202

PR #183, opened 2020-08-25
Don't remove capital cost for lines and links when setting lv, lc limit

### LINKED ISSUE #175: Line capital costs removed for lv limit scenario
The `prepare_networks.py` script removes the capital costs for lines when setting line volume limits. By doing so, we somehow ensure that e.g the option lv1.25 really ends up with a 25% expansion. However I find it a bit misleading. In my point of view we should keep the capital cost  and 

1. either **fix** the lv to the certain expansion, i.e. change the sense of the constraint to ``==``
2. or leave the lv limit as an upper limit only, meaning in some cases it does not reach the limit. 

Any thoughts?

https://github.com/PyPSA/pypsa-eur/blob/8c5efb5252ec6da029e2ad9178ce019d1ed2dec1/scripts/prepare_network.py#L153

FabianHofmann:
Closes #175 

## Checklist

~- [ ] Changes in configuration options are added in all of `config.default.yaml`, `config.tutorial.yaml`, and `test/config.test1.yaml`.~
~- [ ] Changes in configuration options are also documented in `doc/configtables/*.csv` and line references are adjusted in `doc/configuration.rst` and `doc/tutorial.rst`.~

### DIFF
--- a/scripts/prepare_network.py
+++ b/scripts/prepare_network.py
@@ -9,8 +9,8 @@ Prepare PyPSA network for solving according to :ref:`opts` and :ref:`ll`, such a
 - adding an annual **limit** of carbon-dioxide emissions,
 - adding an exogenous **price** per tonne emissions of carbon-dioxide (or other kinds),
 - setting an **N-1 security margin** factor for transmission line capacities,
-- specifying a limit on the **cost** of transmission expansion,
-- specifying a limit on the **volume** of transmission expansion, and
+- specifying an expansion limit on the **cost** of transmission expansion,
+- specifying an expansion limit on the **volume** of transmission expansion, and
 - reducing the **temporal** resolution by averaging over multiple hours.
 
 Relevant Settings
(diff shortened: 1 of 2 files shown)


scripts/prepare_network.py
```diff
@@ -90,86 +91,48 @@ def add_emission_prices(n, emission_prices=None, exclude_co2=False):
...
+    ref = (lines_s_nom @ n.lines[col] +
+           n.links[links_dc_b].p_nom @ n.links[links_dc_b][col])
+
+    costs = load_costs(Nyears, snakemake.input.tech_costs,
+                       snakemake.config['costs'],
+                       snakemake.config['electricity'])
+    update_transmission_costs(n, costs, simple_hvdc_costs=False)
 
-    if lv == 'opt' or float(lv) > 1.0:
+    if factor == 'opt' or float(factor) > 1.0:
```
fneum, line 119: Wondering whether to add an assert statement or warning if factor is lower than 1. In this case transmission network is not extendable and no constraints are added.

fneum (approved):
Yes, I like that. Also makes the code simpler. Had just one thought regarding factor <= 1.0.

scripts/prepare_network.py
FabianHofmann: The code now support lv/lc limits below 1. Then no lower capacity limits are set but the constraint is still set.

PR #193, opened 2020-09-22
Custom busmap for clustering

### LINKED ISSUE #124: Option to provide custom bus mapping instead of kmeans clustering

> I have one question regarding the clustering of nodes. Is it only possible to do so with k-means, or can I also cluster them "by hand" telling the model which node goes into which zone?

> I would be interested to compare the zonal market in the EU today with a nodal one. 

Should not be too difficult to implement this.

Main question is how to provide busmaps.
- As `.csv`? But this is very volatile to upstream changes/reruns of the workflow.
- Providing shapefiles of zones and filtering by coordinates might be the more robust approach. However, maybe not as user-friendly.

Changes would likely have to occur around here:

https://github.com/PyPSA/pypsa-eur/blob/ef9e64d457686b7cd01e15795276bd26635a6656/scripts/cluster_network.py#L200-L233

eb5194:
closes #124 

Add a custom clustermaps, e.g. when running the same experiment with a different setting, but want to maintain the same clustering obtained from the first run to make results more comparable. Then, you can copy the clustermaps of your favor from "resources/" to "data/" and rename it to "custom_clustermaps_elec_s_{n_clusters}.h5"; and set "custom_clustermaps" in your config file to True.

**ToDo**: Test for an own busmapping, e.g. from a csv file? At the moment, busmap has to be in same format as clustermaps_elec_s_{n_clusters}.h5. Suggestions?

## Checklist

- [x] Newly introduced dependencies are added to `environment.yaml` and `environment.docs.yaml`.


fneum (changes requested):
I am aware this is still a draft, but one important suggestion regarding the file handling.

Also, for now keep the {network} wildcard in the filename. It will be removed soon, because all other sectors handled in PyPSA-Eur-Sec.

The custom mapping is from simplified network to clustered network. You can add support for different simplification levels (and consistency) by also adding the `{simpl}` wildcard to the filename.

Custom clustermaps for `simplify_network` rule probably don't make any sense, do they?

fneum:
> ToDo: Test for an own busmapping, e.g. from a csv file? At the moment, busmap has to be in same format as clustermaps_elec_s_{n_clusters}.h5. Suggestions?

Since you only need the custom busmaps and recalculate linemaps anyhow in `get_clustering_from_busmap(...)` [BTW is the function busmap->linemap unique/deterministic?], you could just provide the busmap as an input `.csv` file rather than the full `.h5` store.

The disadvantage is that you need to take manual action to reuse an existing clustermap generated by `cluster_network` with:

```py
import pandas as pd
with pd.HDFStore('resources/clustermaps_elec_s_X.h5', mode='r') as store:
    store.busmap.to_csv("data/custom_busmaps_elec_s_X.csv")
```

The advantage would be that it is much easier to create manual busmaps in a simple `pd.Series`:

```
name
7932    DE0 0
2332    DE0 0
        ...  
7961    DE0 1
8205    DE0 4
Name: busmap, Length: 317, dtype: object
```

Also, this doesn't store unnecessary data.

Then in `cluster_network` you just load the busmaps:

```py
busmapfornclusters = pd.read_csv(snakemake.input.custom_busmaps)
```

fneum:
You could also support both variants (`.h5` and `.csv`) by adding switches based on the file ending, but I think it's cleanest to only support `.csv`.

eb5194:
> Custom clustermaps for `simplify_network` rule probably don't make any sense, do they? Maybe it does.

No I think it makes no sense to customize the simplify script, unless its used with a number, such as elec_s1024

eb5194:
> [BTW is the function busmap->linemap unique/deterministic?]
Yes, should be.

About .csv vs .h5-format; might be worth to include both... not sure yet.

FabianHofmann:
Just for the convention: I think it would make more sense to name `custom_clustermaps` -> `custom_busmap`. This is what is used in pypsa and makes clear that it refers to spatial clustering (not temporal)

eb5194:
> Just for the convention: I think it would make more sense to name `custom_clustermaps` -> `custom_busmap`. This is what is used in pypsa and makes clear that it refers to spatial clustering (not temporal)

In that case, all reference clustermaps that are generated by pypsa-eur in `resources/ `should be renamed to `busmap` as well. Otherwise it will become confusing that pypsa-eur generates "`clustermaps`", but the custom ones are named "`busmap`".

fneum:
Possible distinction here:

- `clustermaps` for all `.h5` since they include both busmaps and linemaps
- `busmaps` for all `.csv` sind they only include the busmap.

fneum (changes requested):
Ok, almost there! Tests fail because

```yaml
clustering:
  custom_clustermaps: False
```

is not also in `test/config.test1.yaml` and `config.tutorial.yaml`.
Then all is left to do is going through the checklist (esp. point 5)

eb5194:
> Then all is left to do is going through the checklist (esp. point 5)

How about a custom busmap for the sites? I.e. elec_s1000_37m.nc. Then 1000 sites can be read from the custom busmap and the transmission grid is at 37 nodes/one node per country. Of course, in this case both could be customized.

We could also do that in a separate PR.

fneum:
> How about a custom busmap for the sites?

If you need this, why not. But in this case it should be a separate PR.

fneum:
status @eb5194, good for another review iteration?

eb5194:
> status @eb5194, good for another review iteration?

Should be, yes! Thanks

scripts/cluster_network.py
```diff
@@ -267,10 +269,9 @@ def plot_busmap_for_n_clusters(n, n_clusters=50):
     n.plot(bus_colors=busmap.map(dict(zip(cs, cr))))
     del cs, cr
 
-def clustering_for_n_clusters(n, n_clusters, aggregate_carriers=None,
-                              line_length_factor=1.25, potential_mode='simple',
-                              solver_name="cbc", algorithm="kmeans",
-                              extended_link_costs=0, focus_weights=None):
+def clustering_for_n_clusters(n, n_clusters, busmapfornclusters=None, aggregate_carriers=None,
```
fneum, line 272:
```suggestion
def clustering_for_n_clusters(n, n_clusters, custom_busmap=False, aggregate_carriers=None,
```

scripts/cluster_network.py
```diff
@@ -280,8 +281,16 @@ def clustering_for_n_clusters(n, n_clusters, aggregate_carriers=None,
         raise AttributeError("potential_mode should be one of 'simple' or 'conservative', "
                              "but is '{}'".format(potential_mode))
 
+    if busmapfornclusters is None: # make sure busmap is given outside of cluster_network.py
+        if snakemake.config['clustering']['custom_busmap']:
+            busmapfornclusters = (pd.read_csv(snakemake.input.custom_busmap, dtype={'name':'str'})
+                                  .set_index('name')).squeeze()
+            logger.info(f"Imported custom busmap from {snakemake.input.custom_busmap}")
+        else:
+            busmapfornclusters = busmap_for_n_clusters(n, n_clusters, solver_name, focus_weights, algorithm)
```
fneum, line 290:
```suggestion
    if custom_busmap:
        busmapfornclusters = (pd.read_csv(snakemake.input.custom_busmap, dtype={'name':'str'})
                              .set_index('name')).squeeze()
        logger.info(f"Imported custom busmap from {snakemake.input.custom_busmap}")
    else:
        busmapfornclusters = busmap_for_n_clusters(n, n_clusters, solver_name, focus_weights, algorithm)
```

scripts/simplify_network.py
```diff
@@ -86,7 +86,7 @@
 logger = logging.getLogger(__name__)
 from _helpers import configure_logging
 
-from cluster_network import clustering_for_n_clusters, cluster_regions
+from cluster_network import clustering_for_n_clusters, cluster_regions, busmap_for_n_clusters
```
fneum, line 89:
```suggestion
from cluster_network import clustering_for_n_clusters, cluster_regions
```

scripts/simplify_network.py
```diff
@@ -329,7 +329,10 @@ def consense(x):
     potential_mode = (consense(pd.Series([snakemake.config['renewable'][tech]['potential']
                                             for tech in renewable_carriers]))
                         if len(renewable_carriers) > 0 else 'conservative')
-    clustering = clustering_for_n_clusters(n, n_clusters, potential_mode=potential_mode,
+
+    busmapfornclusters = busmap_for_n_clusters(n, n_clusters, snakemake.config['solving']['solver']['name'])
+        
+    clustering = clustering_for_n_clusters(n, n_clusters, busmapfornclusters, potential_mode=potential_mode,
```
fneum, line 335:
```suggestion
    clustering = clustering_for_n_clusters(n, n_clusters, custom_busmap=False, potential_mode=potential_mode,
```

scripts/cluster_network.py
```diff
@@ -363,7 +372,7 @@ def consense(x):
             return v
         potential_mode = consense(pd.Series([snakemake.config['renewable'][tech]['potential']
                                              for tech in renewable_carriers]))
-        clustering = clustering_for_n_clusters(n, n_clusters, aggregate_carriers,
+        clustering = clustering_for_n_clusters(n, n_clusters, aggregate_carriers=aggregate_carriers,
```
fneum, line 375:
```suggestion
        custom_busmap = snakemake.config["clustering"]["custom_busmap"]
        clustering = clustering_for_n_clusters(n, n_clusters, custom_busmap, aggregate_carriers,
```

fneum (changes requested):
@eb5194 here's a few suggestions to further simplify the code (esp. avoid layered `if-else` in `cluster_network.py` and avoid call of `busmap_for_n_clusters()` in `simplify_network.py`.

The function `clustering_for_n_clusters()` just gets a `custom_busmap=False` flag that triggers loading the alternative `busmapfornclusters`.

fneum:
And need to address the two last items of the checklist, i.e.: documentation of configuration settings, add release notes.

PR #209, opened 2020-11-09
prepare: fix transmission limit when no HVDC links in network

fneum:
Accessing via `n.links[]` goes through columns if dataframe is empty, and throws `'DataFrame' object has no attribute 'p_nom'`. Now using `.loc` to be more precise regarding columns/rows.


FabianHofmann (approved):
makes sense

PR #219, opened 2021-01-20
simplify: delete bus columns with incorrect entries

martacki:
dead-end busses with possibly different `substation_lv`, `substation_off` and `under_construction` values are being aggregated, but the dead-end bus is simply removed without updating the respective attribute of the bus it is aggregated to.

since we don't need those attributes (they are non-existent after `cluster_network`), I suggest to drop them to avoid confusion


fneum:
Possible to do. Under what circumstances would this be actually incorrect?

These attributes as well as `symbol` and `tags` are removed in `cluster_network` anyway (do you see where exactly? Couldn't find it on the spot. We could do this analogously in `simplify_network`). If we decide to remove them already in `simplify_network` (the fact that you can already pre-cluster speaks in favour of this), maybe these two additional fields should also be removed.

martacki:
> Possible to do. Under what circumstances would this be actually incorrect?
> 
this happens when assigning an univalent bus with a different value of substation_lv/under_construction/substation_off to it's adjacent neighbor. The attributes are not updated.

> These attributes as well as `symbol` and `tags` are removed in `cluster_network` anyway (do you see where exactly? Couldn't find it on the spot. We could do this analogously in `simplify_network`).

If I remember correctly, this happens in pypsa in get_clustering_from_busmap or so, because the aggregation does not consider these attributes and therefore, the output does not contain them anymore. n.buses, n.lines, etc. are then overwritten with the new output (clustering.network).

> If we decide to remove them already in `simplify_network` (the fact that you can already pre-cluster speaks in favour of this), maybe these two additional fields should also be removed.

Sure! But let me test if the code works if the `simpl` wildcard is used, just came to my mind that this may cause a problem.

martacki:
> Sure! But let me test if the code works if the `simpl` wildcard is used, just came to my mind that this may cause a problem.

Indeed was causing a bug. Resolved and now dropping `tags` and `symbol` as well

fneum:
Can go in once tests pass.

PR #169, opened 2020-07-07
switch to upstream glaes/geokit

fneum:
Closes #156.

## Changes proposed in this Pull Request

Use https://github.com/FZJ-IEK3-VSA/glaes and https://github.com/FZJ-IEK3-VSA/geokit not the PyPSA forks.

## Checklist

- [x] I tested my contribution locally and it seems to work fine.
- [ ] Code and workflow changes are sufficiently documented.
- [x] Newly introduced dependencies are added to `environment.yaml` and `environment.docs.yaml`.
~- [ ] Changes in configuration options are added in all of `config.default.yaml`, `config.tutorial.yaml`, and `test/config.test1.yaml`.~
~- [ ] Changes in configuration options are also documented in `doc/configtables/*.csv` and line references are adjusted in `doc/configuration.rst` and `doc/tutorial.rst`.~
- [ ] A note for the release notes `doc/release_notes.rst` is amended in the format of previous release notes.

PR #147, opened 2020-03-19
Update to new atlite version (v0.2).

euronion:
## Changes proposed in this Pull Request

This PR is intended to update PyPSA-EUR to use the new version of atlite, currently being prepared for release in https://github.com/PyPSA/atlite/pull/20.

## Todo

I have identified the following todos, the list might be missing some. Please comment/add if you identify further ones.

- [ ] Adjust all rules directly using `atlite`
- [ ] Adjust all rules using cutouts created by `atlite`, as the cutout structure changes
- [ ] Provide new zenodo repository for cutout downloads and update the rule `retrieve_cutout` accordingly.
- [ ] Crosspost / reference comparison between renewable profiles generated with old and new `atlite`.

## Checklist

- [ ] I tested my contribution locally and it seems to work fine.
- [ ] Code and workflow changes are sufficiently documented.
- [ ] Newly introduced dependencies are added to `environment.yaml` and `environment.docs.yaml`.
- [ ] Changes in configuration options are added in all of `config.default.yaml`, `config.tutorial.yaml`, and `test/config.test1.yaml`.
- [ ] Changes in configuration options are also documented in `doc/configtables/*.csv` and line references are adjusted in `doc/configuration.rst` and `doc/tutorial.rst`.
- [ ] A note for the release notes `doc/release_notes.rst` is amended in the format of previous release notes.

Issue #158, opened 2020-05-26
assert network time and spatial bounds are in cutout


FabianHofmann:
We should add an assertion for the following for 

* all snapshots of the base.nc being in the cutout time selection
* the whole spatial area is covered be the cutout.bounds

FabianHofmann:
This is the current bounding box for all grid cells (default config) and the offshore area, it seems not right that some of the offshore area remains uncovered. However I still have to figure out how the maximal distance of wind turbines plays out here.
![geomap](https://user-images.githubusercontent.com/19226431/82901507-503ae280-9f5e-11ea-9021-72fa222bcfb4.png)

FabianHofmann:
After a small discussion with @euronion we agreed that it would be better to double check such configuration manners in a dedicated file. 

Such would be named `check_config.py` and would contain basic assertions for consistency and validation. Atm for such a case as above where you to double check a ressource, it would probably contain function which you called here.

fneum:
Could be called in `Snakefile` after

https://github.com/PyPSA/pypsa-eur/blob/4d0a86024646f3a468c4fcb6c40dc435b8ab11f5/Snakefile#L11-L12

```py
configfile: "config.yaml"

assert check_config(config), "Configfile not correct."
```

euronion:
In the PR we came across options in `config.yaml` which became deprecated.
It would be a nice addition to have the `check_config(...)` also warn about options in the config file which are deprecated (or soon will).

euronion:
`snakemake` offers some basic config validation using a JSON file ([details](https://snakemake.readthedocs.io/en/stable/snakefiles/configuration.html#snakefiles-config-validation)).
This could partially take care of validation (correct structure and types), deprecation and consistency validation between options cannot be outsoured to this functionality (I think).

PR #224, opened 2021-03-01
Atlite availability

### LINKED ISSUE #132: easy SARAH-2 weather data integration for years other than 2013
see thread at https://groups.google.com/g/pypsa/c/LbGBAruL_Hs

### LINKED ISSUE #156: Update GLAES to current version

Update the `glaes` dependency to the current version and drop the fixed dependency on the `pypsa/glaes` fork. The single hotfix done in `pypsa/glaes` seems to have been implemented upstream in this [commit](https://github.com/FZJ-IEK3-VSA/glaes/commit/9454584b983ffa09eaa18c0fca25eb3ad1579bf6) .

## Current hinderance:
`ec.ExcludeRasterType(...)` seems to support only ranges for raster types in the new version.
But we're using a list of values to be included via `invert=True` here https://github.com/PyPSA/pypsa-eur/blob/bb3477cd693d1c8e77e75e61a1a7e1a4647e6a3c/scripts/build_renewable_profiles.py#L219

It does not seem as if this was ever officially supported.

I can't find a quick solution at the moment to invert the selection, i.e.
1. retrieve a set of `grid_codes` from corine 
2. do the inversion outside of `glaes`
3. Exclude the inverted set of `grid_codes`

### LINKED ISSUE #162: Introduce atlite v0.2
As we are approaching the release of atlite v0.2, there is some direct implications for pypsa-eur-(sec)

# Environment
pin the environment to the current version the master branched in [atlite v0.0.3](https://github.com/PyPSA/atlite/tree/v0.0.3). This should also happen in pypsa-eur-sec. @fneum could you take over this part? 

# Renewable profiles
As pointed out in https://github.com/PyPSA/atlite/issues/81#issuecomment-639190595 there might be some minimal invasive ways to account for the new convention in atlite which we should try out. @euronion you already tried those? At the moment I am not sure how this convention change, might inflict with the new version of glaes #156 ? We should tackle this in one strike

### LINKED ISSUE #230: Error in rule build_shapes:  jobid: 7
## Checklist

- [Y] I am using the current [`master`](https://github.com/PyPSA/pypsa-eur/tree/master) branch or the latest [release](https://github.com/PyPSA/pypsa-eur/releases). Please indicate.
- [Y] I am running on an up-to-date [`pypsa-eur` environment](https://github.com/PyPSA/pypsa-eur/blob/master/envs/environment.yaml). Update via `conda env update -f envs/environment.yaml`.

## Describe the Bug

Installing and running the Tutorial config of pyPSA-EUR as described. Unfortunately, I am meeting the following error on a OSX  running Mojave. I am using PyCharm and Conda for my creating my environments. 

Any idea with this is happening ? 

Thanks for your help

## Error Message

*If applicable, paste any terminal output to help illustrating your problem.*
*In some cases it may also be useful to share your list of installed packages: `conda list`.*

```
Building DAG of jobs...
Using shell: /bin/bash
Provided cores: 1 (use --cores to define parallelism)
Rules claiming more threads will be scaled down.
Job counts:
        count   jobs
        1       add_electricity
        1       add_extra_components
        1       base_network
        1       build_bus_regions
        1       build_powerplants
        4       build_renewable_profiles
        1       build_shapes
        1       cluster_network
        1       prepare_network
        1       retrieve_cutout
        1       retrieve_natura_raster
        1       simplify_network
        1       solve_network
        16
Select jobs to execute...

[Sat Mar 13 15:44:06 2021]
rule build_shapes:
    input: data/bundle/naturalearth/ne_10m_admin_0_countries.shp, data/bundle/eez/World_EEZ_v8_2014.shp, data/bundle/NUTS_2013_60M_SH/data/NUTS_RG_60M_2013.shp, data/bundle/nama_10r_3popgdp.tsv.gz, data/bundle/nama_10r_3gdp.tsv.gz, data/bundle/ch_cantons.csv, data/bundle/je-e-21.03.02.xls
    output: resources/country_shapes.geojson, resources/offshore_shapes.geojson, resources/europe_shape.geojson, resources/nuts3_shapes.geojson
    log: logs/build_shapes.log
    jobid: 7
    resources: mem=500

Traceback (most recent call last):
  File "/Users/eromano/PycharmProjects/pypsa-eur-master/.snakemake/scripts/tmpmnirhu7l.build_shapes.py", line 227, in <module>
    country_shapes = countries()
  File "/Users/eromano/PycharmProjects/pypsa-eur-master/.snakemake/scripts/tmpmnirhu7l.build_shapes.py", line 118, in countries
    df = gpd.read_file(snakemake.input.naturalearth)
  File "/opt/anaconda3/envs/pypsa-eur/lib/python3.8/site-packages/geopandas/io/file.py", line 127, in _read_file
    _check_fiona("'read_file' function")
  File "/opt/anaconda3/envs/pypsa-eur/lib/python3.8/site-packages/geopandas/io/file.py", line 42, in _check_fiona
    raise ImportError(
ImportError: the 'read_file' function requires the 'fiona' package, but it is not installed or does not import correctly.
Importing fiona resulted in: dlopen(/opt/anaconda3/envs/pypsa-eur/lib/python3.8/site-packages/fiona/ogrext.cpython-38-darwin.so, 2): Library not loaded: @rpath/libtbb.dylib
  Referenced from: /opt/anaconda3/envs/pypsa-eur/lib/libtiledb.dylib
  Reason: image not found
[Sat Mar 13 15:44:07 2021]
Error in rule build_shapes:
    jobid: 7
    output: resources/country_shapes.geojson, resources/offshore_shapes.geojson, resources/europe_shape.geojson, resources/nuts3_shapes.geojson
    log: logs/build_shapes.log (check log file(s) for error message)

RuleException:
CalledProcessError in line 122 of /Users/eromano/PycharmProjects/pypsa-eur-master/Snakefile:
Command 'set -euo pipefail;  /opt/anaconda3/envs/pypsa-eur/bin/python3.8 /Users/eromano/PycharmProjects/pypsa-eur-master/.snakemake/scripts/tmpmnirhu7l.build_shapes.py' returned non-zero exit status 1.
  File "/opt/anaconda3/envs/pypsa-eur/lib/python3.8/site-packages/snakemake/executors/__init__.py", line 2326, in run_wrapper
  File "/Users/eromano/PycharmProjects/pypsa-eur-master/Snakefile", line 122, in __rule_build_shapes
  File "/opt/anaconda3/envs/pypsa-eur/lib/python3.8/site-packages/snakemake/executors/__init__.py", line 568, in _callback
  File "/opt/anaconda3/envs/pypsa-eur/lib/python3.8/concurrent/futures/thread.py", line 57, in run
  File "/opt/anaconda3/envs/pypsa-eur/lib/python3.8/site-packages/snakemake/executors/__init__.py", line 554, in cached_or_run
  File "/opt/anaconda3/envs/pypsa-eur/lib/python3.8/site-packages/snakemake/executors/__init__.py", line 2357, in run_wrapper
Shutting down, this might take some time.
Exiting because a job execution failed. Look above for error message
Complete log: /Users/eromano/PycharmProjects/pypsa-eur-master/.snakemake/log/2021-03-13T154405.273426.snakemake.log


```

FabianHofmann:
closes #162 
closes #156
closes #132 
closes #169 
closes #147 
closes #230 

Switch to new `atlite` version `v0.2.1`. The version upgrade comes along with significant speed up for the rule `build_renewable_profiles.py` (~factor 2). A lot of the code which calculated the landuse availability is now outsourced and does not rely on `glaes`, `geokit` anymore. This facilitates the environment building and version compatibility of `gdal`, `libgdal` with other packages.     

It requires an update of the cutout files at zenodo. 

The new implementation is not able to exactly reproduce the old profiles, but very close. **Reasons:**
1. The differences result from a slightly different approach of the distance calculations from raster points: atlite uses the dilation function of `scipy.ndimage`, glaes made a conversion of the raster to geometries -> calculating a buffer -> backconverting to a raster. Non of the approach is objectively better, their results diverge just a very little.    
2. The `gdal` resampling method `average`, is not purely consistent as their might be a small subpixel shift (https://github.com/OSGeo/gdal/issues/2235). 

## Open question

* Do we still need the piezker script (`build_country_flh.py`)? I did not update it. If wanted please shout out. Otherwise we could delete the rule.  

## Checklist


fneum:
All in for deleting/archiving `build_country_flh.py`

euronion:
Aye in favour for retiring `build_country_flh.py`. I'll review in the next days.

fneum:
I'll use `build_cutout=True`, but do you @FabianHofmann already have a new default cutout file?

fneum:
Ok, so first batch is no code review but results/experience review (laptop, <8 threads):

### speed and memory

time-wise much faster, memory-wise a bit worse (8.5 GB vs 6.5 GB).

| carrier    |   new |   old |
|:-----------|------:|------:|
| onwind | 5:40 | 24:30 |
| solar | 6:10 | 18:40 |

this is really cool!

### country-level potentials

shows new/old version, deviation mostly negligible but

- onwind in BE and LU?
- offwind-dc in ES and EE?
- DK seems to exchange offwind DC for AC.

potentials

### distribution of average capacity factors

most notable for offwind-ac but otherwise no complaints

avg-cfs

means extremely close!

| carrier    |   new |   old |
|:-----------|------:|------:|
| offwind-ac | 31.21 | 31.97 |
| offwind-dc | 45.59 | 45.97 |
| onwind     | 18.7  | 18.69 |
| ror        | 53.91 | 53.89 |
| solar      | 11.94 | 11.93 |

fneum:
@euronion if you could focus your review on building the cutout that merges ERA5 and SARAH for solar that would be really great!

envs/environment.yaml
```diff
@@ -29,32 +30,14 @@ dependencies:
...
-  - shapely
-  - libgdal<=3.0.4
   - descartes
 
   - pip:
     - vresutils==0.3.1
     - tsam>=1.1.0
-    - git+https://github.com/PyPSA/glaes.git#egg=glaes
-    - git+https://github.com/PyPSA/geokit.git#egg=geokit
-    - cdsapi
```
fneum, line 60: we could keep `cdsapi`

.travis.yml
```diff
@@ -29,6 +29,9 @@ before_install:
   # list packages for easier debugging
   - conda list
 
+before_script:
+  - 'echo -ne "url: ${CDSAPI_URL}\nkey: ${CDSAPI_TOKEN}\n" > ~/.cdsapirc'
```
fneum, line 33: secrets are stored?

scripts/build_cutout.py
```diff
@@ -92,28 +88,38 @@
...
+    time = [snapshots[0], snapshots[-1]]
+    cutout_params['time'] = slice(*cutout_params.get('time', time))
+
+    if {'x', 'y', 'bounds'}.isdisjoint(cutout_params):
+        regions = (gpd.read_file(snakemake.input.regions_offshore).append(
+                   gpd.read_file(snakemake.input.regions_onshore)))
+        cutout_params['bounds'] = regions.cascaded_union.buffer(0.5).bounds
+    elif {'x', 'y'}.issubset(cutout_params):
+        cutout_params['x'] = slice(*cutout_params['x'])
+        cutout_params['y'] = slice(*cutout_params['y'])
```
fneum, line 119: So if you specify no bounds, you automatically create a cutout of the bounding box of considered countries? Cool!

Snakefile
```diff
@@ -148,7 +150,7 @@ if config['enable'].get('build_cutout', False):
 
 if config['enable'].get('retrieve_cutout', True):
     rule retrieve_cutout:
-        output: directory(expand("cutouts/{cutouts}", **config['atlite'])),
+        output: expand("cutouts/{cutouts}.nc", **config['atlite'])
```
fneum, line 153: :+1: for switch from directory to single .nc file

config.default.yaml
```diff
@@ -62,18 +62,28 @@ electricity:
...
     europe-2013-era5:
-      module: era5
-      xs: [-12., 35.]
-      ys: [72., 33.]
-      years: [2013, 2013]
+      module: era5 # in priority order 
+      x: [-12., 35.]
+      y: [33., 72]
+      dx: 0.3
+      dy: 0.3
```
fneum, line 73: That's the target raster resolution, right?

config.default.yaml
```diff
@@ -62,18 +62,28 @@ electricity:
...
+      time: ['2013', '2013']
```
fneum, line 74: This can be a pandas datetime-like string, right? Like "2013-03"?

config.default.yaml
```diff
@@ -62,18 +62,28 @@ electricity:
...
-      years: [2013, 2013]
+      module: [sarah, era5] # in priority order 
+      x: [-12., 45.]
+      y: [33., 65]
+      dx: 0.2
+      dy: 0.2
+      time: ['2013', '2013']
+      sarah_interpolate: false
+      sarah_dir: 
+      features: [influx, temperature]
```
fneum, line 84: For some of these new attributes it would be good to add documentation to https://pypsa-eur.readthedocs.io/en/latest/configuration.html

doc/release_notes.rst
```diff
@@ -11,6 +11,8 @@ Release Notes
 Upcoming Release
 ================
 
+* Switch to new `atlite` version `v0.2.1`. The version upgrade comes along with significant speed up for the rule `build_renewable_profiles.py` (~factor 2). A lot of the code which calculated the landuse availability is now outsourced and does not rely on `glaes`, `geokit` anymore. This facilitates the environment building and version compatibility of `gdal`, `libgdal` with other packages.
```
fneum, line 14: .rst want double quotation marks for code :)

scripts/build_renewable_profiles.py
```diff
@@ -60,7 +60,7 @@
     **Source:** `GEBCO <https://www.gebco.net/data_and_products/images/gebco_2019_grid_image.jpg>`_
 
 - ``resources/natura.tiff``: confer :ref:`natura`
-- ``resources/country_shapes.geojson``: confer :ref:`shapes`
+- ``resources/ohshore.tiff``: confer :ref:`onshore`
```
fneum, line 63: Is this right?

country shapes is still in the Snakefile and I haven't seen a `onshore.tiff` yet (besides the typo)

scripts/build_renewable_profiles.py
```diff
@@ -180,220 +180,158 @@
...
-
-from scipy.sparse import csr_matrix, vstack
+import geopandas as gpd
+import xarray as xr
+import numpy as np
+import atlite
+import logging
 from pypsa.geo import haversine
-from vresutils import landuse as vlanduse
-from vresutils.array import spdiag
```
fneum, line 199: nice side effect!

scripts/build_renewable_profiles.py
```diff
@@ -180,220 +180,158 @@
...
+        logger.info('Calculate landuse availabilities...')
+        start = time.time()
+        availability = cutout.availabilitymatrix(regions, excluder, **kwargs)
+        duration = time.time() - start
+        logger.info(f'Completed availability calculation ({duration:2.2f}s)')
+    else:
+        availability = cutout.availabilitymatrix(regions, excluder, **kwargs)
 
-    config = snakemake.config['renewable'][snakemake.wildcards.technology]
+    area = cutout.grid.to_crs({'proj': 'cea'}).area / 1e6
```
fneum, line 264: I think one could make this tidier with
```suggestion
    area = cutout.grid.to_crs(epsg=xxxx).area / 1e6
```

test/config.test1.yaml
```diff
@@ -26,8 +26,8 @@ snapshots:
 enable:
   prepare_links_p_nom: false
   retrieve_databundle: true
-  build_cutout: false
-  retrieve_cutout: true
+  build_cutout: true
+  retrieve_cutout: false
```
fneum, line 30: cool, is that something that consumes time while testing?

In general the speedups of the build_renewables_profiles should greatly speedup the CI!

fneum (commented):
The `build_renewable_profiles` looks SO MUCH cleaner! Only small comments, while browsing through.

- [ ] remove `build_flh`
- [ ] new cutouts on zenodo (I can upload those, I have the SARAH already, ERA5 comes soon I imagine)

envs/environment.yaml
fneum: ah, or is that an `atlite` dependency?

.travis.yml
fneum: obviously. tests ran through

envs/environment.yaml
euronion: only `atlite` AFAIK and enforced in the new version.

euronion:
Building cutouts:
* ERA5 works (produces an output, will check the quality as soon as I have the SARAH cutout)
* SARAH doesn't work, yielding the following error:
```
(pypsa-eur-development) pypsa@vm-pypsa:~/share/GitHub/pypsa-eur_development$ snakemake -j8 cutouts/europe-2013-sarah.nc
Building DAG of jobs...
Using shell: /usr/bin/bash
Provided cores: 8
Rules claiming more threads will be scaled down.
Job counts:
        count   jobs                                                                                                                                                                                                                                                                                                                                 
        1       build_cutout                                                                                                                                                                                                                                                                                                                         
        1                                                                                                                                                                                                                                                                                                                                            
Select jobs to execute...

[Tue Mar  2 17:43:34 2021]
rule build_cutout:
    input: resources/regions_onshore.geojson, resources/regions_offshore.geojson                                                                                                                                                                                                                                                                     
    output: cutouts/europe-2013-sarah.nc                                                                                                                                                                                                                                                                                                             
    log: logs/build_cutout/europe-2013-sarah.log                                                                                                                                                                                                                                                                                                     
    jobid: 0                                                                                                                                                                                                                                                                                                                                         
    benchmark: benchmarks/build_cutout_europe-2013-sarah                                                                                                                                                                                                                                                                                             
    wildcards: cutout=europe-2013-sarah                                                                                                                                                                                                                                                                                                              
    threads: 4                                                                                                                                                                                                                                                                                                                                       
    resources: mem=4000                                                                                                                                                                                                                                                                                                                              

INFO:root:Preparing cutout with parameters {'module': ['sarah', 'era5'], 'x': slice(-12.0, 45.0, None), 'y': slice(33.0, 65, None), 'dx': 0.2, 'dy': 0.2, 'time': slice('2013', '2013', None), 'sarah_interpolate': False, 'sarah_dir': '/home/pypsa/share/reanalysis/datasets/SARAH-2/extracted', 'features': ['influx', 'temperature']}.
INFO:atlite.cutout:Building new cutout cutouts/europe-2013-sarah.nc
INFO:atlite.data:Storing temporary files in /tmp/tmp38ddxxwh
INFO:atlite.data:Calculating and writing with module sarah:
Traceback (most recent call last):
  File "/home/pypsa/share/GitHub/pypsa-eur_development/.snakemake/scripts/tmpj1joqdxc.build_cutout.py", line 129, in <module>
    cutout.prepare(features=features)
  File "/home/pypsa/miniconda3/envs/pypsa-eur-development/lib/python3.8/site-packages/atlite/data.py", line 94, in wrapper
    res = func(*args, **kwargs)
  File "/home/pypsa/miniconda3/envs/pypsa-eur-development/lib/python3.8/site-packages/atlite/data.py", line 155, in cutout_prepare
    ds = get_features(cutout, module, missing_features, tmpdir=tmpdir)
  File "/home/pypsa/miniconda3/envs/pypsa-eur-development/lib/python3.8/site-packages/atlite/data.py", line 44, in get_features
    datasets = compute(*datasets)
  File "/home/pypsa/miniconda3/envs/pypsa-eur-development/lib/python3.8/site-packages/dask/base.py", line 563, in compute
    results = schedule(dsk, keys, **kwargs)
  File "/home/pypsa/miniconda3/envs/pypsa-eur-development/lib/python3.8/site-packages/dask/threaded.py", line 76, in get
    results = get_async(
  File "/home/pypsa/miniconda3/envs/pypsa-eur-development/lib/python3.8/site-packages/dask/local.py", line 487, in get_async
    raise_exception(exc, tb)
  File "/home/pypsa/miniconda3/envs/pypsa-eur-development/lib/python3.8/site-packages/dask/local.py", line 317, in reraise
    raise exc
  File "/home/pypsa/miniconda3/envs/pypsa-eur-development/lib/python3.8/site-packages/dask/local.py", line 222, in execute_task
    result = _execute_task(task, data)
  File "/home/pypsa/miniconda3/envs/pypsa-eur-development/lib/python3.8/site-packages/dask/core.py", line 121, in _execute_task
    return func(*(_execute_task(a, cache) for a in args))
  File "/home/pypsa/miniconda3/envs/pypsa-eur-development/lib/python3.8/site-packages/dask/utils.py", line 35, in apply
    return func(*args, **kwargs)
  File "/home/pypsa/miniconda3/envs/pypsa-eur-development/lib/python3.8/site-packages/atlite/datasets/sarah.py", line 190, in get_data
    ds = regrid(ds, coords['lon'], coords['lat'], resampling=Resampling.average)
  File "/home/pypsa/miniconda3/envs/pypsa-eur-development/lib/python3.8/site-packages/atlite/gis.py", line 609, in regrid
    return (xr.apply_ufunc(_reproject,
  File "/home/pypsa/miniconda3/envs/pypsa-eur-development/lib/python3.8/site-packages/xarray/core/computation.py", line 1092, in apply_ufunc
    return apply_dataset_vfunc(
  File "/home/pypsa/miniconda3/envs/pypsa-eur-development/lib/python3.8/site-packages/xarray/core/computation.py", line 410, in apply_dataset_vfunc
    result_vars = apply_dict_of_variables_vfunc(
  File "/home/pypsa/miniconda3/envs/pypsa-eur-development/lib/python3.8/site-packages/xarray/core/computation.py", line 356, in apply_dict_of_variables_vfunc
    result_vars[name] = func(*variable_args)
  File "/home/pypsa/miniconda3/envs/pypsa-eur-development/lib/python3.8/site-packages/xarray/core/computation.py", line 653, in apply_variable_ufunc
    raise ValueError(
ValueError: dimension 'dim3' in 'output_sizes' must correspond to output_core_dims
[Tue Mar  2 17:44:00 2021]
Error in rule build_cutout:
    jobid: 0
    output: cutouts/europe-2013-sarah.nc
    log: logs/build_cutout/europe-2013-sarah.log (check log file(s) for error message)

RuleException:
CalledProcessError in line 148 of /home/pypsa/share/GitHub/pypsa-eur_development/Snakefile:                                                                                                                                                                                                                                                          
Command 'set -euo pipefail;  /home/pypsa/miniconda3/envs/pypsa-eur-development/bin/python3.8 /home/pypsa/share/GitHub/pypsa-eur_development/.snakemake/scripts/tmpj1joqdxc.build_cutout.py' returned non-zero exit status 1.                                                                                                                         
  File "/home/pypsa/miniconda3/envs/pypsa-eur-development/lib/python3.8/site-packages/snakemake/executors/__init__.py", line 2272, in run_wrapper                                                                                                                                                                                                    
  File "/home/pypsa/share/GitHub/pypsa-eur_development/Snakefile", line 148, in __rule_build_cutout                                                                                                                                                                                                                                                  
  File "/home/pypsa/miniconda3/envs/pypsa-eur-development/lib/python3.8/site-packages/snakemake/executors/__init__.py", line 568, in _callback                                                                                                                                                                                                       
  File "/home/pypsa/miniconda3/envs/pypsa-eur-development/lib/python3.8/concurrent/futures/thread.py", line 57, in run                                                                                                                                                                                                                               
  File "/home/pypsa/miniconda3/envs/pypsa-eur-development/lib/python3.8/site-packages/snakemake/executors/__init__.py", line 554, in cached_or_run                                                                                                                                                                                                   
  File "/home/pypsa/miniconda3/envs/pypsa-eur-development/lib/python3.8/site-packages/snakemake/executors/__init__.py", line 2357, in run_wrapper                                                                                                                                                                                                    
Shutting down, this might take some time.
Exiting because a job execution failed. Look above for error message
Complete log: /home/pypsa/share/GitHub/pypsa-eur_development/.snakemake/log/2021-03-02T174333.798825.snakemake.log
```

Any ideas were this could come from?

FabianHofmann:
Yes, wrong xarray version. It has to be 0.16.2 (they actually have a lot of breaking changes). I'll fix it in the environment (it's also fixed in the new conda build)

envs/environment.yaml
FabianHofmann: yes!

.travis.yml
FabianHofmann: yep (added my own)

scripts/build_cutout.py
FabianHofmann: indeed, also works for non-defined time-spans. So leaving everything blank does everything you need (in standard 0.25 x 0.25 resolution).

config.default.yaml
FabianHofmann: Exactly, later we can think about increasing the resolution of the era5 raster (0.25x0.25). For a better comparison, I left it this way.

config.default.yaml
FabianHofmann: Correct

scripts/build_renewable_profiles.py
FabianHofmann: Good catch, that is an artifact from previous attempts to compute a distance field for onshore regions.

scripts/build_renewable_profiles.py
FabianHofmann: Getting there ;)

test/config.test1.yaml
FabianHofmann: Yes, that's temporarily for the CI. Actually, once it's in the 'short term memory' of (in this case) my ecmwf account, retrieving data is very fast. Holds for roughly 2 days (from what I experienced).

euronion:
> 
> 
> Yes, wrong xarray version. It has to be 0.16.2 (they actually have a lot of breaking changes). I'll fix it in the environment (it's also fixed in the new conda build)

That did the trick. Thanks!

scripts/build_renewable_profiles.py
FabianHofmann: done

config.default.yaml
FabianHofmann: is included (seems I have to fix something the referenced atlite doc)

euronion:
With `xarray=0.16.2` the `cutout` creation works smoothly.

For comparison the here are the extents and resolutions:

```
==== era (old) ====
lat: 33.0 <-> 72.00
lon: -12.00 <-> 35.10
Resolution: 0.30 (lon) / -0.30 (lat)
==== era (new) ====
lat: 33.0 <-> 72.00
lon: -12.00 <-> 34.80
Resolution: 0.30 (lon) / 0.30 (lat)
==== sarah (old) ====
lat: 33.0 <-> 65.00
lon: -12.00 <-> 42.20
Resolution: 0.20 (lon) / -0.20 (lat)
==== sarah (new) ====
lat: 33.0 <-> 65.00
lon: -12.00 <-> 45.00
Resolution: 0.20 (lon) / 0.20 (lat)
```

The maximum longitudes don't match, the old `era5` was rounded up (35.1 is the nearest to 35 degress with a resolution of .3), the new `era5` rounded down to 34.8.

I compared the cutouts based on some simple statistics. Except for the `albedo` I'd say the results are fine.
A few details to note though

## ERA5

* new ERA5 is missing the 'pressure' variable - not a problem as I don't think we ever used it
* The `height` values are a bit off
* `albedo` is different. Even substituting `inf` values with `nan` and rerunning the statistics, the results vary significantly. Is the `albedo` relevant for us?

<details>
<summary>Statistics</summary>

```
==== pressure ====
Variable missing from new ERA5 cutout.
==== temperature ====
       	 old		 | 	 new
mean:  	 283.08		 | 	 283.09
min:   	 237.60		 | 	 237.56
max:   	 321.31		 | 	 321.23
std:   	 9.00		 | 	 8.97
count: 	 2e+08		 | 	 2e+08
==== soil temperature ====
       	 old		 | 	 new
mean:  	 283.98		 | 	 283.99
min:   	 254.84		 | 	 253.71
max:   	 305.12		 | 	 304.77
std:   	 6.66		 | 	 6.64
count: 	 2e+08		 | 	 2e+08
==== influx_toa ====
       	 old		 | 	 new
mean:  	 272.11		 | 	 272.11
min:   	 0.00		 | 	 0.00
max:   	 1298.18	 | 	 1298.17
std:   	 360.46		 | 	 360.44
count: 	 2e+08		 | 	 2e+08
==== influx_direct ====
       	 old		 | 	 new
mean:  	 87.83		 | 	 87.81
min:   	 0.00		 | 	 0.00
max:   	 967.38		 | 	 964.25
std:   	 168.26		 | 	 168.01
count: 	 2e+08		 | 	 2e+08
==== runoff ====
       	 old		 | 	 new
mean:  	 0.00		 | 	 0.00
min:   	 -0.00		 | 	 0.00
max:   	 0.01		 | 	 0.01
std:   	 0.00		 | 	 0.00
count: 	 2e+08		 | 	 2e+08
==== roughness ====
       	 old		 | 	 new
mean:  	 0.39		 | 	 0.39
min:   	 0.00		 | 	 0.00
max:   	 1.87		 | 	 1.84
std:   	 0.51		 | 	 0.50
count: 	 2e+08		 | 	 2e+08
==== height ====
       	 old		 | 	 new
mean:  	 187.85		 | 	 186.93
min:   	 -68.65		 | 	 -53.30
max:   	 2462.57	 | 	 2416.21
std:   	 315.52		 | 	 310.80
count: 	 2e+08		 | 	 2e+04
==== albedo ====
       	 old		 | 	 new
mean:  	 nan		 | 	 0.08
min:   	 -inf		 | 	 -22.95
max:   	 inf		 | 	 1.00
std:   	 inf		 | 	 0.11
count: 	 2e+08		 | 	 2e+08
==== influx_diffuse ====
       	 old		 | 	 new
mean:  	 51.63		 | 	 51.63
min:   	 0.00		 | 	 0.00
max:   	 599.57		 | 	 599.58
std:   	 71.31		 | 	 71.21
count: 	 2e+08		 | 	 2e+08
==== wnd100m ====
       	 old		 | 	 new
mean:  	 6.92		 | 	 6.91
min:   	 0.00		 | 	 0.00
max:   	 37.80		 | 	 37.61
std:   	 4.23		 | 	 4.22
count: 	 2e+08		 | 	 2e+08

```

</details>

## SARAH

* `influx_diffuse` and `influx_direct` values are a bit off (artifact from interpolation or did the SARAH underlying data undergo changes?). I don't consider the differences to be substantial.
* `albedo` see above's comment for `era5` .

<details>
  <summary>Statistics</summary>

```
==== influx_direct ====
       	 old		 | 	 new
mean:  	 85.91		 | 	 79.31
min:   	 0.00		 | 	 0.00
max:   	 1017.88	 | 	 998.55
std:   	 176.00		 | 	 170.38
count: 	 4e+08		 | 	 4e+08
==== influx_diffuse ====
       	 old		 | 	 new
mean:  	 63.69		 | 	 58.83
min:   	 0.00		 | 	 0.00
max:   	 499.25		 | 	 498.91
std:   	 88.53		 | 	 87.03
count: 	 4e+08		 | 	 4e+08
==== temperature ====
       	 old		 | 	 new
mean:  	 284.25		 | 	 284.14
min:   	 236.93		 | 	 234.06
max:   	 321.31		 | 	 321.85
std:   	 9.21		 | 	 9.40
count: 	 4e+08		 | 	 4e+08
==== influx_toa ====
       	 old		 | 	 new
mean:  	 287.17		 | 	 287.43
min:   	 -0.25		 | 	 0.00
max:   	 1283.75	 | 	 1298.17
std:   	 374.74		 | 	 374.21
count: 	 4e+08		 | 	 4e+08
==== albedo ====
       	 old		 | 	 new
mean:  	 nan		 | 	 0.08
min:   	 -inf		 | 	 -18.76
max:   	 inf		 | 	 1.00
std:   	 inf		 | 	 0.12
count: 	 4e+08		 | 	 4e+08
```

</details>

FabianHofmann:
thanks @euronion

1. Albedo (ERA5): thats a bit weird, have you had a look at the timeseries? 
2. Influx (SARAH): I have a look at this, with a dx = dy = 0.2 we don't apply the regridding (at least should not. In case we still do, the regridding is differnt from the old, as the old had an error)
3. Bounds: perhaps we set it then to 35.2 in the config

euronion:
Regarding 1.:
* Is the `albedo` feature required?

I checked the time-series. Some - not all - night values in the old dataset were `nan` values, by month:
image

Nevertheless looking at the values per month e.g. March without any `nan`, the values are still different which results in a large aggregation error when aggregating a whole month:
image

The differences in albedo vary across the months
image

The largest differences occur on the edge of sunlight sweeping across the cutout. But other cutout cells not close to the edge of sunrise / dawn also seem affected).

Checking single time slices the values look mostly ok:
image

Any ideas?

Regarding 3.:
I don't think it matters. 34.8 should be safely outside the country boundaries. If we change it, let's make the boundaries for ERA5 and SARAH identical.

config.default.yaml
```diff
@@ -62,18 +62,28 @@ electricity:
 atlite:
   nprocesses: 4
   cutouts:
+    # use 'base' to determine geographical bounds and time span from config
+    # base: 
+      # module: era5
```
euronion, line 67: This seems unnecessary and only be used by the `mock_snakemake(...)` in `build_cutout(...)`. Is this a remnant from dev/testing?

doc/configuration.rst
```diff
@@ -21,7 +21,7 @@ Top-level configuration
    :lines: 5-12,20,27-34
 
 .. csv-table::
-   :header-rows: 1
+   :header-rows: 1
```
euronion, line 24:
```suggestion
   :header-rows: 1
```

doc/configuration.rst
```diff
@@ -95,9 +95,12 @@ Specifies the temporal range to build an energy system model for as arguments to
 ``atlite``
 =============
 
+Define and specify the ``atlite.Cutout`` used for calculating renewable potentials and time-series. All options except for ``features`` are directly used as `cutout parameters <https://atlite.readthedocs.io/en/latest/ref_api.html#cutout>`_.
+
 .. literalinclude:: ../config.default.yaml
    :language: yaml
-   :lines: 62-75
+   :start-at: atlite:
+   :end-before: renewable:
```
euronion, line 103: That's so much better, great!

doc/release_notes.rst
```diff
@@ -11,6 +11,8 @@ Release Notes
 Upcoming Release
 ================
 
+* Switch to new ``atlite`` version ``v0.2.1``. The version upgrade comes along with significant speed up for the rule ``build_renewable_profiles.py`` (~factor 2). A lot of the code which calculated the landuse availability is now outsourced and does not rely on ``glaes``, ``geokit`` anymore. This facilitates the environment building and version compatibility of ``gdal``, ``libgdal`` with other packages.     
+
```
euronion, line 15:
```suggestion
Also rule and script `build_country_flh` are removed as they're no longer used or maintained.
```

doc/release_notes.rst
```diff
@@ -11,6 +11,8 @@ Release Notes
 Upcoming Release
 ================
 
+* Switch to new ``atlite`` version ``v0.2.1``. The version upgrade comes along with significant speed up for the rule ``build_renewable_profiles.py`` (~factor 2). A lot of the code which calculated the landuse availability is now outsourced and does not rely on ``glaes``, ``geokit`` anymore. This facilitates the environment building and version compatibility of ``gdal``, ``libgdal`` with other packages.
```
euronion, line 14: Switch to new major release, `>=v0.2.1`.

envs/environment.docs.yaml
```diff
@@ -9,7 +9,7 @@ dependencies:
   - python<=3.7
   - pip
   - pypsa>=0.17.1
-  - atlite=0.0.3
+  - atlite=0.2.1
```
euronion, line 12: Intentionally fixed or should this also be `atlite>=0.2.1` ?

Are we missing `mamba` in the doc env?

envs/environment.yaml
```diff
@@ -8,12 +8,12 @@ channels:
   - bioconda
   - http://conda.anaconda.org/gurobi
 dependencies:
-  - python
+  - python>=3.8
```
euronion, line 11: Is this necessary? `python=3.7` has its EOL in 2023. If it is necessary this also has to be in the release notes.

scripts/build_cutout.py
```diff
@@ -92,28 +88,38 @@
...
 import logging
+import atlite
+import geopandas as gpd
+import pandas as pd
 from _helpers import configure_logging
 
-import os
-import atlite
+
+logger = logging.getLogger(__name__)
```
euronion, line 97:
```suggestion
```

Duplicate line.

scripts/build_cutout.py
```diff
@@ -92,28 +88,38 @@
...
 
 logger = logging.getLogger(__name__)
 
 if __name__ == "__main__":
     if 'snakemake' not in globals():
         from _helpers import mock_snakemake
-        snakemake = mock_snakemake('build_cutout', cutout='europe-2013-era5')
+        snakemake = mock_snakemake('build_cutout', cutout='base')
```
euronion, line 104: I see `base` is used here again. Any particular reason?
With `base` being commented out in `config.default.yaml` this would not work.
The keeping the `mock_snakemake` on one of the productive cutouts like `europe-2013-era5` seems more reasonable to me.

scripts/build_cutout.py
```diff
@@ -92,28 +88,38 @@
...
     cutout_params = snakemake.config['atlite']['cutouts'][snakemake.wildcards.cutout]
-    for p in ('xs', 'ys', 'years', 'months'):
-        if p in cutout_params:
-            cutout_params[p] = slice(*cutout_params[p])
 
-    cutout = atlite.Cutout(snakemake.wildcards.cutout,
-                        cutout_dir=os.path.dirname(snakemake.output[0]),
-                        **cutout_params)
+    snapshots = pd.date_range(freq='h', **snakemake.config['snapshots'])
+    time = [snapshots[0], snapshots[-1]]
```
euronion, line 110: I am again assuming the `time` if it is not specified in `config.yaml` for the cutout.
The script should rather fail than trying to fix it.

scripts/build_cutout.py
euronion: As with `time` I'd personally favour the script to fail if the bounds are not giving explicitly.
I think we either go we specifying extent and time explicitly or implicitly, but I am against doing it manually (primary) and automatically (fallback).

Rather what we could *always* to, is check whether the bounding boxes are within the cutout extent and - if not - fail.

scripts/build_hydro_profile.py
```diff
@@ -76,20 +75,21 @@
...
+    eia_stats = vhydro.get_eia_annual_hydro_generation(
+        snakemake.input.eia_hydro_generation).reindex(columns=countries)
     inflow = cutout.runoff(shapes=country_shapes,
                            smooth=True,
                            lower_threshold_quantile=True,
                            normalize_using_yearly=eia_stats)
 
     if 'clip_min_inflow' in config:
-        inflow.values[inflow.values < config['clip_min_inflow']] = 0.
+        inflow = inflow.where(inflow > config['clip_min_inflow'], 0)
```
euronion, line 93:
```suggestion
        inflow = inflow.where(inflow >= config['clip_min_inflow'], clip_min_inflow)
```

scripts/build_renewable_profiles.py
```diff
@@ -60,7 +60,6 @@
     **Source:** `GEBCO <https://www.gebco.net/data_and_products/images/gebco_2019_grid_image.jpg>`_
 
 - ``resources/natura.tiff``: confer :ref:`natura`
-- ``resources/country_shapes.geojson``: confer :ref:`shapes`
```
euronion, line 63: This is still present in `Snakefile` , right?

scripts/build_renewable_profiles.py
```diff
@@ -180,220 +179,158 @@
...
 
-    clc = gk.raster.loadRaster(paths["corine"])
-    clc.SetProjection(gk.srs.loadSRS(3035).ExportToWkt())
+    if correction_factor != 1.:
+        logger.info(f'correction_factor is set as {correction_factor}')
 
-    natura = gk.raster.loadRaster(paths["natura"])
+    if not config.get('keep_all_available_areas', True):
+        logger.warning("Argument `keep_all_available_areas` is ignored. "
+                       "Continue with keeping all areas.")
```
euronion, line 220: Is this `keep_all_available_areas` a new flag? Shouldn't we - if it is set to `False` -
rather `raise NotImplementedError("Argument 'keep_all_available_areas=False' is not implemented.")`

scripts/build_renewable_profiles.py
```diff
@@ -180,220 +179,158 @@
...
-    if "max_depth" in config:
-        gebco = gk.raster.loadRaster(paths["gebco"])
-        gebco.SetProjection(gk.srs.loadSRS(4326).ExportToWkt())
+    if isinstance(config.get("corine", {}), list):
+        config['corine'] = {'grid_codes': config['corine']}
 
-    clc = gk.raster.loadRaster(paths["corine"])
-    clc.SetProjection(gk.srs.loadSRS(3035).ExportToWkt())
+    if correction_factor != 1.:
+        logger.info(f'correction_factor is set as {correction_factor}')
```
euronion, line 216: Thank you! I remember this `correction_factor` bugging my when I started with PyPSA-EUR. :)

scripts/build_renewable_profiles.py
```diff
@@ -180,220 +179,158 @@
...
 
     # select only buses with some capacity and minimal capacity factor
     ds = ds.sel(bus=((ds['profile'].mean('time') > config.get('min_p_max_pu', 0.)) &
-                     (ds['p_nom_max'] > config.get('min_p_nom_max', 0.))))
+                      (ds['p_nom_max'] > config.get('min_p_nom_max', 0.))))
 
     if 'clip_p_max_pu' in config:
-        ds['profile'].values[ds['profile'].values < config['clip_p_max_pu']] = 0.
+        min_p_max_pu = config['clip_p_max_pu']
+        ds['profile'] = ds['profile'].where(ds['profile'] >= min_p_max_pu, 0)
```
euronion, line 334:
```suggestion
        ds['profile'] = ds['profile'].where(ds['profile'] >= min_p_max_pu, min_p_max_pu)
```

Why not use `ds.clip(...)` here?

doc/configuration.rst
fneum: Yes! Should have read the documentation...

scripts/build_renewable_profiles.py
fneum: Yes, and also used

envs/environment.docs.yaml
FabianHofmann: Correct, atlite>=0.2.1 is better.
Regarding mamba, we install it beforehand (at least in the CI) and install the env via mamba.

config.default.yaml
euronion: Two thoughts in opposite directions.

1. Yes, it is convinient, but it is not used. That's the differenc between the base network and base cutout. Is it based on `era5` or on `sarah`? None of the renewables are defined for it.

2. With the new `atlite` version we can combine `cutouts` from different datasets unless the variable overlap, right? So nothing is really stopping us from having a true `base` cutout with `era5` and `sarah` data in it? We could merge the two separate cutouts and then switch all renewables profiles to the base cutout. The need for separation is gone.

envs/environment.docs.yaml
euronion, line 12:
```suggestion
  - atlite>=0.2.1
```

euronion: Ok.

scripts/build_renewable_profiles.py
FabianHofmann: It's an old flag, which is not available anymore. Raise an error would be a rough break, a warning that it's ignored from now on is more modest

scripts/build_renewable_profiles.py
FabianHofmann: again, it's set to zero :)

config.default.yaml
FabianHofmann: Good points
1. Yes, for now I would leave it just there, so we can compare the results properly the 'master' version
2. Exactly (if variables overlap the first mentioned module is prioritized). The only advantage in separating them is the keeping the higher resolution in the sarah cutout (0.2 x 0.2). But I think it's good to go with an overall 0.25 x 0.25 resolution

envs/environment.yaml
FabianHofmann: is done

doc/release_notes.rst
FabianHofmann: done

scripts/build_hydro_profile.py
euronion: Gotcha.

scripts/build_renewable_profiles.py
euronion: Hm, ok. I guess in general we should think about how to treat deprecated options (warn about them centrally, when to remove them, ...). If warnings for deprecated config options build up it will get messy over time.

config.default.yaml
euronion: How about we take this idea and deal with it in a separate issue? Let's focus on getting the new atlite version in.
(Sorry, I tend to get side tracked as well).

scripts/build_cutout.py
euronion: So this does not work with the SARAH dataset then? (IMHO a strong point against the code as it is then)

I guess making it automatic sounds nice, but is it worth the trouble?
Cutouts / cutout config shouldn't have to be adjusted or changed too often, I guess really only once
if you initialise the model or change the geographical scope.

I'm not against fancy solutions and automatisims, just afraid of models doing stuff which becomes hard to debug and people no longer having to give some basic thought to the underlying reanalysis data.

FabianHofmann: No, it also works with the Sarah data. I was reffering to your suggestion 

*Rather what we could always do, is check whether the bounding boxes are within the cutout extent and - if not - fail.*

What I meant is that checking that the bounds of the cutout are withing the bounds of the network does not necessarily play out. The sarah data is only defined until 65 degree north, which is why we set the upper bound to this value. So, there are valid cases where you select an sub-area within your network area which still leads to the same result. 

Nevertheless, the automated bound setting would set the whole covered area. 
I imagine a use case where e.g. you want to make a detailed study on the French grid for the years 2013/2014/2015. All you would have to do is define a new cutout with the module and the rest is done by the workflow. Setting bounding boxes correct is sometimes very annoying and in fact you'd have to do exactly the same what is done in the code.

config.default.yaml
FabianHofmann: I totally agree. (no worry, it's good to discuss things)

scripts/build_renewable_profiles.py
FabianHofmann: You are right, perhaps it's a better option to mention it in the release notes and not "during" the workflow?

scripts/build_cutout.py
euronion: This again leaves me with open questions:
1. Why would you recreate (in your example) the cutout, when you already have our default cutout?
2. If you automatically determine the bounds and they are outside 65°N, what does `atlite` do for the grid cells not covered by SARAH?
3. Why is a `.buffer(0.5)` sufficient? Should it be lower or higher and how is it related to the grid resolution.

scripts/build_renewable_profiles.py
euronion: Yes! If the option is removed it has to be noted in the release notes.
If we want to notify the user of unsupported options, then we would think about a small scriptlet checks the `config.yaml` for consistency and deprecated options.

If you want to keep the warning, then I suggest you expand it:
```
if 'keep_all_available_areas' in config:
    # Warning
```

scripts/build_cutout.py
FabianHofmann: 1. Because you consider different years (not covered by  the default cutout) and you wouldn't want to create a cutout for whole Europe when you just need a single country. 
2. They are effectively set to zero. This happens in the conversion to pv with a `.fillna(0)`
3. good point, it should probably be the max of dx and dy

scripts/build_renewable_profiles.py
FabianHofmann: Agree!

scripts/build_cutout.py
euronion: 1. Makes sense.
2. Ok, fair enough.

I'm still not convinced we need this feature and not a big fan of it. But from a code-quality I think it is ok from my POV.
(One or two comments in the code would be a nice ;)

scripts/build_renewable_profiles.py
euronion: Great. Cross-referencing the #158 issue which is open for a confic check script.

FabianHofmann:
Short update on the albedo issue and sarah data:

## Albedo (ERA5)
As @euronion said, single snapshots are very close, but in general the old data has some weird outliers. Here the mean over time (used the 0.3 x 0.3 grid, clipped the north, and rounded some coordinates of the old dataset): 
albedo_mean

and space
albedo_spatialmean

The mean of the new data also fits much better to [official values from other sources](https://www.dwd.de/DE/leistungen/rcccm/int/rcccm_int_salltr.html).

So I think we are good to go here. 

## Sarah data
@euronion thanks again for double-checking. 
The data is a bit "off" because the regridding was fixed in the new atlite version, see https://github.com/PyPSA/atlite/pull/125. But looking deeper into it brought me to a tedious malfunction of the `regrid` function which is now fixed. It did not affect the cutout I sent you. But another tedious thing, affected the bounds of sarah cutouts (old cutout and the one I sent you). So I will send you another sarah cutout, which we then can upload to zenodo.

euronion:
That looks and sounds great. Nice job!

fneum:
Anything left to do?

FabianHofmann:
I will just double check the non-tutorial workflow and I think then we're fine :)

euronion:
I'll briefly check the tutorial workflow.

euronion:
Tutorial:

* Atlite part runs without problems direct problems.
* The progressbar is glitchy, presumably due to multithreading (the % numbers appear and disappear in unordered manner). Not an issue for this PR, wort a issue for `atlite` @FabianHofmann ?
* Using `snakemake --configfile config.tutorial.yaml -jall` without having a `config.yaml` in the folder fails as it tries to create the electricity network from `config.defaul.yaml` and downloads the full `europe-2013-era5`. Cause is this convinience code line in the `Snakefile` which in turn triggers the `build_hydro` rule:
```
if not exists("config.yaml"):
    copyfile("config.default.yaml", "config.yaml")
```

But if one follows the documentation and does `mv config.tutorial.yaml config.yaml` everything works out.

FabianHofmann:
yep, the jittering progressbar appears when having parallel processes that result in parallel progressbars which all want to be displayed at the same spot. 
What we could actually do is to set atlite.show_progress in the config to false. That leaves out the progressbars of the compute_availability process and is actually almost twice as fast (this has to do with the synchronization of callbacks of parallel processes). Downside you don't have a direct feedback. Is that an option?

euronion:
I prefer jiggly feedback on whether the process is alive or not over no feedback. (In favour of the current situation)

fneum:
I see the positive aspects of both :) no preference.

Re config: I see, so the option `--configfile` only overwrites the main config? I think it's fine to leave as is for now.

euronion:
> Re config: I see, so the option `--configfile` only overwrites fields of the main config but does not substitute it fully? I think it's fine to leave as is for now.

Yes. That's intentional behaviour: Use `configfile: <...>` from `Snakefile` as a basis and overwrite (update) with `--configfile <..>`. We could think about removing the lazy copy-mechanism from the `Snakefile` and prefer to have the workflow fail (if no explicit creation of `config.yaml` happens) or encourage the use of `--configfile <...>` . But that's a separate issue. Worth raising?

FabianHofmann:
Short update. Everything works fine, but (!) the memory consumption is not stable for solar profiles when executing with the latest dask version (2021.4.1). I try to find the latest version that works smoothly and then add a requirement to the envs. 

Unfortunately, I have no clue where the memory issue comes from (my guess it has to do with xarray clip functions). But i'm pretty sure this will be solved in the future.

fneum:
https://github.com/dask/dask/issues/7583?

FabianHofmann:
Yes, exactly! That costed me some hours...

fneum:
restarting tests, got stuck on build_cutout csdapi request.

FabianHofmann:
right, which reminds me that we can/should set the test config to `retrieve_cutout: true`

FabianHofmann:
Alright, I think we are there. Thumbs up if I should merge.

fneum:
:partying_face: :balloon: :tada:

FabianHofmann:
Yes, finally :)

euronion:
Indeed. Finally! (= 😄

MichelJD:
Hi together,
I've just encountered an error, while running pypsa-eur-sec with the new atlite-version. In the build_population_layouts.py script, there is still the old style of loading an atlite cutout. I think this needs to be changed. 
Best regards 
Michél

fneum:
Hi, yes, PyPSA-Eur-Sec is not yet updated for the new atlite version. We are working on this. Please use PyPSA-Eur release v0.3.0 or a commit pre #224 in combination with PyPSA-Eur-Sec for now.

PR #232, opened 2021-03-23
Adding marginal cost for storage

pz-max:
Closes # (if applicable).

The current config.yaml contains marginal cost inputs. Though, only the input for solar and wind had an effect on the optimization (tested). This PR fixes the bug, allowing to add marginal cost for energy storage components in the config.yaml file. For instance, changing the marginal cost of H2 to 0.01 would add to the electrolyser and fuel cell marginal cost of 0.01 EUR/MWh.

## Checklist

- [ ] Newly introduced dependencies are added to `envs/environment.yaml` and `envs/environment.docs.yaml`.

### DIFF
--- a/scripts/add_extra_components.py
+++ b/scripts/add_extra_components.py
@@ -155,7 +155,7 @@ def attach_stores(n, costs):
                efficiency=costs.at['battery inverter','efficiency'],
                capital_cost=costs.at['battery inverter', 'capital_cost'],
                p_nom_extendable=True,
-               marginal_cost=costs.at["battery inverter", "marginal_cost"]))
+               marginal_cost=costs.at["battery inverter", "marginal_cost"])
 
 
 def attach_hydrogen_pipelines(n, costs):
--- a/scripts/add_extra_components.py
+++ b/scripts/add_extra_components.py
@@ -114,7 +114,8 @@ def attach_stores(n, costs):
                carrier='H2 electrolysis',
                p_nom_extendable=True,
                efficiency=costs.at["electrolysis", "efficiency"],
-               capital_cost=costs.at["electrolysis", "capital_cost"])
+               capital_cost=costs.at["electrolysis", "capital_cost"],
+               marginal_cost=snakemake.config['costs']['marginal_cost'].get('H2'))
 
         n.madd("Link", h2_buses_i + " Fuel Cell",
                bus0=h2_buses_i,
@@ -123,7 +124,8 @@ def attach_stores(n, costs):
                p_nom_extendable=True,
                efficiency=costs.at["fuel cell", "efficiency"],
                #NB: fixed cost is per MWel
-               capital_cost=costs.at["fuel cell", "capital_cost"] * costs.at["fuel cell", "efficiency"])
+               capital_cost=costs.at["fuel cell", "capital_cost"] * costs.at["fuel cell", "efficiency"],
+               marginal_cost=snakemake.config['costs']['marginal_cost'].get('H2'))
 
     if 'battery' in carriers:
         b_buses_i = n.madd("Bus", buses_i + " battery", carrier="battery", **bus_sub_dict)
@@ -141,7 +143,8 @@ def attach_stores(n, costs):
                carrier='battery charger',
                efficiency=costs.at['battery inverter', 'efficiency'],
                capital_cost=costs.at['battery inverter', 'capital_cost'],
-               p_nom_extendable=True)
+               p_nom_extendable=True,
+               marginal_cost=snakemake.config['costs']['marginal_cost'].get('battery')))
 
         n.madd("Link", b_buses_i + " discharger",
                bus0=b_buses_i,
@@ -149,7 +152,8 @@ def attach_stores(n, costs):
                carrier='battery discharger',
                efficiency=costs.at['battery inverter','efficiency'],
                capital_cost=costs.at['battery inverter', 'capital_cost'],
-               p_nom_extendable=True)
+               p_nom_extendable=True,
+               marginal_cost=snakemake.config['costs']['marginal_cost'].get('battery')))
 
 
 def attach_hydrogen_pipelines(n, costs):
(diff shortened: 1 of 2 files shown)


scripts/add_extra_components.py
```diff
@@ -114,7 +114,8 @@ def attach_stores(n, costs):
                carrier='H2 electrolysis',
                p_nom_extendable=True,
                efficiency=costs.at["electrolysis", "efficiency"],
-               capital_cost=costs.at["electrolysis", "capital_cost"])
+               capital_cost=costs.at["electrolysis", "capital_cost"],
+               marginal_cost=snakemake.config['costs']['marginal_cost'].get('H2'))
```
fneum, line 118:
```suggestion
               marginal_cost=costs.at["electrolysis", "marginal_cost"])
```

fneum (changes requested):
Thanks for your contribution @pz-max! There are a few changes I would suggest:

- `marginal_cost` and `capital_cost`(if specified in `config.yaml`) are overwritten in the `load_costs` function which builds the `costs` dataframe, e.g. if you put `costs: marginal_cost: electrolysis: 1.234` this will be included in `costs`. https://github.com/PyPSA/pypsa-eur/blob/9048af482fdd6cfbe7202efa5561fe6f078c7734/scripts/add_electricity.py#L182-L186 So I would access the marginal cost analogously to the capital cost and not via `snakemake.config...`

- Another point: simply putting 'H2' for both fuel cell and electrolysis is not sufficiently specific. Use and access:
  ```yaml
  costs:
    marginal_cost:
      fuel cell: 1.234
      electrolysis: 5.678
  ```

scripts/add_extra_components.py
```diff
@@ -141,15 +143,17 @@ def attach_stores(n, costs):
                carrier='battery charger',
                efficiency=costs.at['battery inverter', 'efficiency'],
                capital_cost=costs.at['battery inverter', 'capital_cost'],
-               p_nom_extendable=True)
+               p_nom_extendable=True,
+               marginal_cost=snakemake.config['costs']['marginal_cost'].get('battery')))
```
fneum, line 147: syntax error: one `)` too much

scripts/add_extra_components.py
```diff
@@ -141,15 +143,17 @@ def attach_stores(n, costs):
...
 
         n.madd("Link", b_buses_i + " discharger",
                bus0=b_buses_i,
                bus1=buses_i,
                carrier='battery discharger',
                efficiency=costs.at['battery inverter','efficiency'],
                capital_cost=costs.at['battery inverter', 'capital_cost'],
-               p_nom_extendable=True)
+               p_nom_extendable=True,
+               marginal_cost=snakemake.config['costs']['marginal_cost'].get('battery')))
```
fneum, line 156: syntax error, see above

scripts/add_extra_components.py
pz-max: Yes it should be )) instead )))

pz-max:
@fneum It seems to work now. A € sign in the config.yaml "#€/MWh" took me quite some nerves today! 

I implemented also wildcards to play around with the MC's. Though I guess it's a too-rare case to be useful for a PR?
I will also report on the MC recommendations to avoid Unintended Storage Cycling as well as a function on how to measure USC if this is of any use.

pz-max (commented):
FYI. Everything works

scripts/add_extra_components.py
```diff
@@ -106,15 +106,17 @@ def attach_stores(n, costs):
                carrier='H2',
                e_nom_extendable=True,
                e_cyclic=True,
-               capital_cost=costs.at["hydrogen storage", "capital_cost"])
+               capital_cost=costs.at["hydrogen storage", "capital_cost"],
+               marginal_cost=costs.at["H2", "marginal_cost"])
```
fneum, line 110: Shouldn't this be "hydrogen storage" rather than "H2"? To my understanding "H2" is for the `StorageUnit` and would equal the marginal costs for the fuel cell (dispatch only). Also, the marginal cost of a `Store` component refers to the dispatch. What would be the case for a marginal cost on the store rather than the links (electrolysis / fuel cell)?

fneum (changes requested):
One more comment :)

scripts/add_extra_components.py
pz-max: Great points. 
- Yes, we can make it consistent and name it "hydrogen storage" rather than "H2". 
- Maybe also worth changing the carrier name from "H2" to "H2 storage" as the carrier for the fuel cell are named "H2 fuel cell"/"H2 electrolyser"?
- Question. Are the marginal cost at `Store` not penalising the storage level? That's how I imagined its function or what you mean its related to the dispatch?
- I am not sure yet where this makes to add marginal cost to sense to add marginal cost to `Store` but giving the opportunity to do so might be useful. 
- I am writing up an unintended storage cycling (USC) study & how marginal cost can avoid that. In short, you can add marginal cost to either electrolyser, storage, discharger or your close to zero marginal cost generators (renewables) to avoid USC. Solver accuracy plays also a big role.

fneum: > Yes, we can make it consistent and name it "hydrogen storage" rather than "H2".

Great

> Maybe also worth changing the carrier name from "H2" to "H2 storage" as the carrier for the fuel cell are named "H2 fuel cell"/"H2 electrolyser"?

Would prefer to keep it as H2, because the `Store` stores the carrier, whereas electrolysers and fuel cell convert. But I will keep it in mind also considering increased integration with PyPSA-Eur-Sec.

> Question. Are the marginal cost at Store not penalising the storage level? That's how I imagined its function or what you mean its related to the dispatch?

No, it penalises when the store is emptied (dispatched). See e.g. https://github.com/PyPSA/PyPSA/blob/d2c1a0805ccf23acfbabf4cb3bf3a931115b3718/pypsa/opf.py#L1183-L1185 for the pyomo version.

> I am not sure yet where this makes to add marginal cost to sense to add marginal cost to Store but giving the opportunity to do so might be useful.

In this case, I think it'd be better to leave it out then. This functionality is easy to add once needed.

> I am writing up an unintended storage cycling (USC) study & how marginal cost can avoid that. In short, you can add marginal cost to either electrolyser, storage, discharger or your close to zero marginal cost generators (renewables) to avoid USC. Solver accuracy plays also a big role.

Yes, that's what we have been doing. Quite an easy fix.

fneum:
Grazie!

PR #245, opened 2021-05-21
remove six dependency

fneum:


martacki (approved):
didn't find any more dependencies than those  :)

Issue #152, opened 2020-04-07
simplify script aggregates Corsica to Italy


martacki:
full resolution includes one bus for corsica (bus "2306"), that, when simplifying, gets aggregated to Italy "Festland" (bus "2343") which we do not want to happen.

The bug is in the script "simplify_network.py", ie. happens in the step when going from elec.nc to elec_s.nc. The function where the bug happens is "simplify_links", where chains of links (HVDC lines) are aggregated to one representative link (HVDC line).

fneum:
So the solution would be to not aggregate links that cross countries? E.g.

```
IT ---link--- FR ---link--- IT
```

eb5194:
I was thinking how to fix it as well, and that would fix it only by accident. We wouldn't want Corsica to be aggregated to France "Festland" as well if the link was from Corsica to France... right? Is that because it's an island, or an own synchronous zone? Or both?

BTW, I think that there are more than this case were HVDC links are aggregated across country borders. simplify_links doesn't take countries & synchronous zones into account at all. Maybe that has to be added?

euronion:
Is it an option to specifically state the cases, where the simplification should not happen, e.g. an exception list?
While it might not be the most elegant solution, it is usually easier to understand and guaranteed to not accidentally apply to buses where it should not apply to.

eb5194:
it would be definitely super easy to exclude this particular aggregation by removing the corresponding entry from the simplify_links_map. but doesn't fix all the (potential other cases where we aggregate across country/synch-zone bounderies. I could check how many cases we're talking about here?...

nworbmot:
I don't know if this helps, but I noticed looking at the bus region shapes that Corsica is always aggregated to mainland France (even for 256 clusters), whereas Sardinia always remains its own synchronous area and a separate region. Perhaps something is going wrong with the country shapes?

fneum:
@martacki do you have an overview if this is still the case or the issue can be closed?

p-glaum:
@martacki is this issue solved?

martacki:
@p-glaum no... but the work-around is rather ugly, and would add +1 node to the lowest resolution making it 38 nodes. we should discuss if we really want that

fneum:
Closed via #1215

PR #248, opened 2021-05-28
Cleanup simplify

martacki:
Closes # (if applicable).
#152 solved in simplify_links when respecting country borders 
#219  solved by using stubs_clustering and get_clustering_from_busmap

cleanup the original simplify_network script.

1. `simplify_links`
a lot less code, but as far as my understanding goes, does the same thing except adapting the costs as was done before in aggregate_and_move_components. But, from my point of view, pypsa.networkclustering also adapts the costs to the new parameters of lines&links, and the clustering is more transparent this way. The costs issue arises again in 2.

2. `remove_stubs`
use the original stubs_clustering from pypsa.networkclustering, instead of aggregating and moving components manually in this script. Don't change costs as done in aggregate_and_move_components, but adapt them according to pypsa.networkclustering.

_I don't see why generators/storage unit/... costs were manipulated earlier. I suggest to keep them the same, i.e. total capital costs and marginal costs of simplified network are the same as the ones of the electricity network before simplification. They are also the same when moving from elec_s.nc to elec_s_nclusters.nc. So I don't understand why they should be adapted according to links costs, connection costs, or else. If this was by intention, then my suggested changes here shouldn't be merged, and we can close this (draft) PR._

**caution**: For 2., a minor change in pypsa.networkclustering is necessary:
both the changes affect`stubs_clustering`:
1. `def stubs_clustering(network,use_reduced_coordinates=True, line_length_factor=1.0, bus_strategies=dict(country=lambda x: x.value_counts().index[0])):`
and
2. `return get_clustering_from_busmap(network, busmap, line_length_factor=line_length_factor, bus_strategies=bus_strategies)`.

Therefore, checks should be failing. I will open a PR in `pypsa` when we decide to merge.

ToDo:
- [ ] Check if implementation works on a selection of countries
- [ ] Improve code style (help appreciated :) )
- [ ] How to treat n.buses.carrier entries that don't agree in the busmap (idea: use busstrategies)
- [ ] Bug: Generators are not yet aggregated

## Checklist


martacki:
Costs should be aggregated, but better documentation should be added.

PR #252, opened 2021-06-21
GitHub actions CI

fneum:
Building times comparison (minutes)

|         | Travis | GH Actions |
|---------|--------|------------|
| ubuntu  | 7      | 7          |
| macos   | 12     | 8          |
| windows | 21     | 16         |


test/config.test1.yaml
```diff
@@ -20,7 +20,7 @@ countries: ['DE']
 
 snapshots:
   start: "2013-03-01"
-  end: "2014-04-01"
+  end: "2013-03-08"
```
FabianHofmann, line 23: that's intended? (guess it makes sense, since we just want to test the code)

.github/workflows/ci.yaml
```diff
@@ -0,0 +1,47 @@
...
+      uses: conda-incubator/setup-miniconda@v2.1.1 
+      with: # checks out environment 'test' by default
+          mamba-version: "*"
+          channels: conda-forge,defaults
+          channel-priority: true
+
+    - name: Install dependencies  
+      run: |
+        echo -ne "url: ${CDSAPI_URL}\nkey: ${CDSAPI_TOKEN}\n" > ~/.cdsapirc
+        echo -e "  - glpk\n  - ipopt<3.13.3" >> envs/environment.yaml
```
FabianHofmann, line 39: perhaps in the long term we should add these to the default env?

FabianHofmann (approved):
Looks great!

.github/workflows/ci.yaml
fneum: Maybe, so far we have kept the solvers out of the dependencies.

test/config.test1.yaml
fneum: Yes, shaves about 20 seconds off the runtime.

PR #259, opened 2021-06-29
Adapt to new snapshot weightings

fneum:
Closes # (if applicable).

Use `n.snapshot_weightings.generators` or `n.snapshot_weightings.objective`  as substitute for `n.snapshot_weightings`.

@lisazeyen is the use of generators and objective semantically correct?

## Checklist


lisazeyen (commented):
it is a bit tricky with the implementation of the new snapshot weightings. In general, the different snapshot weightings in PyPSA-Eur will probably always be the same and therefore the proposed changes do not matter. For example, if you aggregate a network temporally to 3h resolution, `n.snapshot_weightings.generators=n.snasphots_weightings.stores=n.snapshot_weightings.objective=3` .

The different weightings were introduced mainly to allow you to model typical days with hourly resolution to represent an entire year.
For example, 7 typical days with hourly resolution, then
`n.snapshot_weightings.generators=n.snapshot_weightings.objective=52` , `n.snasphots_weightings.stores=1` .
 Through the different weighting one would not overestimate e.g. standing losses. The nodal balance constraint should still work because the bus injections are not weighted within the constraint.

fneum:
Thanks for the feedback. Yes, it's just a first so it at least works with the current PyPSA master even though the differentiated functionality is not implemented yet.

Issue #251, opened 2021-06-15
p_nom vs. p_nom_min for extendable generators


martacki:
## Checklist

- [x] I am using the current [`master`](https://github.com/PyPSA/pypsa-eur/tree/master) branch or the latest [release](https://github.com/PyPSA/pypsa-eur/releases). Please indicate.
- [x] I am running on an up-to-date [`pypsa-eur` environment](https://github.com/PyPSA/pypsa-eur/blob/master/envs/environment.yaml). Update via `conda env update -f envs/environment.yaml`.

## Describe the Bug

If existing (renewable) powerplants (p_nom>0) are added to the network with the option to be extendable (p_nom_extendable=True), then the lower bound for the optimisation (p_nom_min) is not updated, i.e. the optimisation solves a "greenfield problem" where the outcome might be that the optimal capacity is lower than the already installed one (p_nom_opt < p_nom).

I can imagine 2 solutions to fix this:
a) in add_electricity, **always** set all p_nom_min values to p_nom, for all the generators where p_nom_extendable=True
b) the same as in a) but with an **additional configuration**, such as "brownfield=True" in the config.yaml. Should be true by default; but if False, then the p_nom_min values are not updated (p_nom_min = 0). In this case one could compare if the installed capacities are (close to) the optimum. This could be configured somewhere in electricity:, e.g. after the configuration if installed RE plants should be included.

What do you think?

martacki:
solved with #260

PR #260, opened 2021-06-30
fix p_nom_min values for extendable carriers with positive p_nom

martacki:
Closes # (if applicable).
#251 

I decided to go with solution a) mentioned in #251 to keep the configuration file clean. It can be easily extended to solution b), if desired.

Not sure if the update of p_nom_max is really necessary, but there are definitely cases where it's otherwise lower than p_nom_min. Would it be better to account for this when distributing OPSD capacities? At least in the heuristic approach? (I'd assume that the precise mapping of asset->node is more accurate than the p_nom_max estimate, but is not in case of estimate_capacity_stats). What do you think?

## Checklist

### DIFF
--- a/scripts/_helpers.py
+++ b/scripts/_helpers.py
@@ -125,6 +125,15 @@ def load_network_for_plots(fn, tech_costs, config, combine_hydro_ps=True):
 
     return n
 
+def update_p_nom_max(n):
+    # if extendable carriers (solar/onwind/...) have capacity >= 0,
+    # e.g. existing assets from the OPSD project are included to the network,
+    # the installed capacity might exceed the expansion limit.
+    # Hence, we update the assumptions.
+    
+    n.generators.p_nom_max = (n.generators
+                              .apply(lambda b: b[['p_nom_min','p_nom_max']].max(), axis=1))
+
 def aggregate_p_nom(n):
     return pd.concat([
         n.generators.groupby("carrier").p_nom_opt.sum(),
--- a/scripts/_helpers.py
+++ b/scripts/_helpers.py
@@ -131,8 +131,7 @@ def update_p_nom_max(n):
     # the installed capacity might exceed the expansion limit.
     # Hence, we update the assumptions.
     
-    n.generators.p_nom_max = (n.generators
-                              .apply(lambda b: b[['p_nom_min','p_nom_max']].max(), axis=1))
+    n.generators.p_nom_max = n.generators[['p_nom_min', 'p_nom_max']].max(1)
 
 def aggregate_p_nom(n):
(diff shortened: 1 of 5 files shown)


scripts/_helpers.py
```diff
@@ -125,6 +125,15 @@ def load_network_for_plots(fn, tech_costs, config, combine_hydro_ps=True):
...
     return n
 
+def update_p_nom_max(n):
+    # if extendable carriers (solar/onwind/...) have capacity >= 0,
+    # e.g. existing assets from the OPSD project are included to the network,
+    # the installed capacity might exceed the expansion limit.
+    # Hence, we update the assumptions.
+    
+    n.generators.p_nom_max = (n.generators
+                              .apply(lambda b: b[['p_nom_min','p_nom_max']].max(), axis=1))
```
FabianHofmann, line 135:
```suggestion
    n.generators.p_nom_max = n.generators[['p_nom_min', 'p_nom_max']].max(1)
```

scripts/cluster_network.py
```diff
@@ -379,6 +379,8 @@ def consense(x):
                                                extended_link_costs=hvac_overhead_cost,
                                                focus_weights=focus_weights)
 
+    update_p_nom_max(n)
```
FabianHofmann, line 382: I am just wondering, is this one still necessary? Because we already ensured that p_nom_max >= p_nom_min right?

scripts/simplify_network.py
```diff
@@ -364,6 +364,8 @@ def consense(x):
         n, cluster_map = cluster(n, int(snakemake.wildcards.simpl))
         busmaps.append(cluster_map)
 
+    update_p_nom_max(n)
```
FabianHofmann, line 367: Same here

FabianHofmann (commented):
Sounds reasonable :)

scripts/cluster_network.py
martacki: I was wondering that, too. But the clustering (if potential: conservative) aggregates the p_nom_max values using the minimum value, whilst the p_nom_min ones are now summed up. So there might be cases where this might be necessary.

However, by default, thepotential is 'simple', hence the p_nom_max values are summed.

fneum (approved):
LGTM

martacki:
Weird with those checks...

@FabianHofmann fine by you? :)

FabianHofmann:
yes :)

FabianHofmann (approved):

fneum:
HTTP connection time out error. Can occur sometimes. In this case, you can click on the failed test and in top-right corner click "re-run all jobs"

PR #261, opened 2021-07-07
exporting additional costs, and compatibility adjustments

huckebrink:
- changed boolean parameters for `pd.read_csv` to lists, as it caused errors on windows machines
- added an export for "addedCosts.csv" in `simplify_network._adjust_capital_costs_using_connection_costs` to allow for model comparability.

## Checklist


scripts/simplify_network.py
```diff
@@ -178,13 +178,21 @@ def _compute_connection_costs_to_bus(n, busmap, connection_costs_per_link=None,
...
+            cost_strings = []
+            for b, d in costs.iteritems():
+                cost_strings.append("{:.0f} Eur/MW/a for `{}`".format(d, b))
+                addedCosts.append((b, d))
             logger.info("Displacing {} generator(s) and adding connection costs to capital_costs: {} "
-                        .format(tech, ", ".join("{:.0f} Eur/MW/a for `{}`".format(d, b) for b, d in costs.iteritems())))
+                        .format(tech, ", ".join(cost_strings)))
+
+    addedCosts_df = pd.DataFrame(data=addedCosts, columns=["buses", "costs"])
+    addedCosts_df.to_csv("resources/addedCosts.csv", index=False)
```
fneum, line 195: How about adding a column to `n.generators` rather than creating a new file?

```py
n.generators.loc[costs.index, "added_connection_capital_cost"] = costs
```

fneum (changes requested):
Great catch regarding the boolean values. Just to confirm: This breaks with the new pandas 1.3.0?

Regarding the connection cost, one could record this information simply in another column of `n.generators`. What do you think?

scripts/simplify_network.py
huckebrink: I'm not quite sure with which version the boolean values parameter changed. However, it did break during one of our recent runs and is specified as a list in the [docs for pandas.read_csv](https://pandas.pydata.org/docs/reference/api/pandas.read_csv.html#pandas.read_csv).

Great suggestion! That one liner is surely more elegant.

scripts/simplify_network.py
```diff
@@ -185,6 +185,7 @@ def _adjust_capital_costs_using_connection_costs(n, connection_costs_to_bus):
             n.generators.loc[costs.index, "capital_cost"] += costs
             logger.info("Displacing {} generator(s) and adding connection costs to capital_costs: {} "
                         .format(tech, ", ".join("{:.0f} Eur/MW/a for `{}`".format(d, b) for b, d in costs.iteritems())))
+    n.generators.loc[costs.index, "added_connection_capital_cost"] = costs
```
fneum, line 188: Almost: I think it needs to go inside the for loop to be logged for all technologies. Otherwise, it'll only log the last.
```suggestion
            n.generators.loc[costs.index, "added_connection_capital_cost"] = costs
```

fneum: Took the liberty to commit this suggestion directly, as I would like to get the other fix in quickly. Hope you don't mind.

fneum (approved):

scripts/simplify_network.py
huckebrink: This results in somewhat unexpected behaviour, as all entries in `n.generators.loc[costs.index, "added_connection_capital_cost"]` in `elec_s.nc` (or `elec_s_{clusters}.nc`) are `NaN`. 

From what I understand, operations are performed on the `elec.nc` network in `simplify_network.py`, which is later aggregated to the `elec_s.nc`. Adding a `print` statement in the code yields the expected result, but I fail to find the data in any of the output files.

Might the information be lost somewhere during aggregation?

Please correct me, if I'm mistaken anywhere.

fneum: Ah, you might be right.

pz-max:
> Great catch regarding the boolean values. Just to confirm: This breaks with the new pandas 1.3.0?
> 
> Regarding the connection cost, one could record this information simply in another column of `n.generators` as a one-liner. What do you think?

Hi @fneum, 
On a Linux pc, fresh pypsa-eur instillation, and with pandas=1.2.5:
                         `true_values='t', false_values='f',` --> breaks
                         `true_values=['t'], false_values=['f'],` --> works fine
Means that the current upstream version doesn't work until this is merged. :+1: 
Thanks @huckebrink for solving that.

scripts/simplify_network.py
fneum: Ok, now done in a separate file as you initially suggested. Sorry for taking a while to respond to this.

fneum:
Right, thanks @pz-max for reminding me. There was one open issue regarding the export of the connection costs which I had postponed for a while. Should merge automatically once tests pass.

PR #275, opened 2021-09-14
remove snakemake dependencies in functions

martacki:

remove snakemake "dependencies" from functions, such that they can be easily imported and used in other python scripts/jupyter notebooks, etc. Currently many function imports result in the following error:
`NameError: name 'snakemake' is not defined`

Work in progress, some scripts still missing.

Works locally.


scripts/add_electricity.py
```diff
@@ -422,8 +413,7 @@ def attach_hydro(n, costs, ppl):
                inflow=inflow_t.loc[:, hydro.index])
 
 
-def attach_extendable_generators(n, costs, ppl):
-    elec_opts = snakemake.config['electricity']
+def attach_extendable_generators(n, costs, ppl, elec_opts = {'extendable_carriers': {'Generator': []}}):
```
fneum, line 416: Might be better to have `elec_opts` as compulsory argument.

scripts/build_shapes.py
```diff
@@ -107,11 +107,10 @@ def _simplify_polys(polys, minarea=0.1, tolerance=0.01, filterremote=True):
     return polys.simplify(tolerance=tolerance)
 
 
-def countries():
-    cntries = snakemake.config['countries']
+def countries(naturalearth, cntries=[]):
```
fneum, line 110: cntries compulsory?

scripts/cluster_network.py
```diff
@@ -268,12 +265,10 @@ def clustering_for_n_clusters(n, n_clusters, custom_busmap=False, aggregate_carr
     else:
         raise AttributeError(f"potential_mode should be one of 'simple' or 'conservative' but is '{potential_mode}'")
 
-    if custom_busmap:
-        busmap = pd.read_csv(snakemake.input.custom_busmap, index_col=0, squeeze=True)
-        busmap.index = busmap.index.astype(str)
-        logger.info(f"Imported custom busmap from {snakemake.input.custom_busmap}")
-    else:
+    if custom_busmap is False:
```
fneum, line 268:
```suggestion
    if not custom_busmap:
```

scripts/cluster_network.py
```diff
@@ -268,12 +265,10 @@ def clustering_for_n_clusters(n, n_clusters, custom_busmap=False, aggregate_carr
     else:
         raise AttributeError(f"potential_mode should be one of 'simple' or 'conservative' but is '{potential_mode}'")
 
-    if custom_busmap:
-        busmap = pd.read_csv(snakemake.input.custom_busmap, index_col=0, squeeze=True)
-        busmap.index = busmap.index.astype(str)
-        logger.info(f"Imported custom busmap from {snakemake.input.custom_busmap}")
```
fneum, line 274: Another way than moving this elsewhere, would be to pass a file through the `custom_busmap` kwarg.

e.g. `busmap = pd.read_csv(custom_busmap, index_col=0, squeeze=True)`

scripts/prepare_network.py
```diff
@@ -70,21 +70,14 @@
 logger = logging.getLogger(__name__)
 
 
-def add_co2limit(n, Nyears=1., factor=None):
-
-    if factor is not None:
-        annual_emissions = factor*snakemake.config['electricity']['co2base']
-    else:
-        annual_emissions = snakemake.config['electricity']['co2limit']
+def add_co2limit(n, co2limit=1.487e+9, Nyears=1.):
```
fneum, line 73: I would remove the default for `co2limit`. As it assumes some spatial scope of the model.

scripts/prepare_network.py
```diff
@@ -151,7 +140,7 @@ def average_every_nhours(n, offset):
     return m
 
 
-def apply_time_segmentation(n, segments):
+def apply_time_segmentation(n, segments, solver_name="cplex"):
```
fneum, line 143: Shouldn't we use an open-source solver as default, like cbc?

scripts/prepare_network.py
```diff
@@ -237,16 +225,17 @@ def set_line_nom_max(n):
     for o in opts:
         m = re.match(r'^\d+seg$', o, re.IGNORECASE)
         if m is not None:
-            n = apply_time_segmentation(n, m.group(0)[:-3])
+            n = apply_time_segmentation(n, m.group(0)[:-3], solver_name=snakemake.config["solving"]["solver"]["name"])
```
fneum, line 228:
```suggestion
            solver_name = snakemake.config["solving"]["solver"]["name"]
            n = apply_time_segmentation(n, m.group(0)[:-3], solver_name=solver_name)
```

needs line break

scripts/prepare_network.py
```diff
@@ -237,16 +225,17 @@ def set_line_nom_max(n):
...
     for o in opts:
         if "Co2L" in o:
             m = re.findall("[0-9]*\.?[0-9]+$", o)
             if len(m) > 0:
-                add_co2limit(n, Nyears, float(m[0]))
+                co2limit=float(m[0])*snakemake.config['electricity']['co2base']
+                add_co2limit(n, Nyears, co2limit)
             else:
-                add_co2limit(n, Nyears)
+                add_co2limit(n, Nyears, snakemake.config['electricity']['co2limit'])
```
fneum, line 238: The order of arguments is incorrect, here.

scripts/prepare_network.py
```diff
@@ -237,16 +225,17 @@ def set_line_nom_max(n):
...
+            n = apply_time_segmentation(n, m.group(0)[:-3], solver_name=snakemake.config["solving"]["solver"]["name"])
             break
 
     for o in opts:
         if "Co2L" in o:
             m = re.findall("[0-9]*\.?[0-9]+$", o)
             if len(m) > 0:
-                add_co2limit(n, Nyears, float(m[0]))
+                co2limit=float(m[0])*snakemake.config['electricity']['co2base']
+                add_co2limit(n, Nyears, co2limit)
```
fneum, line 236: The order of arguments is incorrect (due to changes to the add_co2limit function).

scripts/prepare_network.py
```diff
@@ -237,16 +225,17 @@ def set_line_nom_max(n):
...
-            n = apply_time_segmentation(n, m.group(0)[:-3])
+            n = apply_time_segmentation(n, m.group(0)[:-3], solver_name=snakemake.config["solving"]["solver"]["name"])
             break
 
     for o in opts:
         if "Co2L" in o:
             m = re.findall("[0-9]*\.?[0-9]+$", o)
             if len(m) > 0:
-                add_co2limit(n, Nyears, float(m[0]))
+                co2limit=float(m[0])*snakemake.config['electricity']['co2base']
```
fneum, line 235:
```suggestion
                co2limit = float(m[0]) * snakemake.config['electricity']['co2base']
```

fneum (changes requested):
This is a very useful change and how it should be!

I have a few style comments:

```py
function(arg1 = 3, arg5 = 'sdfa')
```

It would be nice to change them to a common style:

```py
function(
    arg1=4,
    arg2='asdfas',
)
```

i.e. line breaks between arguments and no space between `=` in the function calls (only there)

I think we should steer all functions to as few as possible defaults.

Sometimes you changed the order. Could you double-check whether the order in the function calls is correct?

scripts/add_electricity.py
```diff
@@ -563,22 +548,32 @@ def add_nice_carrier_names(n, config=None):
     n = pypsa.Network(snakemake.input.base_network)
     Nyears = n.snapshot_weightings.objective.sum() / 8760.
 
-    costs = load_costs(Nyears)
-    ppl = load_powerplants()
+    costs = load_costs(tech_costs = snakemake.input.tech_costs, config = snakemake.config['costs'],
+                       elec_config = snakemake.config['electricity'], Nyears = Nyears)
```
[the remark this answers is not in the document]
FabianHofmann: Let's keep it for the pre-commit code formatting

FabianHofmann:
@martacki would you take over the next steps? 

* resolving the remarks made by @fneum 
* use arguments in function calls instead of keyword argument style, for example 
   ```
   load_costs(tech_costs = snakemake.input.tech_costs,
                                         config = snakemake.config['costs'],
                                         elec_config=snakemake.config['electricity'], Nyears = Nyears)
   ```
   should look like 
   ```
   load_costs(snakemake.input.tech_costs, snakemake.config['costs'], snakemake.config['electricity'], Nyears = Nyears)
   ```
* Maybe change `snakemake.input` into `paths`, as it was done in some scripts before. This makes the function call more slim. Same can be done with `snakemake.config`. So the example would look like
   ```
   load_costs(paths.tech_costs, config['costs'], config['electricity'], Nyears=Nyears)
   ```

scripts/prepare_network.py
martacki: Agree, but I think that shouldn't be done in this PR because we're improving style not content here

scripts/cluster_network.py
martacki: isn't that exactly how it is now? if not, sorry, then I don't get your point.

scripts/prepare_network.py
martacki: yup, already changed in master apparently.

martacki:
> This is a very useful change and how it should be!
> 
> I have a few style comments:
> 
> ```python
> function(arg1 = 3, arg5 = 'sdfa')
> ```
> 
> It would be nice to change them to a common style:
> 
> ```python
> function(
>     arg1=4,
>     arg2='asdfas',
> )
> ```
> 
> i.e. line breaks between arguments and no space between `=` in the function calls (only there)
> 
> I think we should steer all functions to as few as possible defaults.
> 
> Sometimes you changed the order. Could you double-check whether the order in the function calls is correct?

do you want this done in this PR? I think that could be done elsewhere because it's not related to snakemake dependencies...

martacki:
> 

happens when everyone is stirring in the same pot. no worries, thanks, i was getting a headache from all the github-failed-CI-run e-mails :D

martacki:
>     * Maybe change `snakemake.input` into `paths`, as it was done in some scripts before. This makes the function call more slim. Same can be done with `snakemake.config`. So the example would look like
>       ```
>       load_costs(paths.tech_costs, config['costs'], config['electricity'], Nyears=Nyears)
>       ```

I made a helpers function `retrieve_snakemake_keys` for that and call it in the beginning of every script so that the style is consistent. Otherwise we do it in some scripts but then allow exceptions in others because there's only one or few snakemake call(s). But that's just confusing to users. Hope that's fine for you!

martacki:
Just ran all scripts on the cluster, they all seem to terminate just fine with one exception: `build_cutout` doesn't work, but there's an error not related to the suggested changes here (`Exception: Missing/incomplete configuration file: /root/.cdsapirc`).

For all the snakemake-related stuff: maybe someone could double-check the changes? @FabianHofmann @fneum
I'm sure I've overseen some kw-arg style function calls, but all snakemake related dependencies should be removed now.

scripts/simplify_network.py
```diff
@@ -390,22 +386,29 @@ def consense(x):
...
+
+    n = pypsa.Network(paths.network)
 
     n, trafo_map = simplify_network_to_380(n)
 
-    n, simplify_links_map = simplify_links(n)
+    Nyears = n.snapshot_weightings.objective.sum() / 8760
+    technology_costs = load_costs(tech_costs = paths.tech_costs,
+                                  config = config['costs'],
+                                  elec_config = config['electricity'], Nyears = Nyears)
```
fneum, line 398: I think here you missed to remove the keyword arguments and add them as positional arguments?

scripts/prepare_network.py
fneum: But you are adding this hard coded number here. This would just as well function without a default because it is never called as `add_co2limit(n)`:

```py
def add_co2limit(n, co2limit, Nyears=1.)
```

fneum (changes requested):
@martacki thanks for the revisions! I think it's very close to being ready to merge. Just found two small issues.

But I am not so convinced of the `retrieve_snakemake_keys()` function. Somehow the returned variables lose the information that they come from snakemake and are given different names (e.g. input -> path, ouput -> out). I understand the motivation to make the code more compact. How about this alternative:

```py
sm = snakemake
sm.input
sm.output
sm.config
```

euronion:
> @martacki thanks for the revisions! I think it's very close to being ready to merge. Just found two small issues.
> 
> But I am not so convinced of the `retrieve_snakemake_keys()` function. Somehow the returned variables lose the information that they come from snakemake and are given different names (e.g. input -> path, ouput -> out). I understand the motivation to make the code more compact. How about this alternative:
> 
> ```python
> sm = snakemake
> sm.input
> sm.output
> sm.config
> ```

I second this.

Also sometime needed:
```
sm.params
sm.wildcards
sm.log
```

martacki:
No strong preference from my side, but personally I'd just leave the full `snakemake.XX` name, saves an unnecessary variable and is clearer what files are being used. But up to you, I can also adapt the `sn=snameke` for every script -- -

You fine with changing the names back, @FabianHofmann ?

FabianHofmann:
Sounds good! I'll do that then

martacki:
Changed the snakemake keys back, now using no abbreviations or any other kind of extraction for the snakemake arguments. If we decide to use such or similar abbreviation/tool/feature for this in the future, let's propose it in another PR as discussed in todays coffee break!

fneum:
Ok, let's do this!

Issue #153, opened 2020-04-28
Consider wake losses


fneum:

euronion:
Referencing [GlobalEnergyGIS](https://github.com/niclasmattsson/GlobalEnergyGIS):
Wake losses for wind farms are considered with 11.5% of wind farm output, see the [supplementary materials.](https://github.com/niclasmattsson/GlobalEnergyGIS/blob/master/paper/Mattsson%20et%20al.%202019%20-%20Supplementary%20-%20An%20autopilot%20for%20energy%20models.pdf)

nworbmot:
I suggest we follow the approach of the 2018 paper [Temporally explicit and spatially resolved global offshore wind energy potentials](https://doi.org/10.1016/j.energy.2018.08.153) which has a well documented methodology for both the installable density of offshore wind (see also Issue https://github.com/PyPSA/pypsa-eur/issues/210) and the wake losses.
They arrive at 0.97 MW/km^2 density when taking account of spacing between turbines within arrays, and spacing between arrays, and a 88.55% efficiency factor for wake losses (I'm guessing GLobalEnergyGIS took their 11.5% figure from here).
Calculations assume:
- 5 MW turbines with rotor diameter of 126m (smaller than current installations)
- Arrays of 10x10 = 100 turbines with 10 x diameter spacing between each turbine, i.e. density of 3.15 MW/km^2 = 5 MW/(10*0.126 km)^2 within the array
- Spacing of 10 km between the arrays, leading to average density of 0.97 MW/km^1 = 500 MW / (10*10*0.126 km + 10 km)^2

This is the spacing can be applied *before* all the exclusions, since marine reserves can provide the space? Maybe? What about depth exclusions? Cleanest way would be to place each array properly...

euronion:
[windpowerlib](https://github.com/wind-python/windpowerlib) has two slightly more refined approach, see [their description here](https://windpowerlib.readthedocs.io/en/stable/model_description.html#wake-losses):

They either 
* reduce the wind speed `v`  in the wind farm
* reduce the efficiency of the power curve by a constant factor `c` (as suggested above) or a wind-speed depend factor `c(v)`.

In either case, the modifications are heuristics based on measured data from a few specific wind farms.

I'd consider modifying the wind speed "gold standard" if different wind turbines with different characteristic power curves (for different wind classes) are used. Until then I think the simple approach of reducing the power density serves (as suggested above) serves its purpose well.

fneum:
[...]

PR #278, opened 2021-09-23
add correction factor to offshore wind for wake losses

fneum:
temporary fix for #153


fneum:
as discussed with @nworbmot

PR #289, opened 2021-11-10
Separate load data retrieval from processing with new rule `retrieve_load_data`

fneum:


Snakefile
```diff
@@ -67,7 +67,14 @@ if config['enable'].get('retrieve_databundle', True):
         script: 'scripts/retrieve_databundle.py'
 
 
+rule retrieve_load_data:
+    input: HTTP.remote("data.open-power-system-data.org/time_series/2019-06-05/time_series_60min_singleindex.csv", keep_local=True, static=True)
```
FabianHofmann, line 71:
```suggestion
    input: HTTP.remote(config["load"]["url"], keep_local=True, static=True)
```
Shouldn't we leave the url adress in the config and not hard code it here?

FabianHofmann (commented):
Looks sensible, just one remark

Snakefile
euronion: We discussed it locally and found it to be more comprehensible, if the input data (i.e. also the URL) is located in the Snakefile, as all the other data which we retrieve from the internetz.

FabianHofmann: I agree, all data paths (even remote ones) are in the Snakefile, model setting in the config. Sounds good

PR #291, opened 2021-11-14
update environment and address deprecations

fneum:
depending on whether all CI tests pass, I still need to remove the commented lifted version bounds.

### DIFF
--- a/envs/environment.yaml
+++ b/envs/environment.yaml
@@ -21,7 +21,7 @@ dependencies:
   - openpyxl
   - pycountry
   - seaborn
-  - snakemake-minimal<=6.8.0 # encoding issue maybe solved by https://github.com/snakemake/snakemake/pull/1146
+  - snakemake-minimal
   - memory_profiler
   - yaml
   - pytables
--- a/envs/environment.yaml
+++ b/envs/environment.yaml
@@ -2,17 +2,17 @@
 #
 # SPDX-License-Identifier: MIT
 
-name: pypsa-eur-nov14
+name: pypsa-eur
 channels:
   - conda-forge
   - bioconda
   - http://conda.anaconda.org/gurobi
 dependencies:
-  - python=3.8 # fix because gurobi has no builds for python 3.9
+  - python>=3.8
   - pip
   - mamba # esp for windows build
 
-  - pypsa>=0.18
+  - pypsa>=0.18.1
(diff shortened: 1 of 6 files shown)


envs/environment.yaml
```diff
@@ -2,19 +2,19 @@
 #
 # SPDX-License-Identifier: MIT
 
-name: pypsa-eur
+name: pypsa-eur-nov14
```
FabianHofmann, line 5: Not sure about this here, or is it not intended? If we want to do this, I'd rather use the pypsa-eur version as a suffix.

envs/environment.yaml
```diff
@@ -2,19 +2,19 @@
...
 channels:
   - conda-forge
   - bioconda
   - http://conda.anaconda.org/gurobi
 dependencies:
-  - python>=3.8
+  - python=3.8 # fix because gurobi has no builds for python 3.9
```
FabianHofmann, line 11: Don't think this is necessary. I have python 3.9 and gurobi works smoothly (however cplex is not available on python 3.9)

scripts/cluster_network.py
```diff
@@ -140,6 +140,9 @@
 from pypsa.networkclustering import (busmap_by_kmeans, busmap_by_spectral_clustering,
                                      _make_consense, get_clustering_from_busmap)
 
+import warnings
+warnings.filterwarnings(action='ignore', category=UserWarning)
```
FabianHofmann, line 144: Out of curiosity - what warning is suppressed?

FabianHofmann (changes requested):

envs/environment.yaml
fneum: Ups. Let's keep pypsa-eur.

envs/environment.yaml
fneum: Interesting. It works. But why if the builds for py39 are not in the conda channel? https://anaconda.org/Gurobi/gurobi/files

scripts/cluster_network.py
fneum: KMeans user warning that dataset contains features that are not used in the clustering.

fneum: Maybe @martacki can comment on this deprecation warning and whether it's something we need to address?

FabianHofmann (approved):

PR #293, opened 2021-11-17
introduce hierarchical agglomeratice clustering (hac)

martacki:

...story continues...

Introduce hierarchical agglomeraticve clustering method into pypsa-eur, now that it is (finally) included in PyPSA [PyPSA/PyPSA#289](https://github.com/PyPSA/PyPSA/pull/289).

**(additional) changes**:
1. simplify now by default aggregates buses that are no substations.
2. clustering now defaults to the new method "hac", if not specified for kmeans.
3. clustering method in simplify_network can be different from the method used in cluster_network.

Locally the function(s) worked fine.

## Checklist


scripts/cluster_network.py
```diff
@@ -170,6 +170,45 @@ def weighting_for_country(n, x):
     return (w * (100. / w.max())).clip(lower=1.).astype(int)
 
 
+def get_feature_for_hac(n, buses_i, feature=None): #buses_i = n.buses.index
```
fneum, line 173:
```suggestion
def get_feature_for_hac(n, buses_i=None, feature=None):

    if buses_i is None:
        buses_i = n.buses.index
```

Would it make sense to make `buses_i` an optional argument?

scripts/cluster_network.py
```diff
@@ -221,12 +260,18 @@ def n_bounds(model, *n_id):
...
+def busmap_for_n_clusters(n, n_clusters, solver_name, focus_weights=None, algorithm="kmeans", feature=None, **algorithm_kwds):
     if algorithm == "kmeans":
         algorithm_kwds.setdefault('n_init', 1000)
         algorithm_kwds.setdefault('max_iter', 30000)
         algorithm_kwds.setdefault('tol', 1e-6)
 
+    if algorithm == "hac":
+        feature = get_feature_for_hac(n, buses_i=n.buses.index, feature=feature)
+    elif feature is not None:
+        logger.info(f"keyword argument feature is only valid for algorithm 'hac'."
```
fneum, line 272:
```suggestion
        logger.warning(f"Keyword argument feature is only valid for algorithm 'hac'."
```
This should be a warning.

scripts/cluster_network.py
```diff
@@ -377,6 +426,8 @@ def consense(x):
                                                line_length_factor=line_length_factor,
                                                potential_mode=potential_mode,
                                                solver_name=snakemake.config['solving']['solver']['name'],
+                                               algorithm=snakemake.config.get('clustering', {}).get('cluster_network', {}).get('algorithm', 'kmeans'),
+                                               feature=snakemake.config.get('clustering', {}).get('cluster_network', {}).get('feature', None),
```
fneum, line 430:
```suggestion
                                               algorithm=cluster_config.get('algorithm', 'kmeans'),
                                               feature=cluster_config.get('feature', None),
```

You could move the common expression:

```py
snakemake.config.get('clustering', {}).get('cluster_network', {})
```

further up between `custom_busmap = ` and `clustering = `:

```py
cluster_config = snakemake.config.get('clustering', {}).get('cluster_network', {})
```

scripts/simplify_network.py
```diff
@@ -400,12 +402,14 @@ def consense(x):
...
-    if snakemake.config.get('clustering', {}).get('simplify', {}).get('to_substations', False):
+    if snakemake.config.get('clustering', {}).get('simplify_network', {}).get('to_substations', False):
         n, substation_map = aggregate_to_substations(n)
         busmaps.append(substation_map)
 
     if snakemake.wildcards.simpl:
-        n, cluster_map = cluster(n, int(snakemake.wildcards.simpl))
+        n, cluster_map = cluster(n, int(snakemake.wildcards.simpl),
+                                 algorithm=snakemake.config.get('clustering', {}).get('simplify_network', {}).get('algorithm', 'hac'),
+                                 feature=snakemake.config.get('clustering', {}).get('simplify_network', {}).get('feature', None))
```
fneum, line 412:
```suggestion
                                 algorithm=cluster_config.get('algorithm', 'hac'),
                                 feature=cluster_config.get('feature', None))
```

See comment above:

```py
cluster_config = snakemake.config.get('clustering', {}).get('simplify_network', {})
```

fneum (changes requested):
Looking good!

So HAC will be the new default? Could you highlight this in a release note? It's quite a significant change.

Two maps with clustered regions (~200 nodes) with previous k-means default and new HAC defaults would also very nicely complement this PR. Do you have something like that at hand?

martacki:
thanks @fneum for suggestions and improvements for the code, should all be in.

However, I want to make a change to the network "topology" sort-of and change two country-labels of n.buses. I'll do this in a new PR (somewhat cleaner), but the new PR should be in before this one.
HAC aggregates only connected nodes, but there are two nodes (one in HR, substation "Dakovo" and one in LU, substation "Vianden") that have no connections to other nodes within their own country. This throws a warning for HAC, because we do the clustering within each country seperately. It doesn't break HAC, but if HR/LU get assigned 2 nodes, this particlular one will never be aggregated and this is not the best partition.

I'll add some figs and definitely a note for the release_notes as well!

martacki:
Ok, the following commit is rather ugly...

We can just remove the commit if we decide we don't want that "feature", but accounting for it results in a better representation of the grid topology of the network after clustering.

Here's a map extract for better reference.
LU+HR

martacki:
and, as requested, here are some plots for comparison (k-meas vs. hac-time) for different spatial resolutions:

[hac-time+kmeans_50.pdf](https://github.com/PyPSA/pypsa-eur/files/8005813/hac-time%2Bkmeans_50.pdf)
[hac-time+kmeans_100.pdf](https://github.com/PyPSA/pypsa-eur/files/8005810/hac-time%2Bkmeans_100.pdf)
[hac-time+kmeans_150.pdf](https://github.com/PyPSA/pypsa-eur/files/8005806/hac-time%2Bkmeans_150.pdf)

doc/configtables/clustering.csv
```diff
@@ -1,3 +1,8 @@
 ,Unit,Values,Description
-simplify,,,
--- to_substations,bool,"{'true','false'}","Aggregates all nodes without power injection (positive or negative, i.e. demand or generation) to electrically closest ones"
+simplify_network,,,
+-- to_substations,bool,"One of {'true','false'}","Aggregates all nodes without power injection (positive or negative, i.e. demand or generation) to electrically closest ones"
+-- algorithm,str,"One of {‘kmenas’, ‘hac’}",
```
fneum, line 4:
```suggestion
-- algorithm,str,"One of {‘kmeans’, ‘hac’}",
```

doc/configtables/clustering.csv
```diff
@@ -1,3 +1,8 @@
...
+-- feature,str,"Str in the format ‘carrier1+carrier2+...+carrierN-X’, where CarrierI can be from {‘solar’, ‘onwind’, ‘offwind’, ‘ror’} and X is one of {‘cap’, ‘time’}.",
+cluster_network,,,
+-- algorithm,str,"One of {‘kmenas’, ‘hac’}",
```
fneum, line 7:
```suggestion
-- algorithm,str,"One of {‘kmeans’, ‘hac’}",
```

envs/environment.yaml
```diff
@@ -54,5 +53,6 @@ dependencies:
   - tabula-py
 
   - pip:
+    - git+https://github.com/pypsa/pypsa.git#egg=pypsa
```
fneum, line 56: remove (see above)

scripts/cluster_network.py
```diff
@@ -173,6 +173,50 @@ def weighting_for_country(n, x):
...
+        feature_data = pd.DataFrame(index=buses_i, columns=carriers)
+        for carrier in carriers:
+            try:
+                # without simpl wildcard (bus names are "X X"):
+                feature_data[carrier] = (n.generators_t.p_max_pu.filter(like=carrier).mean()
+                                         .rename(index=lambda x: x.split(' ')[0]))
+            except:
+                # with simpl wildcard (bus names are "X X X"):
+                feature_data[carrier] = (n.generators_t.p_max_pu.filter(like=carrier).mean()
+                                         .rename(index=lambda x: x.split(' ')[0] + ' ' + x.split(' ')[1]))
```
fneum, line 199: I don't particularly like the try-except block. Can you pass the {simpl} wildcard as function argument and convert try-except to if-else?

scripts/cluster_network.py
```diff
@@ -173,6 +173,50 @@ def weighting_for_country(n, x):
...
+        feature_data = pd.DataFrame(columns=buses_i)
+        for carrier in carriers:
+            try:
+                # without simpl wildcard (bus names are "X X"):
+                feature_data = feature_data.append(n.generators_t.p_max_pu.filter(like=carrier)
+                                                   .rename(columns=lambda x: x.split(' ')[0]))[buses_i]
+            except:
+                # with simpl wildcard (bus names are "X X X"):
+                feature_data = feature_data.append(n.generators_t.p_max_pu.filter(like=carrier)
+                                                   .rename(columns=lambda x: x.split(' ')[0] + ' ' + x.split(' ')[1]))[buses_i]
```
fneum, line 211: see above

scripts/cluster_network.py
fneum: I think it can also be shortened. Maybe something like:

```py
if simpl:
    retrieve_func = lambda x: x.split(' ')[0] + ' ' + x.split(' ')[1]
else:
    retrieve_func = lambda x: x.split(' ')[0]

feature_data[carrier] = n.generators_t.p_max_pu.filter(like=carrier).mean().rename(index=retrieve_func)

File: scripts/simplify_network.py [hunk 10]
```diff
@@ -401,12 +402,29 @@ def consense(x):
...
+        busmaps.append(busmap_hac)
+
     if snakemake.wildcards.simpl:
-        n, cluster_map = cluster(n, int(snakemake.wildcards.simpl), snakemake.config)
+        # treatment of outliers (nodes without a profile for a considered carrier) for "simplify"
+        if cluster_config.get("algorithm", "hac") == "hac":
+            carriers = cluster_config.get("feature", "solar+onwind-time").split('-')[0].split('+')
+            buses_i = list(set(n.buses.index)-set(n.generators.query("carrier in @carriers").bus))
+            n, busmap_hac = aggregate_to_substations(n, buses_i)
+            busmaps.append(busmap_hac)
```
fneum, line 423: This code block is very similar to the one above. Could you mend it into a single function that is called at both locations?

scripts/simplify_network.py
```diff
@@ -401,12 +402,29 @@ def consense(x):
 
     busmaps = [trafo_map, simplify_links_map, stub_map]
 
-    if snakemake.config.get('clustering', {}).get('simplify', {}).get('to_substations', False):
+    cluster_config = snakemake.config.get('clustering', {}).get('simplify_network', {})
+    if cluster_config.get('to_substations', False):
         n, substation_map = aggregate_to_substations(n)
         busmaps.append(substation_map)
 
+    # treatment of outliers (nodes without a profile for considered carrier) for "cluster_network"
```
fneum, line 410: maybe add a short comment how they are "treated"?

fneum (changes requested):
Looks good! But I still had a few comments and suggestions for improvements. Could you also add a release note with two or three sentences how to use it and what to keep in mind?

scripts/cluster_network.py
```diff
@@ -221,12 +265,44 @@ def n_bounds(model, *n_id):
...
+                new_country = list(set(n.buses.loc[neighbor_bus].country)-set([country]))[0]
+
+                logger.info(f"overwriting country `{country}` of bus `{disconnected_bus}` to new country `{new_country}`, "
+                            "because it is disconnected from its inital inter-country transmission grid.")
+                n.buses.at[disconnected_bus, "country"] = new_country
+        return n
+
+    if algorithm == "hac":
+        feature = get_feature_for_hac(n, buses_i=n.buses.index, feature=feature)
+        n = fix_country_assignment_for_hac(n)
```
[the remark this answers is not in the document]
fneum: Ah, actually you see that in the graphics you attached earlier. It looks a bit odd. Would definitely default to not changing the assignment.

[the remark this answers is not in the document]
martacki: I'd vote for no configurable option. We already aggregate some nodes in `simplify_network` to other countries that they were initially assigned to (more precicely, in the `remove_stubs` routine which also isn't configurable). That's a much larger change compared to these two buses. As far as I've analysed, currently no other nodes are are affected than the two I have attached plots for.

scripts/cluster_network.py
coroa, line 199: I by off-chance had a small look into the code, out of interest for the hierarchical agglomerative clustering. I don't fully understand the feature_data concept yet, but it looks like you are trying to get a dataframe that has buses as the index and either carriers or time (as hours) on the columns.

Can't you just use the actual information of which bus the generators are attached to from the generators df, ie. something like
```python
        feature_data = {}
        for carrier in carriers:
            gen_i = n.generators.index[n.generators.carrier == carrier]
            feature_data[carrier] = n.generators.p_max_pu[gen_i].mean().set_axis(n.generators.loc[gen_i, "bus"])
        feature_data = pd.DataFrame(feature_data)
```
and
```python
        feature_data = []
        p_max_pu = n.generators_t.p_max_pu.set_axis(n.generators_t.p_max_pu.index.astype(str))
        for carrier in carriers:
            gen_i = n.generators.index[n.generators.carrier == carrier]
            feature_data.append(p_max_pu[gen_i].set_axis(n.generators.loc[gen_i, "bus"], axis="columns"))
        feature_data = pd.concat(feature_data).T
```

coroa: Not sure how this behaves with `windoff-ac/dc`. That might be different from your implementation.

martacki: There are actually a few style things I don't like yet in this pr, and it's lying around for too long. Therefore: Thanks for your suggestion, will try that! Getting a df with the buses as idx and carriers or time (in hours) on the cols is correct.

martacki: > Not sure how this behaves with `windoff-ac/dc`. That might be different from your implementation.

could resolve this with a `df.query.`

scripts/cluster_network.py
```diff
@@ -221,13 +265,45 @@ def n_bounds(model, *n_id):
...
+            m.buses = m.buses.query("country in @country")
+            m.lines = m.lines.query("bus0 in @m.buses.index and bus1 in @m.buses.index")
+            m.links = m.links.query("bus0 in @m.buses.index and bus1 in @m.buses.index")
+
+            _, labels = csgraph.connected_components(m.adjacency_matrix(), directed=False)
+            component = pd.Series(labels, index=m.buses.index)
+            component_sizes = component.value_counts()
+
+            if len(component_sizes)>1:
+                disconnected_bus = component[component==component_sizes[component_sizes==component_sizes.min()].index[0]].index
```
coroa, line 290: Since `value_counts` sorts, the minimum count is automatically the last one

```suggestion
                disconnected_bus = component[component==component_sizes.index[-1]].index
```

scripts/cluster_network.py
```diff
@@ -221,13 +265,45 @@ def n_bounds(model, *n_id):
...
+        from scipy.sparse import csgraph
+
+        # overwrite country of nodes that are disconnected from their country-topology
+        for country in n.buses.country.unique():
+            m = n.copy()
+            m.buses = m.buses.query("country in @country")
+            m.lines = m.lines.query("bus0 in @m.buses.index and bus1 in @m.buses.index")
+            m.links = m.links.query("bus0 in @m.buses.index and bus1 in @m.buses.index")
+
+            _, labels = csgraph.connected_components(m.adjacency_matrix(), directed=False)
```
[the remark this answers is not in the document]
martacki: hm, doesn't do the trick because I need the adjacency of the network with deleted lines & links (lines 282-283)

scripts/cluster_network.py
martacki: resolved with the suggestions of @coroa, because now no distinction is needed anymore.

scripts/cluster_network.py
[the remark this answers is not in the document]
coroa: Ok, i guess the documentation is not accurate. I was reading
```
    busorder : pd.Index subset of network.buses.index
       Basis to use for the matrix representation of the adjacency matrix
       (default: buses.index (network) or buses_i() (sub_network))
```
to mean giving a subset of buses would just naturally work. But in the actual implementation the `get_indexer` call will produce -1 indices in the get_indexer, which will supposedly trip coo_matrix then.

Could of course update the implementation of `adjacency_matrix`:
```python
    bus0_inds = np.concatenate(bus0_inds)
    bus1_inds = np.concatenate(bus1_inds)
    weight_vals = np.concatenate(weight_vals)

    return sp.sparse.coo_matrix((weight_vals, (bus0_inds, bus1_inds)),
                                shape=(no_buses, no_buses))
```

to 

```python
    bus0_inds = np.concatenate(bus0_inds)
    bus1_inds = np.concatenate(bus1_inds)
    weight_vals = np.concatenate(weight_vals)

    disconnected_b = (bus0_inds = -1) | (bus1_inds == -1)
    if disconnected_b.any():
        bus0_inds = bus0_inds[~disconnected_b]
        bus1_inds = bus1_inds[~disconnected_b]
        weight_vals = weight_vals[~disconnected_b]

    return sp.sparse.coo_matrix((weight_vals, (bus0_inds, bus1_inds)),
                                shape=(no_buses, no_buses))
```

Maybe that is too much overhead for this PR though.

Then, I can only recommend the `m = n[n.buses.country.isin(country)]` short-cut to produce the smaller pypsa network.

scripts/simplify_network.py
```diff
@@ -401,12 +402,29 @@ def consense(x):
...
+    if snakemake.config.get("clustering", {}).get("cluster_network", {}).get("algorithm", "hac") == "hac":
```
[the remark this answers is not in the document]
coroa: if it should only happen during the `cluster_network` step, then yes do it from there. If you need to move a function for that, then do it.

fneum (approved):
Happy to merge this.

Although I've been thinking about whether it would be wise to first introduce this new feature and then make it a default in the version after the next. This would just mean changing the default config from `hac` to `kmeans`. What do you think @FabianHofmann @martacki?

@martacki could you also add a few lines and instructions to the release notes?

fneum (approved):

Issue #298, opened 2021-12-10
Missing Lines around Diele


martacki:
There are two missing lines in PyPSA-Eur connecting the substations Diele-Rhede in Germany and Diele-Meeden accross the dutch-german border.

## Checklist
- [x] I am using the current [`master`](https://github.com/PyPSA/pypsa-eur/tree/master) branch or the latest [release](https://github.com/PyPSA/pypsa-eur/releases). Please indicate.
- [x] I am running on an up-to-date [`pypsa-eur` environment](https://github.com/PyPSA/pypsa-eur/blob/master/envs/environment.yaml). Update via `conda env update -f envs/environment.yaml`

This does not result in any kind of programming error, all scripts work. However, it does impact modeling results.

martacki:
resoved with #308

PR #118, opened 2020-01-12
data/entsoegridkit: Update gridkit data files

coroa:
Addresses parts of #55. Closes #74.

## Changes proposed in this Pull Request

Updating the entsoe map extract to 2020. 

The previous map extract is from Dec 2017. Additions and fixes like Malta have been introduced upstream.

https://github.com/PyPSA/GridKit has been updated to retrieve up-to-date map extracts using the single script https://github.com/PyPSA/GridKit/blob/master/entsoe/runall_in_docker.sh. Since the oid identifiers have changed, one still needs to manually adjust
https://github.com/PyPSA/pypsa-eur/blob/master/data/parameter_corrections.yaml and the `replaces` column in https://github.com/PyPSA/pypsa-eur/blob/master/data/links_tyndp.csv .

Maybe the heuristic for splitting lines with annotations into multiple voltage levels is not up to date anymore, either, ie. https://github.com/PyPSA/pypsa-eur/issues/55#issuecomment-519140197 . Check!

## Checklist

- [ ] `parameter_corrections.yaml` has been updated
- [ ] `links_tyndp.csv` has been updated
- [ ] Annotations are properly split into multiple lines
- [ ] I tested my contribution locally and it seems to work fine.
- [ ] A note for the release notes `doc/release_notes.rst` is amended in the format of previous release notes.

PR #308, opened 2022-01-21
Update entsoe gridextract

martacki:
Closes # (if applicable).
#298 

update the entsoe grid extract. not yet finished.

Remaining ToDos:
- [x] link parameters should be checked (in parameter_corrections.yaml); for now: old update p_nom values removed
- [x] link parameters should be checked (in parameter_corrections.yaml); for now: old updated bus1/bus0 values removed around corsica/sardinia
- [x] check the "unconnected buses" that are removed in base_network
- [x] check the buses without country assignment in base_network
- [x] check if rest of the scripts after base_network work (probably most important ;) )
- [x] check what's happening with the converters (why does there appear a converter with the same id three times?)
- [x] lines around bus diele/rhede still missing: why?

## Checklist


martacki:
A few stats (same ones as were considered in #118, but **including** TYNDP) :
|              | old  | new  |
|--------------|------|------|
| buses        | 5323 | 5401 |
| lines        | 6572 | 6693 |
| links        | 68   | 71   |
| transformers | 622  | 628  |

TWkm :
| | old | new |
| - | - | - |
| links | 5.36 | 6.24 |
| lines | 337.24 | 339.90 |
| - 380 kV | 267.99 | 270.55 |
| - 300 kV | 14.75 | 14.82 |
| - 220 | 54.49 | 54.52 |

martacki:
[base_old.pdf](https://github.com/PyPSA/pypsa-eur/files/7972059/base_old.pdf)
[base_new.pdf](https://github.com/PyPSA/pypsa-eur/files/7972064/base_new.pdf)

this new extract was generated because some lines were missing around diele. the new extract now finds and allocates them; in the plot below the red nodes and red lines are new, the underlying lines are from the old base network zoomed into this particular area:
[diele-area.pdf](https://github.com/PyPSA/pypsa-eur/files/7972084/diele-area.pdf)

The problem affected a few lines, not only these two, but those two are probably the most impactful representatives. in total ~ 30 lines. It is fixed in my own fork of the pypsa version of gridkit (https://github.com/martacki/GridKit).

fneum:
ready for review @martacki

martacki:
@fneum I think I solved one bug, but found another (at least its related) ... -.- let me check this first. The minimum bug is that the line lengths of the "new" lines that were missing before is wrong. But I already have an idea on this front...

martacki:
@fneum Yup, lengths of the corrupted lines were indeed affected, but the rest seems okay-ish. I fixed it using the great-circle/haversine distance (see my GridKit fork if needed) for _all_ corrupted lines.

I updated above tables; as only TWkm values have changed, the plots look the same, therefore no update needed there.

Can be reviewed now... hope nothing major comes up anymore.

BuschmannJ:
@martacki Thank you very much for fixing this bug. I appreciate that a lot!

It seems that some LineStrings are wrong located. Link 14748_2 for instance should be located between bus 5784 and 5786 but its linestring is much longer (~50km vs ~2km) and between bus 5784 and bus 5767 (NL). There are 23 lines which have a ratio >3 for linestring-length / length.

martacki:
@BuschmannJ  thanks for pointing this out. The line you mention is one of the corrupted ones, and the Linesting of the line is the same as the one of line 14748. In the old version, both these lines (14748 and 14748_2) were missing.

The reason for this is  or other lines are separated to two or more segments is the way GridKit works. It searches for substations within a certain radius at the location where a given linestring ends. Both the nodes 5784 and 8786 lie within this radius, therefore the line is "corrupted" and an artificial segment is added (here, that's line 14784 connecting 5684 and 5686). Summing up, the line has now two segments and was discarded in the previous version. This radius however is important as not all linestrings are precise and coincide with a substation. I don't know what we can do about it. In this case, this is even not too bad, because otherwise node 5784 would be isolated from the network and would need a manual correction.

To overcome the error and include the corrupted lines in the network, I separated all corrupted lines to their individual segments (2, in this case). The trouble is, that there is only one line-string given for the "total line" (that's why it's the same for all the segments). I corrected the length using the haversine distance. If you check it on the map, the distance between nodes 5784 and 5786 is ~2km. The linestring now should have approx. the length that 14748+14748_2 have in total. All the 21 lines you refer to have such a "twin". That is, the line has some index and the twin has index "_2", "_3" or so. Does that make sense?

fneum:
@martacki do you have a few plots of clustered networks at hand for old and new extract? E.g. 128, 181, 256, 512?

martacki:
Spotted a missing `parameter_correction` prepareing the plots. Here you go:

[clustered_network_to_512.pdf](https://github.com/PyPSA/pypsa-eur/files/8141629/clustered_network_to_512.pdf)
[clustered_network_to_256.pdf](https://github.com/PyPSA/pypsa-eur/files/8141632/clustered_network_to_256.pdf)
[clustered_network_to_181.pdf](https://github.com/PyPSA/pypsa-eur/files/8141634/clustered_network_to_181.pdf)
[clustered_network_to_128.pdf](https://github.com/PyPSA/pypsa-eur/files/8141635/clustered_network_to_128.pdf)
[clustered_network_to_37.pdf](https://github.com/PyPSA/pypsa-eur/files/8141636/clustered_network_to_37.pdf)

fneum (approved):
Looks excellent! I compared the base network and clustered networks and could not find any issues. Major correction is the cross-border capacity between NL and DE. Thanks also @p-glaum for checking the network connectivity. At first, I was concerned about the removed 380 kV connections in Northern Portugal but these also show up in the updated interactive map. So, no problem.

martacki:
Also, special thanks to @coroa for helping out with the entsoe keys! Much appreciated.

PR #313, opened 2022-02-10
Fix random state for kmean clustering

koen-vg:
When the kmeans algorithm is used to cluster networks, this is not deterministic by default. The result is that repeated runs of the `simplify_network` and `cluster_network` rules can and usually do produce different results that vary randomly. This makes results less reproducible when given only a pypsa-eur configuration file.

The fix is to supply a fixed random state to the k-means algorithm.

It might be considered if this should be done PyPSA itself, in the `busmap_by_kmeans` functions in `networkclustering.py`. That's an equally valid option, but maybe a bit opinionated.

I have not checked in a very rigorous way if there are any other sources of randomness in the network building process, but I don't think so.

## Checklist

- [ ] N/A Code and workflow changes are sufficiently documented.
- [ ] N/A Newly introduced dependencies are added to `envs/environment.yaml` and `envs/environment.docs.yaml`.
- [ ] N/A Changes in configuration options are added in all of `config.default.yaml`, `config.tutorial.yaml`, and `test/config.test1.yaml`.
- [ ] N/A Changes in configuration options are also documented in `doc/configtables/*.csv` and line references are adjusted in `doc/configuration.rst` and `doc/tutorial.rst`.

### DIFF
--- a/doc/release_notes.rst
+++ b/doc/release_notes.rst
@@ -24,6 +24,8 @@ Upcoming Release
 
 * Resource definitions for memory usage now follow [Snakemake standard resource definition](https://snakemake.readthedocs.io/en/stable/snakefiles/rules.html#standard-resources) ```mem_mb`` rather than ``mem``.
 
+* Network building is made deterministic by supplying a fixed random state to network clustering routines.
+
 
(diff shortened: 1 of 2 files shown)


fneum:
That's a good idea!

PR #319, opened 2022-03-02
Snakefile: fix memory function for clusters == 'all'

FabianHofmann:

When setting the cluster wildcard to 'all' the memory function raises an error as no number is passed to `int()`. The fix hard-codes the memory limit to the resources needed for a full `pypsa-eur` network with all nodes. 

## Checklist

### DIFF
--- a/Snakefile
+++ b/Snakefile
@@ -313,6 +313,8 @@ def memory(w):
             break
     if w.clusters.endswith('m'):
         return int(factor * (18000 + 180 * int(w.clusters[:-1])))
+    elif w.clusters == "all":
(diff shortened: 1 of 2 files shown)


fneum (approved):
That is a massive memory requirement of 738 GB?

FabianHofmann:
Indeed I was a bit generous there. But this is just an upper limit, not a minimum requirement. It assumes the "worst" case. Otherwise the optimization of the full network would not be supported.

fneum:
Yes, that's fair enough. Wasn't arguing against it.

PR #320, opened 2022-03-03
solve_network: fix load shedding attributes

FabianHofmann:
Fix the carrier attribute name for load shedding generators. 

The `nice_name` and `color` attribute are now hard-coded. Should we rather add them to the config.default.yaml?  I am not sure, because this requires passing extra parts of the config to the `prepare_network` function. Also load shedding is not a real technology.   

## Checklist

- [-] Code and workflow changes are sufficiently documented.
- [-] Newly introduced dependencies are added to `envs/environment.yaml` and `envs/environment.docs.yaml`.
- [-] Changes in configuration options are added in all of `config.default.yaml`, `config.tutorial.yaml`, and `test/config.test1.yaml`.
- [-] Changes in configuration options are also documented in `doc/configtables/*.csv` and line references are adjusted in `doc/configuration.rst` and `doc/tutorial.rst`.
- [-] A note for the release notes `doc/release_notes.rst` is amended in the format of previous release notes.


fneum:
I’d vote for config entries.

martacki:
> I’d vote for config entries.

Nice, that also bothered me for a while now ! but I agree with @fneum. You could import `add_nice_carrier_names` from `add_electricity.py` and execute it right after adding loadshedding to the network, that's at least what I normally do to have the load-colors for plotting etc

FabianHofmann:
> > I’d vote for config entries.
> 
> Nice, that also bothered me for a while now ! but I agree with @fneum. You could import `add_nice_carrier_names` from `add_electricity.py` and execute it right after adding loadshedding to the network, that's at least what I normally do to have the load-colors for plotting etc

good idea!

fneum (approved):
We can start with this.

PR #327, opened 2022-03-10
fix the plot_network snakemake rule

thesethtruth:
Closes # (if applicable).

## Changes proposed in this Pull Request:

1. Removed the dependency of _helper.retrieve_snakemake_keys from plot_network (now uses snakemake.attr)
2. Fixed the reference to the config["plotting"] which was broken
3. Added opts as argument for the config["plotting"] dict tot the plotting functions.

## Checklist

### DIFF
--- a/scripts/plot_network.py
+++ b/scripts/plot_network.py
@@ -20,8 +20,7 @@ Description
 """
 
 import logging
-from _helpers import (retrieve_snakemake_keys, load_network_for_plots,
-                      aggregate_p, aggregate_costs, configure_logging)
+from _helpers import (load_network_for_plots, aggregate_p, aggregate_costs, configure_logging)
 
 import pandas as pd
 import numpy as np
@@ -182,7 +181,7 @@ def plot_map(n, ax=None, attribute='p_nom', opts={}):
     return fig
 
 
-def plot_total_energy_pie(n, ax=None):
+def plot_total_energy_pie(n, ax=None, opts={}):
     if ax is None: ax = plt.gca()
 
     ax.set_title('Energy per technology', fontdict=dict(fontsize="medium"))
@@ -200,7 +199,7 @@ def plot_total_energy_pie(n, ax=None):
             t1.remove()
             t2.remove()
 
-def plot_total_cost_bar(n, ax=None):
+def plot_total_cost_bar(n, ax=None, opts={}):
     if ax is None: ax = plt.gca()
 
     total_load = (n.snapshot_weightings.generators * n.loads_t.p.sum(axis=1)).sum()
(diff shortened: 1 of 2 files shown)


scripts/plot_network.py
```diff
@@ -182,7 +181,7 @@ def plot_map(n, ax=None, attribute='p_nom', opts={}):
     return fig
 
 
-def plot_total_energy_pie(n, ax=None):
+def plot_total_energy_pie(n, ax=None, opts={}):
```
martacki, line 184: if opts is empty, then the function will throw an `KeyError`. Would suggest, therefore, that it's not an optional kwarg.

```suggestion
def plot_total_energy_pie(n, opts, ax=None):
```

scripts/plot_network.py
```diff
@@ -200,7 +199,7 @@ def plot_total_energy_pie(n, ax=None):
             t1.remove()
             t2.remove()
 
-def plot_total_cost_bar(n, ax=None):
+def plot_total_cost_bar(n, ax=None, opts={}):
```
martacki, line 202: same as above.
 ```suggestion
def plot_total_cost_bar(n, opts, ax=None):
```

File: scripts/plot_network.py [hunk 3]
```diff
@@ -259,25 +258,31 @@ def split_costs(n):
...
+    map_figsize = config["plotting"]['map']['figsize']
+    map_boundaries = config["plotting"]['map']['boundaries']
 
     n = load_network_for_plots(paths.network, paths.tech_costs, config)
 
     scenario_opts = wildcards.opts.split('-')
 
     fig, ax = plt.subplots(figsize=map_figsize, subplot_kw={"projection": ccrs.PlateCarree()})
-    plot_map(n, ax, wildcards.attr, config)
+    plot_map(n, ax, wildcards.attr, config["plotting"])
```
martacki, line 277: with the non-optional `opts` `kwarg`, ordering has to be fixed... could you check it?

scripts/plot_network.py
```diff
@@ -259,25 +258,31 @@ def split_costs(n):
...
     set_plot_style()
 
-    paths, config, wildcards, logs, out = retrieve_snakemake_keys(snakemake)
+    paths, config, wildcards, logs, out = (
+        snakemake.input,
+        snakemake.config,
+        snakemake.wildcards,
+        snakemake.log,
+        snakemake.output,
+    )
```
martacki, line 267: Technically, this is correct, but in some prior discussions we've decided to not use these names/abbreviatons anymore. Could you maybe therefore use directly `snakemake.input` instead of `paths` and `snakemake.output` instead of `out`? The logs are not needed, as far as I can see. We could keep the remaining abbreviations 
```suggestion
    config, wildcards = snakemake.config, snakemake.wildcards
```

martacki:
thanks @thesethtruth for taking care! Just some minor comments, what do you think?

scripts/plot_network.py
thesethtruth: > thanks @thesethtruth for taking care! Just some minor comments, what do you think?

Hi Martha! Thanks for your comments. Sorry for the delayed response, I had a holiday break. I will start working on your comments right away.

thesethtruth: Completely agree with you on the non-optionality of the argument. I implemented the same scheme as I saw in the existing code, as not to break any implementations further down the line. Let me go over those arguments and check whether I can indeed change those as you suggest.

thesethtruth:
@martacki done! Thanks for the comments.
I've also rebased to the latest commits from main.
Let me know what you think.

martacki:
looks good, thanks for bug-hunting and contributing, @thesethtruth !

martacki (approved):

fneum:
closes #344

PR #333, opened 2022-03-18
Constrain Snakefile rule build_renewable_profiles with wildcard_constraint

euronion:
* Introduce a `wildcard_constraint` to limit `build_renewable_profiles` to certain technologies.
* Remove potential ambuitiy for `resources/profile_hydro.nc`
* Remove optional inclusion of `build_hydro_profile`: Unnecessary, as rule is only executed, if `add_electricity` wants it.

Side effect:
* Easier to understand `snakemake` errors when trying to run with an unsupported `technology` :

```
# before PR
$ snakemake -call -n resources/profile_hello.nc
Building DAG of jobs...
InputFunctionException in line 185 of .../pypsa-eur/Snakefile:
Error:
  KeyError: 'hello'
Wildcards:
  technology=hello
Traceback:
  File ".../Snakefile", line 191, in <lambda>

# after PR
$ snakemake -call -n resources/profile_hello.nc
Building DAG of jobs...
MissingRuleException:
No rule to produce resources/profile_hello.nc (if you use input functions make sure that they don't raise unexpected exceptions).

```

## Checklist

### DIFF
--- a/Snakefile
+++ b/Snakefile
@@ -201,7 +201,7 @@ rule build_renewable_profiles:
     benchmark: "benchmarks/build_renewable_profiles_{technology}"
     threads: ATLITE_NPROCESSES
     resources: mem_mb=ATLITE_NPROCESSES * 5000
-    wildcard_constraints: technology="^(?!hydro).*$" # Any technology other than hydro
+    wildcard_constraints: technology="(?!hydro).*" # Any technology other than hydro
     script: "scripts/build_renewable_profiles.py"
 
 
--- a/Snakefile
+++ b/Snakefile
@@ -201,19 +201,19 @@ rule build_renewable_profiles:
     benchmark: "benchmarks/build_renewable_profiles_{technology}"
     threads: ATLITE_NPROCESSES
     resources: mem_mb=ATLITE_NPROCESSES * 5000
+    wildcard_constraints: technology="solar|onwind|offwind-ac|offwind-dc"
     script: "scripts/build_renewable_profiles.py"
 
 
-if 'hydro' in config['renewable'].keys():
-    rule build_hydro_profile:
-        input:
-            country_shapes='resources/country_shapes.geojson',
-            eia_hydro_generation='data/bundle/EIA_hydro_generation_2000_2014.csv',
-            cutout="cutouts/" + config["renewable"]['hydro']['cutout'] + ".nc"
-        output: 'resources/profile_hydro.nc'
-        log: "logs/build_hydro_profile.log"
-        resources: mem_mb=5000
-        script: 'scripts/build_hydro_profile.py'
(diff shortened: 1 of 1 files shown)


euronion:
Ok, CI failed because the test `config.yaml` did not contain the `hydro` key. The new push should fix that.

The aim was to make not the rule optional, but present a more deterministic set of rules which is not based on the existance of config values.

Output in case `build_hydro_profiles` is called now without an `hydro` key being defined in the config:

```
$ snakemake -call -n resources/profile_hydro.nc
Building DAG of jobs...
MissingInputException in line 208 of .../Snakefile:
Missing input files for rule build_hydro_profile:
config['renewable']['hydro']['cutout'] not configured
```

fneum:
What about a ruleorder?

euronion:
Rule order makes sense IMHO if the rules are ambigious. In our case the rules don't need to be ambigious, so I'd prefer to use that route rather than keeping them ambigiuous.

Snakefile
```diff
@@ -201,19 +201,19 @@ rule build_renewable_profiles:
     benchmark: "benchmarks/build_renewable_profiles_{technology}"
     threads: ATLITE_NPROCESSES
     resources: mem_mb=ATLITE_NPROCESSES * 5000
+    wildcard_constraints: technology="solar|onwind|offwind-ac|offwind-dc"
```
fneum, line 204: Can we also do this backward?, i.e. ~hydro? Reason is that it's quite straightforward to add new technologies just through the config.yaml under `renewable`. E.g. you could just provide a new section `offwind-floating` and it would work.

Snakefile
```diff
@@ -201,19 +201,19 @@ rule build_renewable_profiles:
...
-    rule build_hydro_profile:
-        input:
-            country_shapes='resources/country_shapes.geojson',
-            eia_hydro_generation='data/bundle/EIA_hydro_generation_2000_2014.csv',
-            cutout="cutouts/" + config["renewable"]['hydro']['cutout'] + ".nc"
-        output: 'resources/profile_hydro.nc'
-        log: "logs/build_hydro_profile.log"
-        resources: mem_mb=5000
-        script: 'scripts/build_hydro_profile.py'
+rule build_hydro_profile:
```
fneum, line 208: great, this is a better solution IMO

euronion: You mean the old one with the optional include of the rule? Or the new one?

Snakefile
euronion: Sure, no strong preference there. Updated.

Snakefile
fneum: the new implementation. Sorry that wasn't clear.

euronion: ok

fneum (approved):

PR #336, opened 2022-03-20
remove build/retrieve natura raster, directly use shapefile

### LINKED ISSUE #334: Drop `build_natura_raster` and `retrieve_natura_raster`
With the new `atlite` exclusion calculator it should be straightforward to directly use the NATURA shapefile rather than rasterising first.

fneum:
Closes #334  (if applicable).

- removes build_natura_raster
- removes retrieve_natura_raster
- directly loads NATURA 2000 shapefile in build_renewable_profiles exclusion container
- removes NATURA eligibility analysis from tests due to file size of shape
- uses mask to selectively load NATURA dataset

Adds 15-20 seconds to each run of build_renewable_profiles. IMO this is acceptable given the reduction of code.

## Checklist


euronion:
I dislike the idea of excluding natura from the `build_renewable_profiles` simply due to file size; that's a hotfix addressing symptoms rather than the problem. The CI should run on as close a representation of the original data as possible. I think this should be a motivation to implement caching of the *full* databundle to use with the CI and switch the testing away from the tutorial databundle to the real databundle?

Some more conceptual thoughts as your PR is already touching the natura data:

* I suggest to switch from the shapefile to the geopackage version of natura: https://sdi.eea.europa.eu/data/b1777027-6c85-4d19-bdf2-5840184d6e13?path=%2FNatura2000_end2020_gpkg . This reduces the number of natura files to 1 (from multi-file fileformat SHP). The file size doesn't change significantly.
* I suggest to implement a `download_natura2000` rule which downloads the data from its original source.
* The natura data in the databundle should be updated accordingly (GPKG instead of SHP)

I think for the databundle our aim in general should be:
* Have individual rules for downloading each datafile from its original source
* Have the databundle mirror the original sources for archiving purposes

PR #345, opened 2022-03-24
clean out all nuts3 references

fneum:
The goal is to eliminate shape reprojection from NUTS3 to model regions. Instead, use gridded data from 

https://datadryad.org/stash/dataset/doi:10.5061/dryad.dk1j0

and 

https://www.worldpop.org/

Checkout:

https://github.com/martacki/pypsa-ua_gdp_data

### DIFF
--- a/Snakefile
+++ b/Snakefile
@@ -149,16 +149,12 @@ rule build_shapes:
     input:
         naturalearth='data/naturalearth/ne_10m_admin_0_countries.shp',
         eez='data/eez/World_EEZ_v11_20191118_gpkg/eez_v11.gpkg',
-        nuts3='data/bundle/NUTS_2013_60M_SH/data/NUTS_RG_60M_2013.shp',
-        nuts3pop='data/bundle/nama_10r_3popgdp.tsv.gz',
-        nuts3gdp='data/bundle/nama_10r_3gdp.tsv.gz',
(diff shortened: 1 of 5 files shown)


fneum:
So using 5 arcmin dataset looks not quite close to NUTS3-level dataset

image

euronion:
Maybe mismatching units? NUTS3 dataset was pop in abs. values IIRC, the gridded pop data should be pop density.

martacki:
dunno if you intend to use another formula now. For ua and md, that seems to be the currently best solution, however I have the feeling that for germany the current nuts3 data provides a much better distribution...

nuts3 for germany:
nuts3_germany
suggested data sources (according to the formulas used in my private repo that you're quoting):
gdp+pop_germany

fneum:
I just wanted to sketch how it could work based on what you did for UA/MD. The motivation would be that it simplifies the code (the whole nuts3 function would be superfluous) and we could possible go to higher resolutions than nuts3.

euronion:
@FabianHofmann Can we reuse the `atlite` code used for the `ExclusionCalculator` to implement the calculation faster than what `xagg` has to offer? (Determine overlaps between `regions` / shapes and a raster / the population raster data).

fneum:
wontfix

PR #348, opened 2022-03-28
add_electricity: remove vresutils.costdata.annuity dependency

fneum:


scripts/add_electricity.py
```diff
@@ -105,6 +104,18 @@
 def normed(s): return s/s.sum()
 
 
+def annuity(n, r):
```
euronion, line 107:
```suggestion
def calculate_annuity(n, r):
```

fneum: Done!

PR #349, opened 2022-03-28
build_bus_regions: move voronoi partition from `vresutils` to script

fneum:
Up for discussion: This would be a step towards eliminating all `vresutils` dependencies.

### DIFF
--- a/scripts/build_bus_regions.py
+++ b/scripts/build_bus_regions.py
@@ -47,9 +47,10 @@ from _helpers import configure_logging
 import pypsa
 import os
 import pandas as pd
+import numpy as np
 import geopandas as gpd
-
-from vresutils.graph import voronoi_partition_pts
+from shapely.geometry import Polygon
+from scipy.spatial import Voronoi
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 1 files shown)


scripts/build_bus_regions.py
```diff
@@ -61,6 +62,66 @@ def save_to_geojson(s, fn):
...
+                # for a MultiPolygon pick the part with the largest area
+                poly = max(poly.geoms, key=lambda pg: pg.area)
+            except:
+                pass
+            return poly
+        polygons = [demultipolygon(poly) for poly in polygons]
+
+    polygons_arr = np.empty((len(polygons),), 'object')
+    polygons_arr[:] = polygons
+    return polygons_arr
```
FabianHofmann, line 122:
```suggestion
    return np.array(polygons, dtype=object)
```

FabianHofmann (commented):
Is the `no_multipolygons` tag important?

fneum:
> Is the no_multipolygons tag important?

It's not used anywhere. Fine to remove in my opinion. We can add back in from `vresutils` should the need occur later.

scripts/build_bus_regions.py
```diff
@@ -61,6 +62,55 @@ def save_to_geojson(s, fn):
     s.to_file(fn, driver='GeoJSON', schema=schema)
 
 
+def voronoi_partition_pts(points, outline, no_multipolygons=False):
```
FabianHofmann, line 65:
```suggestion
def voronoi_partition_pts(points, outline):
```

scripts/build_bus_regions.py
```diff
@@ -61,6 +62,55 @@ def save_to_geojson(s, fn):
...
+    """
+    Compute the polygons of a voronoi partition of `points` within the
+    polygon `outline`. Taken from
+    https://github.com/FRESNA/vresutils/blob/master/vresutils/graph.py
+    Attributes
+    ----------
+    points : Nx2 - ndarray[dtype=float]
+    outline : Polygon
+    no_multipolygons : bool (default: False)
+        If true, replace each MultiPolygon by its largest component
```
FabianHofmann, line 75:
```suggestion
```

fneum:
You happy?

FabianHofmann (approved):

PR #352, opened 2022-04-04
add existing wind and solar capacities based on IRENASTATS

fneum:
see release notes.


FabianHofmann:
> Looking at the data, this seems to be useful on global scale as well?

Indeed

PR #353, opened 2022-04-04
European energy security

fneum:
Since this branch includes important features, we should incorporate the PR, even though it is quite large. The features in compact are

	
* Added existing renewable capacities for all countries based on IRENA statistics (IRENASTAT) using new ``powerplantmatching`` version:
  * The corresponding ``config`` entries changed, cf. ``config.default.yaml``:	
    * old: ``estimate_renewable_capacities_from_capacity_stats``	
    * new: ``estimate_renewable_capacities``	
  * The estimation is endabled by setting the subkey ``enable`` to ``True``.	
  * Configuration of reference year for capacities can be configured (default: ``2020``) 	
  * The list of renewables provided by the OPSD database can be used as a basis, using the tag ``from_opsd: True``. This adds the renewables from the database and fills up the missing capacities with the heuristic distribution. 	
  * Uniform expansion limit of renewable build-up based on existing capacities can be configured using ``expansion_limit`` option	
    (default: ``false``; limited to determined renewable potentials)	
  * Distribution of country-level capacities proportional to maximum annual energy yield for each bus region	

* The config key ``renewable_capacities_from_OPSD`` is deprecated and was moved under the section, ``estimate_renewable_capacities``. To enable it, set ``from_opsd`` to `True`.	

* Add operational reserve margin constraint analogous to `GenX implementation <https://genxproject.github.io/GenX/dev/core/#Reserves>`_.	
  Can be activated with config setting ``electricity: operational_reserve:``.	

* Add function to add global constraint on use of gas in :mod:`prepare_network`. This can be activated with `electricity: gaslimit:` given in MWh.	

* Add configuration option to implement Energy Availability Factors (EAFs) for conventional generation technologies.	

* A new section ``conventional`` was added to the config file. This section contains configurations for conventional carriers.  

* Implement country-specific EAFs for nuclear power plants based on IAEA 2018-2020 reported country averages. These are specified under the ``energy_availibility_factor`` key in the config entry ``conventional`` and specify the static `p_max_pu` values.

* The powerplants that have been shut down before 2021 are filtered out. 	

* ``powerplantmatching>=0.5.1`` is now required for ``IRENASTATS``.	

* The inclusion of renewable carriers is now specified in the config entry ``renewable_carriers``. Before this was done by commenting/uncommenting sub-sections in the `renewable` config section. 

* Now, all carriers that should be extendable have to be listed in the config entry ``extendable_carriers``. Before, renewable carriers were always set to be extendable. For backwards compatibility, the workflow is still looking at the listed carriers under the ``renewable`` key. In the future, all of them have to be listed under ``extendable_carriers``. 	

* It is now possible to set conventional power plants as extendable by adding them to the list of extendable ``Generator`` carriers in the config.	

* Listing conventional carriers in ``extendable_carriers`` but not in ``conventional_carriers``, sets the corresponding conventional power plants as extendable without a lower capacity bound of today's capacities.	

* Now, conventional carriers have an assigned capital cost by default.	

* The ``build_year`` and ``lifetime`` column are now defined for conventional power plants. 	 

# Project Roadmap

Below, the project roadmap which we followed for the gas stop . 

## Research on input assumptions 

- [ ] Distribution of Biogas potentials in Europe based on EUROSTAT 
- [x] Historic power output of CHP's in order to retrieve must-run requirements ( @pz-max @chrstphtrs).
   - [x] regular summer period  
   - [x] regualt winter period - Can we quantify the effect of district heating requirements?  
   - [x] war period - How did plants react on price shocks? Is this already representative as a "real" minimum requirement?
   - [x] Result. Plot created in this [repo](https://github.com/pz-max/eu-powerplant-validation) suggests that there was no obvious price-shock elasticity of the demand.
- ~~[ ] Assumptions on gas pipelines done by the princeton group ( @nworbmot, @Irieo )~~
- [x] Elasticity of the industrial gas demand ( @Irieo )

### Input Data

- [x] Up-to-date representation of conventional European power plants (@FabianHofmann @p-glaum)
  - [x] Update of `powerplantmatching` including latest shutdowns (powerplantmatching update 0.5.2)
  - [x] fix https://github.com/FRESNA/powerplantmatching/issues/64
  - [x] Validation of capacity totals for now (time points in the future like 2025 are of second order priority) 
  - [x] Validation of some random plants and plants that were shutdown just very recently
- [x] Up-to-date representation of existing renewable European power plants ( @fneum @euronion )
   - [x] New scaling of renewable capacity totals based on the IRENA data statistics - IRENA report on 2021 numbers should be due and released any time soon; but for now only 2020 data; #352 (with limited expansion potential)   (@fneum @euronion )
  - [x] Check heuristic RES capacity distribution against a country with site-level data, add plot ( @fneum )
  - ~[ ] Integration of renewable power plants from OPSD dataset to improve spatial distribution of capacities. Since OPSD is a bit out of date (several GW per country missing) a scaling of the OPSD renewable capacities from IRENA might be necessary. (optional)~
- [ ] Distinction of CHP and PP. CHP should be added to the model using a dedicated carrier in order to implement must-run conditions ( @pz-max @chrstphtrs)
- [x] ~Update [OPSD electricity demand time series ](https://data.open-power-system-data.org/time_series).~ Discarded for now because data net added would only be from May 2019 to mid-2020. New dataset only contains data from 2015-2020, so would have to keep both.
- [ ] Preparations to run multiple weather years.
  - [x] Download cutouts. ( @fneum )
  - [ ] Leap years! Remove 29 February always! Or don't pick leap years.
- [x] Make it easy to run through multiple configfiles without needing to duplicate the repository. PyPSA-Eur-Sec gives inspiration with "run-name" folders as folder prefixes. (@fneum )
- [x] Correct representation of current network data: Some TYNDP lines are built by now, their existing capacities are however discarded because they are uniformly assumed to be under construction (https://github.com/pypsa/pypsa-eur/blob/b63271f12f2e9d1ffab8d2cdfaae837a4d70492d/scripts/base_network.py#L238), but https://github.com/pypsa/pypsa-eur/blob/2fc1ea0255bdffe45cc9158a7463c2508bb02b8f/data/links_tyndp.csv includes a "built" status now. https://github.com/PyPSA/pypsa-eur/pull/354 BTW: for status quo studies the config.yaml setting should be `under_construction: 'remove'`. ( @fneum )

### Optimization 
- [x] Add reserve margin constraints (@FabianHofmann, @fneum  ) #358 
- [ ] Add must-run requirements for CHP plants ( @FabianHofmann @pz-max @chrstphtrs)
- [x] Global gas consumption limit 
 / Add [global constraint](https://pypsa.readthedocs.io/en/latest/components.html#global-constraints) on primary energy usage of natural gas or other carriers (@fneum ) (confirmed to be working correctly)
- [x] Nuclear power plants, projection of temporary shutdowns and/or maximum energy outputs p.u.

Did we forget something?

### Scenarios to model
- [x] baseline
- [ ] different years
- [x] less gas
- [ ] more expensive gas
- [ ] with & w/o German nuclear shutdown (& restart)
- [ ] forced min/max gas volumes in storage


fneum:
## Existing Wind and Solar Capacities

in GW:

| bus   |   offwind |   onwind |   solar |
|:------|----------:|---------:|--------:|
| AL    |         0 |        0 |       0 |
| AT    |         0 |        3 |       2 |
| BA    |         0 |        0 |       0 |
| BE    |         2 |        2 |       5 |
| BG    |         0 |        0 |       1 |
| CH    |         0 |        0 |       2 |
| CZ    |         0 |        0 |       2 |
| DE    |         7 |       54 |      53 |
| DK    |         1 |        4 |       1 |
| EE    |         0 |        0 |       0 |
| ES    |         0 |       27 |      11 |
| FI    |         0 |        2 |       0 |
| FR    |         0 |       17 |      11 |
| GB    |        10 |       14 |      13 |
| GR    |         0 |        4 |       3 |
| HR    |         0 |        0 |       0 |
| HU    |         0 |        0 |       1 |
| IE    |         0 |        4 |       0 |
| IT    |         0 |       10 |      21 |
| LT    |         0 |        0 |       0 |
| LU    |         0 |        0 |       0 |
| LV    |         0 |        0 |       0 |
| ME    |         0 |        0 |       0 |
| MK    |         0 |        0 |       0 |
| NL    |         2 |        4 |      10 |
| NO    |         0 |        3 |       0 |
| PL    |         0 |        6 |       3 |
| PT    |         0 |        5 |       1 |
| RO    |         0 |        3 |       1 |
| RS    |         0 |        0 |       0 |
| SE    |         0 |        9 |       1 |
| SI    |         0 |        0 |       0 |
| SK    |         0 |        0 |       0 |

The distribution of capacities according to heuristic: "maximum annual energy yield":

solar-p_nom
onwind-p_nom
offwind-p_nom

Snakefile
```diff
@@ -84,7 +84,7 @@ rule build_powerplants:
     input:
         base_network="networks/base.nc",
```
martacki, line 85: why doesn't `base.nc` (and other associated, like `regions_onshore` etc.) go in RDIR as well? Depending on config, could be a different selection of countries etc. and could produce a different output...

fneum: Compromise / practical consideration. I wanted to avoid the build_renewable_profiles is re-run for every new config file.

scripts/add_electricity.py
```diff
@@ -301,25 +302,46 @@ def attach_wind_and_solar(n, costs, input_profiles, technologies, line_length_fa
...
+        factors = config[carrier].get("energy_availability_factors")
+
+        if isinstance(v, float):
+            # Single value affecting all generators of technology k indiscriminantely of country
+            n.generators.loc[idx, "p_max_pu"] = v
+        elif isinstance(v, dict):
+            v = pd.Series(v)
+            # Values affecting generators of technology k country-specific
+            # First map generator buses to countries; then map countries to p_max_pu
+            n.generators.p_max_pu.update(n.generators.loc[idx].bus.map(v).dropna())
```
martacki, line 343: updated also the formula, not only variable names. thanks for the description in the comments! unfortunately the formula given in the code didn't do what expected (for me). the following however did:

```suggestion
        if isinstance(factors, float):
            # Single value affecting all generators of technology k indiscriminantely of country
            n.generators.loc[idx, "p_max_pu"] = factors
        elif isinstance(factors, dict):
            factors = pd.Series(factors)
            # Values affecting generators of technology k country-specific
            # First map generator buses to countries; then map countries to p_max_pu
            (n.generators.p_max_pu.update(
                n.generators.loc[idx].bus.map(n.buses.country).dropna().map(factors))
            )
```

scripts/add_electricity.py
```diff
@@ -556,30 +576,65 @@ def add_nice_carrier_names(n, config):
...
+
 
     attach_load(n, snakemake.input.regions, snakemake.input.load, snakemake.input.nuts3_shapes,
                 snakemake.config['countries'], snakemake.config['load']['scaling_factor'])
 
     update_transmission_costs(n, costs, snakemake.config['lines']['length_factor'])
 
-    carriers = snakemake.config['electricity']['conventional_carriers']
-    attach_conventional_generators(n, costs, ppl, carriers)
+    attach_conventional_generators(n, costs, ppl, conventional_carriers, extendable_carriers)
```
martacki, line 602: why not include the fancy capacity updates?

```suggestion
    attach_conventional_generators(n, costs, ppl, conventional_carriers, extendable_carriers, **snakemake.config["conventional"])
```

scripts/add_electricity.py
```diff
@@ -556,30 +576,65 @@ def add_nice_carrier_names(n, config):
 
     costs = load_costs(snakemake.input.tech_costs, snakemake.config['costs'], snakemake.config['electricity'], Nyears)
     ppl = load_powerplants(snakemake.input.powerplants)
+    
+    if "renewable_carriers" in snakemake.config['electricity']:    
+        renewable_carriers = set(snakemake.config['renewable'])
+    else: 
+        logger.warning("Missing key `renewable_carriers` under config entry `electricity`. "
+                       "In future versions, this will raise an error. "
+                       "Falling back to carriers listed under `renewable`.")
```
martacki, line 585: not sure if I get this right, but what if one does not want to include renewables in the network?

martacki (commented):
Sorry I didn't make it through all the scripts (yet), but I have a couple of questions and comments for a subset of the changes, mainly regarding the add_electricity script. Really like the improvement of the default network assumptions! I think this really does improve the model overall

config.default.yaml
```diff
@@ -39,29 +37,50 @@ enable:
...
   co2limit: 7.75e+7 # 0.05 * 3.1e9*0.5
   co2base: 1.487e+9
   agg_p_nom_limits: data/agg_p_nom_minmax.csv
 
-  extendable_carriers:
-    Generator: []
-    StorageUnit: [] # battery, H2
-    Store: [battery, H2]
-    Link: []
+  operational_reserve: # like https://genxproject.github.io/GenX/dev/core/#Reserves
```
fneum, line 45: We should disable `operational_reserve` by default.

scripts/cluster_network.py
```diff
@@ -349,7 +352,7 @@ def plot_busmap_for_n_clusters(n, n_clusters, fn=None):
 
     if snakemake.wildcards.clusters.endswith('m'):
         n_clusters = int(snakemake.wildcards.clusters[:-1])
-        aggregate_carriers = pd.Index(n.generators.carrier.unique()).difference(renewable_carriers)
+        aggregate_carriers = None
```
martacki, line 355: This will break the intention of this method. Either keep as before, or change to:
```suggestion
        aggregate_carriers = snakemake.config["electricity"].get("conventional_carriers")
```
However, with my suggestion `ror` would be also higher resolved. The old version did cluster `ror` to the same resolution as conventional plants. But this might have been a minor bug, now that I think of it.

martacki: Also, I see that there's a bug in case `aggregate_carriers` is an empty list. But theoretically, you *could* run a simulation without conventional carriers and want to use the `m`-wildcard option.... I'll put that in an issue. :drop:

scripts/add_electricity.py
```diff
@@ -501,38 +515,44 @@ def attach_OPSD_renewables(n, techs):
...
+    expansion_limit = config["electricity"]["estimate_renewable_capacities"]["expansion_limit"]
 
-    capacities = (pm.data.Capacity_stats().powerplant.convert_country_to_alpha2()
-                  [lambda df: df.Energy_Source_Level_2]
-                  .set_index(['Fueltype', 'Country']).sort_index())
+    if not len(countries) or not len(tech_map): return
 
-    countries = n.buses.country.unique()
+    capacities = pm.data.IRENASTAT().powerplant.convert_country_to_alpha2()
+    capacities = capacities.query("Year == @year and Technology in @tech_map and Country in @countries")
```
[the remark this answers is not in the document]
FabianHofmann: Acutally no, capacities contains the total installed capacities for each year (not the new installations)

scripts/add_electricity.py
```diff
@@ -301,25 +302,46 @@ def attach_wind_and_solar(n, costs, input_profiles, technologies, line_length_fa
                    p_max_pu=ds['profile'].transpose('time', 'bus').to_pandas())
 
 
-def attach_conventional_generators(n, costs, ppl, carriers):
+def attach_conventional_generators(n, costs, ppl, conventional_carriers, extendable_carriers, **config):
 
+    carriers = set(conventional_carriers) | set(extendable_carriers['Generator'])
```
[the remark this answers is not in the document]
FabianHofmann: you are totally right, that was a bit laziness from my side. If you don't mind i'd like to postpone it until the `renewable_carriers` and `conventional_carriers` are  established. Then we can easily import all of them as sets and pass the `extendable_carriers` as 
```python 
extendable_carriers - renewable_carriers
``` 
to this functions.

scripts/cluster_network.py
FabianHofmann: great, thanks for spotting that

config.default.yaml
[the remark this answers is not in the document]
FabianHofmann: added to the function itsefl

scripts/add_electricity.py
FabianHofmann: perhaps we should just require renewables?

FabianHofmann: perhaps we should just require renewables?

scripts/build_powerplants.py
```diff
@@ -122,21 +131,21 @@ def add_custom_powerplants(ppl, custom_powerplants, custom_ppl_query=False):
...
-        logging.warning(f"No powerplants known in: {', '.join(cntries_without_ppl)}")
+    substations = n.buses.query('substation_lv')
+    ppl = map_country_bus(ppl, substations)
 
     bus_null_b = ppl["bus"].isnull()
     if bus_null_b.any():
-        logging.warning(f"Couldn't find close bus for {bus_null_b.sum()} powerplants")
+        logging.warning(f"Couldn't find close bus for {bus_null_b.sum()} powerplants. "
+                        "Removing them from the powerplants list.")
+        ppl = ppl[~bus_null_b]
```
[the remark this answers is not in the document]
FabianHofmann: should be the same right? It is just filtering out nan's, meaning the ppl which did not get a bus assigned. (and the implemented version is a bit faster)

FabianHofmann (commented):
@martacki and @fneum thank you both for the comments, I think addressed nearly all of them, I make some final tests now.

martacki:
> @martacki and @fneum thank you both for the comments, I think addressed nearly all of them, I make some final tests now.

Hope I could help! Though I didn't have the time to go through `prepare_network` & `solve_network` yet as I didn't have the time...

Also, don't forget to upload the `data/nuclear-eafs.csv` - and there's a bug in `add_electricity` pandas doesn't have a `.read_file(...)`, should be `.read_csv(...)` instead; but I'm sure you'll catch that in your runs.

doc/release_notes.rst
```diff
@@ -7,6 +7,79 @@
...
+* Add operational reserve margin constraint analogous to `GenX implementation <https://genxproject.github.io/GenX/dev/core/#Reserves>`_.	
+  Can be activated with config setting ``electricity: operational_reserve:``.	
+
+* Add function to add global constraint on use of gas in :mod:`prepare_network`. This can be activated with `electricity: gaslimit:` given in MWh.	
+
+* Add configuration option to implement Energy Availability Factors (EAFs) for conventional generation technologies.	
+
+* A new section ``conventional`` was added to the config file. This section contains configurations for conventional carriers.  
+
+* Implement country-specific EAFs for nuclear power plants based on IAEA 2018-2020 reported country averages. These are specified under the ``energy_availibility_factor`` key in the config entry ``conventional`` and specify the static `p_max_pu` values.
```
martacki, line 60: but i'm not sure if it wouldn't be better to specify the path in the Snakefile instead of the config file. Up to you, I'd say

```suggestion
* Implement country-specific EAFs for nuclear power plants based on IAEA 2018-2020 reported country averages. These are specified in ``data/nuclear-eafs.csv`` and specify the static `p_max_pu` values.
```

martacki:
Oh, and maybe add a decription as well in `doc/configtables/opts.csv` for `CH4L`?

scripts/prepare_network.py
```diff
@@ -233,8 +243,22 @@ def set_line_nom_max(n, s_nom_max_set=np.inf, p_nom_max_set=np.inf):
...
+    for o in opts:
+        if "CH4L" in o:
+            m = re.findall("[0-9]*\.?[0-9]+$", o)
+            if len(m) > 0:
+                limit = float(m[0]) * 1e6
+                add_gaslimit(n, limit, Nyears)
+                logger.info("Setting gas usage limit according to wildcard value.")
+            else:
+                add_gaslimit(n, snakemake.config["electricity"].get("gaslimit"), Nyears)
+                logger.info("Setting gas usage limit according to config value.")
```
martacki, line 261: so for the wildcard value, the limit is specified in... TWh, while the config value must be given in MWh? (release_notes indicate that limit to be given in MWh in any case, if I remember correctly)

FabianHofmann:
> Also, don't forget to upload the `data/nuclear-eafs.csv` - and there's a bug in `add_electricity` pandas doesn't have a `.read_file(...)`, should be `.read_csv(...)` instead; but I'm sure you'll catch that in your runs.

Correct, is done!

config.default.yaml
```diff
@@ -166,6 +185,10 @@ renewable:
     hydro_max_hours: "energy_capacity_totals_by_country" # one of energy_capacity_totals_by_country, estimate_by_large_installations or a float
     clip_min_inflow: 1.0
 
+conventional: 
+  nuclear:
+    energy_availability_factors: "data/nuclear-eafs.csv" # float of file name
```
fneum, line 190: This should maybe show up in `snakemake.inputs` of the rule where it is used.

Option A: make this config setting a boolean, fixed file reference to `data/nuclear-eafs.csv` in Snakefile

Option B: refer to file referenced here as optional input in `snakemake.inputs`: `eaf=config["conventional"]["nuclear"]["energy_availability_factors"] if config["conventional"]["nuclear"]["energy_availability_factors"] else []

scripts/add_electricity.py
```diff
@@ -301,25 +302,47 @@ def attach_wind_and_solar(n, costs, input_profiles, technologies, line_length_fa
...
-           p_nom=ppl.p_nom,
+           p_nom_min=ppl.p_nom.where(ppl.carrier.isin(conventional_carriers), 0),
+           p_nom=ppl.p_nom.where(ppl.carrier.isin(conventional_carriers), 0),
+           p_nom_extendable=ppl.carrier.isin(extendable_carriers['Generator']),
            efficiency=ppl.efficiency,
            marginal_cost=ppl.marginal_cost,
-           capital_cost=0)
+           capital_cost=ppl.capital_cost,
+           build_year=ppl.datein.fillna(0).astype(int),
+           lifetime=(ppl.dateout - ppl.datein).fillna(9999).astype(int),
```
fneum, line 327: Isn't the default for lifetime `np.inf`? 
```suggestion
           lifetime=(ppl.dateout - ppl.datein).fillna(np.inf),
```

scripts/add_electricity.py
```diff
@@ -423,7 +446,7 @@ def attach_hydro(n, costs, ppl, profile_hydro, hydro_capacities, carriers, **con
 
 
 def attach_extendable_generators(n, costs, ppl, carriers):
-
+    logger.warning("The function `attach_extendable_generators` is deprecated in v0.0.5.")
```
fneum, line 449:
```suggestion
    logger.warning("The function `attach_extendable_generators` is deprecated in v0.5.0.")
```

doc/release_notes.rst
```diff
@@ -7,6 +7,85 @@
...
+  * Uniform expansion limit of renewable build-up based on existing capacities can be configured using ``expansion_limit`` option	
+    (default: ``false``; limited to determined renewable potentials)	
+  * Distribution of country-level capacities proportional to maximum annual energy yield for each bus region	
+
+* The config key ``renewable_capacities_from_OPSD`` is deprecated and was moved under the section, ``estimate_renewable_capacities``. To enable it, set ``from_opsd`` to `True`.	
+
+* Add operational reserve margin constraint analogous to `GenX implementation <https://genxproject.github.io/GenX/dev/core/#Reserves>`_.	
+  Can be activated with config setting ``electricity: operational_reserve:``.	
+
+* Add function to add global constraint on use of gas in :mod:`prepare_network`. This can be activated with `electricity: gaslimit:` given in MWh.
```
fneum, line 54:
```suggestion
* Add function to add global constraint on use of gas in :mod:`prepare_network`. This can be activated by including the keyword ``CH4L`` in the ``{opts}`` wildcard which enforces the limit set in ``electricity: gaslimit:`` given in MWh thermal. Alternatively, it is possible to append a number in the `{opts}` wildcard, e.g. `CH4L200` which limits the gas use to 200 TWh thermal.
```

doc/release_notes.rst
```diff
@@ -7,6 +7,85 @@
...
+* Fix crs bug. Change crs 4236 to 4326.
+
+* Update rasterio version to correctly calculate exclusion raster
+
+* Remove rules to build or retrieve rasterized NATURA 2000 dataset. Renewable potential calculation now directly uses the shapefiles.
+
+* Cache data and cutouts folders. This cache will be updated weekly.
+
+* Add rule to automatically retrieve Natura2000 natural protection areas. Switch of file format to GPKG.
+
```
fneum, line 87:
```suggestion
* Add option to set CO2 emission prices through `{opts}` wildcard: `Ep<number>`, e.g. `Ep180`, will set the EUR/tCO2 price.

* Add option to alter marginal costs of a carrier through `{opts}` wildcard: `<carrier>+m<factor>`, e.g. `gas+m2.5`, will multiply the default marginal cost for gas by factor 2.5.
```

scripts/prepare_network.py
fneum: I made a suggestion above to make this difference in unit clear.

doc/release_notes.rst
```diff
@@ -7,6 +7,85 @@
...
+* Add operational reserve margin constraint analogous to `GenX implementation <https://genxproject.github.io/GenX/dev/core/#Reserves>`_.	
+  Can be activated with config setting ``electricity: operational_reserve:``.	
+
+* Add function to add global constraint on use of gas in :mod:`prepare_network`. This can be activated with `electricity: gaslimit:` given in MWh.	
+
+* Add configuration option to implement Energy Availability Factors (EAFs) for conventional generation technologies.	
+
+* A new section ``conventional`` was added to the config file. This section contains configurations for conventional carriers.  
+
+* Implement country-specific EAFs for nuclear power plants based on IAEA 2018-2020 reported country averages. These are specified `data/nuclear_eafs.csv` and translate to static `p_max_pu` values.
```
fneum, line 60: .rst :)
```suggestion
* Implement country-specific EAFs for nuclear power plants based on IAEA 2018-2020 reported country averages. These are specified ``data/nuclear_eafs.csv`` and translate to static ``p_max_pu`` values.
```

doc/release_notes.rst
```diff
@@ -7,6 +7,85 @@
...
+  [`#277 <https://github.com/PyPSA/pypsa-eur/pull/277>`_].
+
+* The default deployment density of AC- and DC-connected offshore wind capacity is reduced from 3 MW/sqkm
+  to a more conservative estimate of 2 MW/sqkm [`#280 <https://github.com/PyPSA/pypsa-eur/pull/280>`_].
+
+* Following discussion in `#285 <https://github.com/PyPSA/pypsa-eur/issues/285>`_ we have disabled the
+  correction factor for solar PV capacity factors by default while satellite data is used.
+  A correction factor of 0.854337 is recommended if reanalysis data like ERA5 is used.
+
+* Resource definitions for memory usage now follow [Snakemake standard resource definition](https://snakemake.readthedocs.io/en/stable/snakefiles/rules.html#standard-resources) ```mem_mb`` rather than ``mem``.
```
fneum, line 24: .rst :)
```suggestion
* Resource definitions for memory usage now follow `Snakemake standard resource definition <https://snakemake.readthedocs.io/en/stable/snakefiles/rules.html#standard-resources>`_ ``mem_mb`` rather than ``mem``.
```

fneum (commented):
Almost good to go in! Just a few finishing touches

config.default.yaml
fneum: Now with the general implementation, I wouldn’t insist on handling all the possible file references in Snakefile.

fneum:
I will merge this around mid day unless there are any objections.

FabianHofmann:
> I will merge this around mid day unless there are any objections.

last changes from my side in #388

fneum (commented):
Approved from my side.

FabianHofmann:
Workflow runs through with default settings. Let's do it!

pz-max:
ooooh boy... That will take time to implement in PyPSA-Africa :dancers: 
FYI @davide-f

davide-f:
Really a lot to do xD
I really love the reserve requirements spec! That is a major priority I think

PR #358, opened 2022-04-07
Add operational reserve margin constraint analogous to GenX

fneum:
Co-authored-by: @FabianHofmann


FabianHofmann (commented):
Nice, can we add a dedicated on and off switch (without having to comment out)?

PR #323, opened 2022-03-04
Fix deprecation warning for 'closed' with pandas 1.4

euronion:
DNMY. Merge with next-next release. Will force `pandas>=1.4`.

## Checklist

- [x] I tested my contribution locally and it seems to work fine.
- [x] Code and workflow changes are sufficiently documented.
- [x] Newly introduced dependencies are added to `envs/environment.yaml` and `envs/environment.docs.yaml`.
- [x] Changes in configuration options are added in all of `config.default.yaml`, `config.tutorial.yaml`, and `test/config.test1.yaml`.
- [x] Changes in configuration options are also documented in `doc/configtables/*.csv` and line references are adjusted in `doc/configuration.rst` and `doc/tutorial.rst`.
- [x] A note for the release notes `doc/release_notes.rst` is amended in the format of previous release notes.

PR #365, opened 2022-04-29
bugfixes for manual load adjustments

martacki:
If manual_adjustment exceed the `DateRange` of `load`, errors occur. This PR fixes this potential error by re-reading the raw load data for the required `DateRange`.
At the same time, fix `FutureWaring` where argument `closed` is deprecated in favor of `inclusive` in `config.default.yaml` & `config.tutorial.yaml`. 

## Checklist

### DIFF
--- a/config.default.yaml
+++ b/config.default.yaml
@@ -26,7 +26,7 @@ clustering:
 snapshots:
   start: "2013-01-01"
   end: "2014-01-01"
-  closed: 'left' # end is not inclusive
+  inclusive: 'left' # end is not inclusive
 
 enable:
   prepare_links_p_nom: false
--- a/config.default.yaml
+++ b/config.default.yaml
@@ -26,7 +26,7 @@ clustering:
 snapshots:
   start: "2013-01-01"
   end: "2014-01-01"
-  inclusive: 'left' # end is not inclusive
+  closed: 'left' # end is not inclusive
 
 enable:
   prepare_links_p_nom: false
--- a/config.tutorial.yaml
+++ b/config.tutorial.yaml
@@ -26,7 +26,7 @@ clustering:
 snapshots:
   start: "2013-03-01"
   end: "2013-04-01"
-  inclusive: 'left' # end is not inclusive
+  closed: 'left' # end is not inclusive
 
 enable:
   prepare_links_p_nom: false
--- a/config.tutorial.yaml
+++ b/config.tutorial.yaml
@@ -26,7 +26,7 @@ clustering:
 snapshots:
   start: "2013-03-01"
   end: "2013-04-01"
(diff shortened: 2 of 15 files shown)


config.default.yaml
```diff
@@ -26,7 +26,7 @@ clustering:
 snapshots:
   start: "2013-01-01"
   end: "2014-01-01"
-  closed: 'left' # end is not inclusive
+  inclusive: 'left' # end is not inclusive
```
fneum, line 29: This will break compatibility with quite recent pandas versions. We'll do it once needed with #323

config.tutorial.yaml
```diff
@@ -26,7 +26,7 @@ clustering:
 snapshots:
   start: "2013-03-01"
   end: "2013-04-01"
-  closed: 'left' # end is not inclusive
+  inclusive: 'left' # end is not inclusive
```
fneum, line 29: see above

scripts/build_load_data.py
```diff
@@ -116,14 +116,19 @@ def max_consecutive_nans(ds):
...
     stop = pd.Timestamp(stop)
-    if start-delta in load.index and stop in load.index and cntry in load:
-        load.loc[start:stop, cntry] = load.loc[start-delta:stop-delta, cntry].values
+    if (start in load.index and stop in load.index):
+        if start-delta in load.index and stop-delta in load.index and cntry in load:
+            load.loc[start:stop, cntry] = load.loc[start-delta:stop-delta, cntry].values
+        elif fn_load is not None:
+            duration = pd.date_range(freq='h', start=start-delta, end=stop-delta)
+            load_raw = load_timeseries(fn_load, duration, [cntry], powerstatistics)
+            load.loc[start:stop, cntry] = load_raw.loc[start-delta:stop-delta, cntry].values
```
fneum, line 128: [Hazem-IEG](https://github.com/Hazem-IEG)

fneum, line 128: Could this be shorter by reading in the time series segments anew in any case? Would be good to avoid nested if/else blocks. I haven't fully understood when the previous implementation fails. Could you give an example?

martacki: Reading raw data takes a while, would try to omit whenever possible.

Breaks with current code for example for the year 2010, because whole January for GB is missing. The code tries to copy January from a previous/future year (I think future, but I'm not sure). However, those indices are not in `load`. Therefore, there's a length mismatch in the assignment: lhs has the length of January (744), rhs is empty.

config.default.yaml
martacki: ah, haven't seen this PR, sorry

fneum (approved):
LGTM

PR #372, opened 2022-06-03
restrict rasterio version

pz-max:
There is an issue in Atlite related to the rasterio version. The rasterio version 2.1.10 is causing troubles in the exclusion raster downsampling part: https://github.com/PyPSA/atlite/issues/238 . We solved this in Atlite already but need to wait for a new Atlite release. Until then the exclusion raster error can occur in PyPSA-Eur.

Until we update the Atlite version we should change the environment and constrain the rasterio version

## Checklist

- [x] Newly introduced dependencies are added to `envs/environment.yaml`.

### DIFF
--- a/doc/release_notes.rst
+++ b/doc/release_notes.rst
@@ -72,6 +72,8 @@ Upcoming Release
 
 * Fix crs bug. Change crs 4236 to 4326.
 
+* Update rasterio version to correctly calculate exclusion raster
+
 
 PyPSA-Eur 0.4.0 (22th September 2021)
 =====================================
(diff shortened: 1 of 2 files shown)


fneum:
Thanks @pz-max for this temporary fix! One could have also thought of installing the master of atlite instead (until the next version is released), as the fix is a bit indirect this way. But: This PR does what it ought to.

pz-max:
Good point. You are right. Next time the PR will be better🙂👍🏽

Get Outlook for iOS<https://aka.ms/o0ukef>
________________________________

PR #376, opened 2022-06-14
build_renewable_profiles: set show progress default to False

FabianHofmann:
Together with new `atlite` version fastens the landuse calculation by a factor of 4-5. So for whole of Europe, this takes around ~1 min with 6 cores.
Downside: One does not see the progressbar anymore, but I think it is a worthy trade-off.

### DIFF
--- a/scripts/build_renewable_profiles.py
+++ b/scripts/build_renewable_profiles.py
@@ -203,7 +203,7 @@ if __name__ == '__main__':
     pgb.streams.wrap_stderr()
 
     nprocesses = int(snakemake.threads)
-    noprogress = not snakemake.config['atlite'].get('show_progress', True)
+    noprogress = not snakemake.config['atlite'].get('show_progress', False)
     config = snakemake.config['renewable'][snakemake.wildcards.technology]
     resource = config['resource'] # pv panel config / wind turbine config
     correction_factor = config.get('correction_factor', 1.)


fneum:
I tend to agree that we wouldn't need a progress bar for a rule that takes 1 minute. Do you know what share of the speed up is due to disabling the progress bar and the new atlite version?

FabianHofmann:
Takes about 5 min with the progress bar. Could add an extra logging.

fneum:
Wow, I wasn't aware the progressbar makes things so slow. Let's disable the progressbar.

pz-max:
@FabianHofmann should we track this observed progress bar performance issue in Atlite?

The commit where the tqdm progressbar was introduced:
https://github.com/PyPSA/atlite/commit/6debc3707fef57f2407d726238f3d3d50ca8fd86

Options to optimise the progress bar in tqdm (reduces overhead):
https://github.com/tqdm/tqdm/blob/4f208e72552c4d916aa4fe6a955349ee8b2ed353/README.rst#monitoring-thread-intervals-and-miniters
The `dynamic_miniters` promises to adjust the monitoring intervals depending on the problem. Maybe then also short times can be effectively tracked then with a progress bar.

FabianHofmann:
@pz-max do you know whether this works for parallel subprocesses? As I understand, this is what is decreasing down the overall performance. tqdm waits for the callback of a individual process, and since these are not perfectly synchronized it sometimes keeps them on hold.

pz-max:
@FabianHofmann  this needs to be tested, I am not sure yet

PR #377, opened 2022-06-15
Rescale era5 wind speed with GWA3.1

p-glaum:
This PR provides a jupyter notebook with a draft method to rescale the era5 wind speed data with a bias correction factor. For the rescaling, the Global Wind Atlas (GWA) v3.1 is used. The atlite cutout and the GWA data must be downloaded manually. The GWA data can be downloaded from the following link: https://globalwindatlas.info/api/gis/global/wind-speed/100


pz-max:
From the Jupyter notebook:
image

p-glaum:
- do further validation of the wind speeds:
    - compare wind speed from atlite to GWA
    - compare power output with turbine power curve for GWA and atlite

PR #379, opened 2022-06-20
cluster-network: add strategies for conventionals

FabianHofmann:
Adds strategies for generators which might not have consensus values when clustered. This does not apply to any of the current power plants but might be necessary in the future.

### DIFF
--- a/scripts/cluster_network.py
+++ b/scripts/cluster_network.py
@@ -281,7 +281,14 @@ def clustering_for_n_clusters(n, n_clusters, custom_busmap=False, aggregate_carr
         aggregate_generators_carriers=aggregate_carriers,
         aggregate_one_ports=["Load", "StorageUnit"],
         line_length_factor=line_length_factor,
-        generator_strategies={'p_nom_max': p_nom_max_strategy, 'p_nom_min': pd.Series.sum},
+        generator_strategies={'p_nom_max': p_nom_max_strategy, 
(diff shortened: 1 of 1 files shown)


scripts/cluster_network.py
```diff
@@ -281,7 +281,14 @@ def clustering_for_n_clusters(n, n_clusters, custom_busmap=False, aggregate_carr
         aggregate_generators_carriers=aggregate_carriers,
         aggregate_one_ports=["Load", "StorageUnit"],
         line_length_factor=line_length_factor,
-        generator_strategies={'p_nom_max': p_nom_max_strategy, 'p_nom_min': pd.Series.sum},
+        generator_strategies={'p_nom_max': p_nom_max_strategy, 
+                              'p_nom_min': pd.Series.sum, 
+                              'p_min_pu': pd.Series.mean, 
+                              'marginal_cost': pd.Series.mean,
```
fneum, line 287: Is there an easy way to make this capacity-weighted?

fneum (approved):
No objection from my side.

scripts/cluster_network.py
FabianHofmann: Not that I knew of, in pypsa this is pretty much hard-coded, https://github.com/PyPSA/PyPSA/blob/00aa795f52056fa9f845151d353ca66703bf6873/pypsa/networkclustering.py#L92. Would be nice to have that option easily accessible

PR #382, opened 2022-06-21
Landlock country fix

### LINKED ISSUE #265: Support selection of only landlocked countries
Currently, if the user picks only landlocked countries (e.g. `['AT']`), the offshore shapes are empty which leads to an error when `geopandas` tries to write this empty GeoDataFrame to a file.

I don't have a good idea for a fix yet, since it has some downstream effects, too; e.g. `regions_offshore.geojson`. I would somehow want to avoid a case distinction in the `Snakefile`.

xref https://groups.google.com/g/pypsa/c/CgLgxBUwutw

pz-max:
Closes #265 
And also fixes an issue with hydro powerplants. See commit [3cc42ac](https://github.com/PyPSA/pypsa-eur/pull/382/commits/3cc42ac7a73d4aa40fdfe25bb44980252b074ad1). Without the fix, the following assert is triggered:
https://github.com/PyPSA/pypsa-eur/blob/4e60e8d6751bb0a60778a91a1b4aad290bd23fc2/scripts/add_electricity.py#L345-L346

- Create an empty offshore shape if none exists with extended `save_to_geojson()`.
- Harmonize `save_to_geojson()` and moving it to `_helpers.py` 
- Add a case distinction whenever offshore shapes are used.
- Add better UX for debugging by adding `os.chdir(os.path.dirname(os.path.abspath(__file__)))` above `mock_snakemake()`

## Test
- [x] 1 landlock country ["AT"] (Austria)
- [x] 2 landlock countries ["CZ","AT"] (Czech republic and Austria)
- [x] "DE", if CI runs through
## Checklist

- [ ] Changes in configuration options are added in all of `config.default.yaml`, `config.tutorial.yaml`, and `test/config.test1.yaml`. **-> A CI test for land-lock countries would be cool but probably is not necessary considering the overhead of updating files.**

### DIFF
--- a/scripts/add_electricity.py
+++ b/scripts/add_electricity.py
@@ -86,6 +86,7 @@ It further adds extendable ``generators`` with **zero** capacity for
 import logging
 from _helpers import configure_logging, update_p_nom_max
 
+import os
 import pypsa
 import pandas as pd
 import numpy as np
@@ -548,6 +549,7 @@ def add_nice_carrier_names(n, config):
 if __name__ == "__main__":
     if 'snakemake' not in globals():
         from _helpers import mock_snakemake
+        os.chdir(os.path.dirname(os.path.abspath(__file__)))
         snakemake = mock_snakemake('add_electricity')
     configure_logging(snakemake)
 
@@ -566,6 +568,10 @@ if __name__ == "__main__":
     attach_conventional_generators(n, costs, ppl, carriers)
 
     carriers = snakemake.config['renewable']
+    if os.stat(snakemake.input.offshore_region).st_size == 0:
+        logger.info("No offshore file exist. Remove offshore carriers")
+        carriers.pop("offwind-ac")
+        carriers.pop("offwind-dc")
     attach_wind_and_solar(n, costs, snakemake.input, carriers, snakemake.config['lines']['length_factor'])
 
     if 'hydro' in snakemake.config['renewable']:
--- a/scripts/add_electricity.py
+++ b/scripts/add_electricity.py
@@ -329,6 +329,10 @@ def attach_hydro(n, costs, ppl, profile_hydro, hydro_capacities, carriers, **con
 
     ppl = ppl.query('carrier == "hydro"').reset_index(drop=True)\
              .rename(index=lambda s: str(s) + ' hydro')
+    if len(ppl.loc[ppl.bus.isna()])>0:
+        # TODO: Why some hydro plants don't have a bus?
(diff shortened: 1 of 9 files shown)


scripts/build_bus_regions.py
```diff
@@ -112,15 +121,20 @@ def voronoi_partition_pts(points, outline):
 if __name__ == "__main__":
     if 'snakemake' not in globals():
         from _helpers import mock_snakemake
+        os.chdir(os.path.dirname(os.path.abspath(__file__)))
```
pz-max, line 124: required this for local debugging in VScode

scripts/add_electricity.py
```diff
@@ -328,6 +329,10 @@ def attach_hydro(n, costs, ppl, profile_hydro, hydro_capacities, carriers, **con
 
     ppl = ppl.query('carrier == "hydro"').reset_index(drop=True)\
              .rename(index=lambda s: str(s) + ' hydro')
+    if len(ppl.loc[ppl.bus.isna()])>0:
+        # TODO: Why some hydro plants don't have a bus?
+        logger.warning(f"Remove hydro powerplants with no bus: {ppl.loc[ppl.bus.isna()].index}")
+        ppl = ppl.loc[~ppl.bus.isna()]
```
pz-max, line 335: @davide-f maybe this fixes your bug in  https://github.com/pypsa-meets-africa/pypsa-africa/pull/376#issuecomment-1163729951

pz-max:
@p-glaum do you want to review it before @fneum ?
Happy to go through the PR with you and also discuss that stuff over a discord session

scripts/add_electricity.py
davide-f: I'll look into it

p-glaum:
Hi @pz-max, yes sure, I can have a look at it.

FabianHofmann:
Hey max, allow me to be picky as I think the changes are a bit bloated. I'd rather go for another approach here: 

1. Define a `regions_geojson_schema` in the `_helpers.py`. This will be used in the `gpd.GeoDataFrame.to_file`  function wherever there is an export of onshore and offshore shapes. 
2. Replace all the `save_to_geojson` calls by the pattern `regions.to_file(<fn>, schema=regions_geojson_schema)`. This ensures that empty geodataframes can be exported and afterwards imported.
3. Update the cluster_regions function to also keep the whole set of columns that we have for the initial regions geodataframe. 

I just played around a little bit with the functions. I'd set up the alternative approach and then we decide.

fneum:
Thanks @pz-max for your initiative to tackle this long-standing issue!

PR #383, opened 2022-06-23
Landlock country fix alternative

### LINKED ISSUE #265: Support selection of only landlocked countries

FabianHofmann:
Closes #382 
closes #265

## Checklist


pz-max:
Looks much more compact. Great! :)
Hope the CI runs through.
Testing it on:
- ["AT"]  # landlock
- ["AT", "CZ"]  # multiple countries

FabianHofmann:
My internet in the train is terrible. @pz-max could you do me the favour and try your config settings with the up to date version?

pz-max:
> My internet in the train is terrible. @pz-max could you do me the favour and try your config settings with the up to date version?

@FabianHofmann I am now in the bus for 18h with okish internet connection. I tried to run the branch for both cases ['AT'] and ['AT','CZ']. The following error appeared:

> File "/home/max/OneDrive/PHD-Flexibility/07_pypsa-africa/0github/pypsa-africa/pypsa-eur-test4/pypsa-eur/.snakemake/scripts/tmp1hvdqnee.base_network.py", line 581, in base_network
> _set_countries_and_substations(n, config, country_shapes, offshore_shapes)
> File "/home/max/OneDrive/PHD-Flexibility/07_pypsa-africa/0github/pypsa-africa/pypsa-eur-test4/pypsa-eur/.snakemake/scripts/tmp1hvdqnee.base_network.py", line 398, in _set_countries_and_substations
> offshore_shapes = gpd.read_file(offshore_shapes).set_index('name')['geometry']
> File "/home/max/anaconda3/envs/pypsa-eur-landlock/lib/python3.9/site-packages/pandas/util/_decorators.py", line 311, in wrapper
> return func(*args, **kwargs)
> File "/home/max/anaconda3/envs/pypsa-eur-landlock/lib/python3.9/site-packages/pandas/core/frame.py", line 5494, in set_index
> raise KeyError(f"None of {missing} are in the columns")
> **KeyError: "None of ['name'] are in the columns"**

I see one problem with the here suggested solution. I think we cannot just provide an empty dataframe that solves everything because we use the `offshore_shapes` and later the `offshore_regions` to execute functions. If data such as columns or the values in the dataframes are missing errors might occur as above.

I think to avoid issues with the import:
- We need as in the previous drafted PR file size check & and write out of empty files for the offshore stuff (not to break the workflow)

To avoid any operational errors, e.g. when geopandas function are intended to be executed on the empty file:
- We might need to add the case distinctions as previously suggested.
 
In future, it might make also sense to distentangle any onshore & offshore functions e.g. within the script or even with the rules (build_shapes -> build_offshore_shape + build_onshore_shape).... Atm some of them are mixxed up. But this is probably only a nice to have.

FabianHofmann:
@pz-max it appears that the new geopandas version supports writing out empty geojsons! The only downside: the reimported empty dataframe has lost all columns but the `geometry` column. Forcing a reindex after the imports of offshore shapes makes everything happen.

It runs through for me now. However, one has to disable offshore tech if no offshore shapes are existent. Otherwise the workflow will raise an error. 

I prefer this implementation over the other approach as it is cleaner. Especially, getting rid of the `save_to_geojson` function is a point I would like to have. Let me know if you have second thoughts.

scripts/_helpers.py
```diff
@@ -4,7 +4,9 @@
 
 import pandas as pd
 from pathlib import Path
+from collections import OrderedDict
```
fneum, line 7: Is this necessary?

fneum (approved):
This solution looks good!

PR #384, opened 2022-06-24
Enable parallel computing with new dask version

FabianHofmann:
Apparently, in the new dask version the argument `num_workers` is ignored and falls back to 1. This extremely slows down the whole workflow. For the moment I cannot come up with something better than this. Upside: Every thing is parallel and fast again, downside: One does not see the progressbar anymore (I cannot say why...) 

## Checklist

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -19,6 +19,7 @@ gurobi.log
 /data
 /data/links_p_nom.csv
 /cutouts
+/dask-worker-space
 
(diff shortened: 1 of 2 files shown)


fneum (approved):
I ran the build_renewable_profiles yesterday and I can say a small loss in convenience is acceptable for restored parallel computing. Thanks @FabianHofmann!

PR #386, opened 2022-06-27
Revert "remove build/retrieve natura raster, directly use shapefile"

fneum:
Reverts PyPSA/pypsa-eur#336


euronion:
Out of curiosity, what was the reason for reverting?

FabianHofmann:
It took very long...

PR #387, opened 2022-06-27
remove build/retrieve natura raster, directly use shapefile (2nd try)

fneum:
Reverts PyPSA/pypsa-eur#386


FabianHofmann:
I had a try with the branch, I still think the increase of computation time is still to high. Tested with default config, the machine broke as soon as the parallel processes begin. Loading the natura vectors into memory also takes a while (~20 sec). 
Perhaps we should rather start to modularize the Snakefile instead?

PR #388, opened 2022-06-28
generalize conventional attr handling through config

FabianHofmann:

Generalize conventional attribute assignment through config. I think this a more intuitive and abstract handling of adding attrs through the config (EAFs is a quite special case which would now be generalized)

Edit: Like this the user can flexibly add ramp_limits, p_min_pu values or modify existent emission factors. And snakemake requires the input files.

@fneum @martacki let me know what yo think.


scripts/add_electricity.py
```diff
@@ -331,17 +331,20 @@ def attach_conventional_generators(n, costs, ppl, conventional_carriers, extenda
...
+        for key in list(set(conventional_carriers[carrier]) & set(n.generators)):
+
+            values = conventional_config[carrier][key]
+
+            if isinstance(values, str) and str(values).startswith("data/"):
+                # Values affecting generators of technology k country-specific
+                # First map generator buses to countries; then map countries to p_max_pu
+                values = pd.read_csv(values, index_col=0).iloc[:, 0]
+                bus_values = n.buses.country.map(values)
+                n.generators[key].update(n.generators.loc[idx].bus.map(bus_values).dropna())
```
fneum, line 344: Optional: You could use `snakemake.input` here. E.g. look for `snakemake.input.conventional_nuclear_p_max_pu`, something like:

```py
fn = snakemake.input.get(f"conventional_{key}_{values}")
if fn is not None:
    pd.read_csv(fn, ....)
    ....
```

fneum (changes requested):
I very much like the idea! The file management looks a bit clunky, but I like that it is so general and I think it's fine.

Snakefile
```diff
@@ -217,7 +217,8 @@ rule add_electricity:
         load='resources/load.csv',
         nuts3_shapes='resources/nuts3_shapes.geojson',
         **{f"profile_{tech}": f"resources/profile_{tech}.nc"
-           for tech in config['renewable']}
+           for tech in config['renewable']},
+        **{"conventional_{carrier}_{attrs}": fn for carrier in config.get('conventional', {None: {}}).values() for fn in carrier.values() if str(fn).startswith("data/")},
```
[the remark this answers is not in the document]
fneum: Are you sure? These are not wildcards.

scripts/add_electricity.py
```diff
@@ -331,17 +331,20 @@ def attach_conventional_generators(n, costs, ppl, conventional_carriers, extenda
...
-            # Single value affecting all generators of technology k indiscriminantely of country
-            n.generators.loc[idx, "p_max_pu"] = factors
-        elif isinstance(factors, str):
-            factors = pd.read_csv(factors, index_col=0)['factor']
-            # Values affecting generators of technology k country-specific
-            # First map generator buses to countries; then map countries to p_max_pu
-            bus_factors = n.buses.country.map(factors)
-            n.generators.p_max_pu.update(n.generators.loc[idx].bus.map(bus_factors).dropna())
+
+        for key in list(set(conventional_carriers[carrier]) & set(n.generators)):
```
martacki, line 335: does that work? `conventional_carriers` is a list of carriers, `carrier` is a `str`, so what is `conventional_carriers[carrier]`?

martacki: also, n.generators is a `pd.Dataframe`, maybe you meant `n.generators.carrier`?

why not toe following? (I'm not sure what you're trying to filter)
```suggestion
        n.generators.loc[idx].carrier.unique():
```

martacki: ahm, you probably wanted to check if the key given in config is in generator columns?
```suggestion
        for key in list(set(conventional_config[carrier]) & set(n.generators)):
```

scripts/add_electricity.py
martacki: prefer original code because we dislike `snakemake.input` within functions.

Snakefile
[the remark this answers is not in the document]
martacki: Worked for me at least, don't know about windows/...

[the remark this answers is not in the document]
FabianHofmann: would work without the f string, but the `attrs` getter is important

scripts/add_electricity.py
FabianHofmann: you are just right

martacki:
> I very much like the idea! The file management looks a bit clunky, but I like that it is so general and I think it's fine.

I like it to be more general, too!

PR #417, opened 2022-09-16
lift rasterio restriction to allow atlite 0.2.9

fneum:
atlite 0.2.9 requires minimum rasterio 1.3, so with current environment restrictions never installed

rasterio 1.2.10 should still be forbidden, in case earlier atlite versions are installed


FabianHofmann (approved):

FabianHofmann:
you have a clue why windows CI does not want to do its job?

fneum:
I think it's because I lifted the ipopt version restriction, where Win has problems with finding executables of newer ipopt versions.

On the other hand Ubuntu had troubles now with the older ipopt versions.

So, I think the solution would be to make the appendix to the `environment.yaml` in the Github Action dependent on `matrix.os`. What do you think?

PR #418, opened 2022-09-19
Config: enable res installations per default

FabianHofmann:
This is helpful for other projects (as BE) to enable short-term scenarios. 

## Checklist


p-glaum (commented):
Looks fine to me!

p-glaum (approved):

PR #442, opened 2022-11-15
config: replace `closed` with `inclusive` in snapshot creation

FabianHofmann:
Fix a pandas deprecation


fneum:
see #323, forces pandas>=1.4, sure we want to merge yet?

FabianHofmann:
Ah sorry, forgot about that PR and the reason why we didn't merge it so far. I'll remove this PR and merge #323 when we think the time is ready.

PR #451, opened 2022-12-16
Fix load variable in reserve constraint

pz-max:
There is a reserve constraint set which considers the total demand per snapshot.
The total demand per snapshot was assigned as: `demand = n.loads_t.p.sum(1)`
However, this will always result in `0`

My suggestions would be: `demand = n.loads_t.p_set.sum(1)`

## Checklist

- [x] A note for the release notes `doc/release_notes.rst` is amended in the format of the previous release notes.


fneum:
Good catch !

PR #452, opened 2022-12-16
Fix EQ constraint in case no hydro exist

pz-max:
The PyPSA-Eur tutorial will fail if someone tests the EQ constraint.
The bug appears because no spillage variable exists if there is no inflow.
Since the tutorial for 'BE' has no 'hydro' but 'DE' has 'hydro', the current bug was not spotted before.

The solution is to check if there is an inflow and then execute the spillage-related code.

## Checklist


pz-max:
Ready for review and merge.
Change is only an if / else addition

fneum (approved):

PR #467, opened 2023-02-08
env: add pygeos

FabianHofmann:


fneum:
This currently causes problems. With pigeons installed, some rules mysteriously don’t run through.

PR #468, opened 2023-02-08
Revert "env: add pygeos"

fneum:
Reverts PyPSA/pypsa-eur#467


FabianHofmann:
oh okay, in pypsa-eur or / sec?

fneum:
I think sec if I remember correctly

FabianHofmann:
It worked for me locally. But one of pygoes or shapely >= 2.0 is actually needed in pypsa-eur-sec. So alternative is upgrading shapely.

PR #469, opened 2023-02-08
env: update shapely to >=2.0

FabianHofmann:


FabianHofmann:
See also https://github.com/pypsa-meets-earth/pypsa-earth/pull/578. Some unexpected warnings are given out in the intersection calculation, but @davide-f this seems to be harmless right?

PR #474, opened 2023-02-15
importing rioxarray instead of xarray to open rasterio file

virio-andreyana:
closes #644
closes #476 

Importing rioxarray instead of xarray to open rasterio file. This is to fix the warning in rule build_ship_raster: open_rasterio is Deprecated in favor of rioxarray. For information about transitioning, see: https://corteva.github.io/rioxarray/stable/getting_started/getting_started.html 

changes had been made in the environment and environment.fixed to add rioxarray

## Checklist

- [x] I tested my contribution locally and it seems to work fine.
- [ ] Code and workflow changes are sufficiently documented.
- [x] Newly introduced dependencies are added to `envs/environment.yaml` and `envs/environment.docs.yaml`.
- [ ] Changes in configuration options are added in all of `config.default.yaml`, `config.tutorial.yaml`, and `test/config.test1.yaml`.
- [ ] Changes in configuration options are also documented in `doc/configtables/*.csv` and line references are adjusted in `doc/configuration.rst` and `doc/tutorial.rst`.
- [ ] A note for the release notes `doc/release_notes.rst` is amended in the format of previous release notes.

PR #476, opened 2023-02-16
use ship raster tif file directly instead of converted nc file

FabianHofmann:
closes #474 

## Checklist


fneum:
The build shapes script is no longer necessary, is it?

Has this been tested for functionality and performance degression already?

fneum:
Hmm, I think we shouldn't skip cutting to the geographical scope of the model. The extracted `.tif` is 10 GB large. Also RAM requirements are higher for me.

FabianHofmann:
Did you see large differences? Performance wise it was just a couple of seconds for me and memory use was also not high. Alternatively we could clip the tif file before writing to file.

fneum:
I’d definitely clip it. I am more worried about the 10 GB of disk space than little higher memory usage.

PR #480, opened 2023-02-22
ci: reduce duplications of test/config.test1.yaml

### LINKED ISSUE #432: New test pipeline

Problem: Currently, we need to update the `test/config.test1.yaml` which is basically a special version of the `config.tutorial.yaml`. Further, adding new tests is quite annoying since there will be a lot of code duplication in the config.

Solution: Instead of duplicating the tutorial version for each test, we could build it from the `config.tutorial.yaml`.

This enables:
- Creating multiple tests/ more test coverage of functionalities because the test config is super small
- One disadvantage is that multiple test would require more time in the CI

Check out the recent PyPSA-Earth PR to see what is required for the new test pipeline: https://github.com/pypsa-meets-earth/pypsa-earth/pull/466 We only test the full option set for Linux to avoid long CI times.

fneum:
Closes #432

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -95,6 +95,4 @@ jobs:
       run: |
         conda activate pypsa-eur
         conda list
-        cp test/config.test1.yaml config.yaml
-        snakemake --cores all solve_all_networks
-        rm -rf resources/*.nc resources/*.geojson resources/*.h5 networks results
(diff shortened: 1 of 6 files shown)


fneum:
@pz-max will like this.

pz-max:
💯😃 next week at TUB, remind me that I want to show smt in context of the scenario management related to this

PR #484, opened 2023-03-06
Merge pypsa eur sec

### LINKED ISSUE #450: Consider using `conda: ` directives in snakemake rules
## `Conda:` directives in snakemake rules.

I would propose a minor enhancement to the Snakefile, which is to add `conda:` directives to each rule; see https://snakemake.readthedocs.io/en/stable/snakefiles/deployment.html#integrated-package-management. The implication for running pypsa-eur is that you would no longer have to activate the pypsa-eur conda environment before calling snakemake, but instead call snakemake with the `--use-conda` argument.

Pros:
- The potential for using the wrong conda environment is much reduced.
- This makes it trivial to manage several parallel pypsa-eur installation with potentially different conda environments (different versions, different projects, development, etc.)
- Snakemake automatically takes care of rebuilding the environment when you change the specification.
- This makes it possible to use pypsa-eur as a snakemake module and use different conda environments in the "parent" workflow.
- As far as I'm aware, if you don't supply the `--use-conda` argument, snakemake will run the workflow as usual, meaning that this change would be fully backwards compatible.

Cons:
- Yet another keyword in every single snakemake rule.
- Makes it slightly more cumbersome to add additional dependencies to the environment; they must now be specified in the "environment.yaml" instead of installing them "on the go" in the live environment. (Although it can be argued that forcing a more declarative style like this has advantages too, since it makes it more obvious what the environment looks like.)
- You might _still_ have to build the pypsa-eur environment "manually" for use in notebooks, etc., and may still need to keep _this_ environment up to date manually.

Other implications:
- By default, recent versions of snakemake will re-run any rules using conda environment "x" after the specification for "x" was changed. This makes a lot of sense from a reproducibility point of view, since the results of a rule could change when the software environment changes. From a practical point of view, however, it can be annoying if you for example add a dependency to the environment and this causes the whole workflow to re-run. Of course, this can be circumvented by a `snakemake [...] --touch [...]` stategy; it might be an idea to write about that in the pypsa-eur documentation if this change is adopted.

And example of what this would look like can be found here: https://github.com/koen-vg/pypsa-eur/blob/64aed45cd3d94ab48cdaed0cc02c16b6aec7dfc5/Snakefile

Overall I think this would be a good step in the direction of making pypsa-eur easier to work with in a fully reproducible way and makes it harder to make mistakes with the conda environment.

Of people are in favour of this change, I would of course be happy to open a pull request.

### LINKED ISSUE #463: Merge PyPSA-Eur-Sec into PyPSA-Eur
We plan to merge [PyPSA-Eur-Sec](https://github.com/PyPSA/pypsa-eur-sec) into [PyPSA-Eur](https://github.com/PyPSA/pypsa-eur) in March 2023. The PyPSA-Eur-Sec repository would then be deprecated. PyPSA-Eur would include all features from PyPSA-Eur-Sec (additional energy sectors as well as myopic and perfect foresight).

## Reasons for merger
- Future energy systems will have enough sector-coupling that modelling electricity separately will be hard to justify
- The workflow is currently very messy, with PyPSA-Eur-Sec removing items (like storage and conventional generators) from PyPSA-Eur and then reading them in in a different way
- PyPSA-Eur-Sec has features that PyPSA-Eur can benefit from (myopic and perfect foresight, new [scenario management](https://github.com/PyPSA/pypsa-eur-sec/pull/250), existing wind and solar via IRENA database)
- There are confusing duplications between config files
- It is difficult managing the different library dependencies between the two repositories
- Snakemake sub-workflows are [deprecated](https://snakemake.readthedocs.io/en/stable/snakefiles/modularization.html) (to be replaced by modules) and anyway hard to work with
- Merging avoids duplication of ancillary scripts (solving, summarising, plotting, scenario management)

## Reasons against the merger
- PyPSA-Eur-Sec is more complicated, with more complicated data and library dependencies than PyPSA-Eur, which may put off newbies and casual users; we can mitigate this difficulty by turning off non-electricity sectors by default

## Planned steps
- introduce basic CI testing to both PyPSA-Eur and PyPSA-Eur-Sec so we can detect problems (could test inputs but also outputs e.g. for load shedding, energy mix from temporally clustered results, prices, etc.)
- pull perfect foresight code into PyPSA-Eur-Sec (@lisazeyen will do this in February 2023)
- merge the PyPSA-Eur and PyPSA-Eur-Sec snakemake workflows and data with minimal additional changes (perhaps as an initial test to see what goes wrong)
- slowly iron out the discrepancies (adding/removing of components before/after clustering, etc.)
- test it rigorously
- document everything

We estimate it will take 2-3 weeks and plan it for March 2023.

## Potential difficulties
- PyPSA-Eur adds components at high resolution (renewables, conventional power plants, load), then clusters, whereas PyPSA-Eur-Sec removes some of these and then adds back components after the clustering; this should be harmonised so that only components that need high resolution are added before clustering (probably renewables, conventional power plants and load are needed, since they are used for clustering in some methods), then rest is added afterwards (e.g. storage that doesn't depend on high resolution and is not used for clustering)
- PyPSA-Eur uses PyPSA's generator component for conventional power plants, whereas PyPSA-Eur-Sec uses links (since need to track fuel that can be fossil or green-hydrogen-based, as well as for CO2 tracking); tentative suggestion: only use links in future
- PyPSA-Eur-Sec has IRENA database for existing wind and solar, whereas PyPSA-Eur has its own methods
- The myopic and perfect foresight in PyPSA-Eur-Sec needs to be "back-ported" to work with electricity-only PyPSA-Eur
- The deduction of heating and industry electricity demand from total electricity demand in PyPSA-Eur-Sec needs to be harmonised with PyPSA-Eur (should be fine since the deduction should only happen if these sectors are chosen in the config)
- PyPSA-Eur-Sec handles CO2 with links and stores (because of CCU/S) whereas PyPSA-Eur uses a global constraint on CO2 (suggestion: use PyPSA-Eur-Sec format)
- The new [scenario management](https://github.com/PyPSA/pypsa-eur-sec/pull/250) needs to work across both
- We need a clever design that simplifies everything by default for new users (e.g. defaults to electricity only and overnight)
- We need to provide a migration path for existing users

### LINKED ISSUE #612: CI: Report summary
We should consider provide a GH artefact from the CI. This could either be in the form of

- csv and pdf files from `make_summary` / `plot_summary` in a zip file (not so fancy) 
- a snakemake report (fancy) 
- a streamlit html file (super fancy)

### LINKED ISSUE #613: Disable progressbar in tqdm for CI runs
Following this [merged PR](https://github.com/PyPSA/pypsa-eur/pull/478), it is useful to disable the progressbar for the CI runs as it looks bad when analysing the CI. @davide-f  shows in [PyPSA-Earth a solution](https://github.com/pypsa-meets-earth/pypsa-earth/blob/5da36819ba1873ebbf3d1f40558137fafd2b24ed/scripts/_helpers.py#L314) which could be tested in PyPSA-Eur.

fneum:
Closes #463
Closes #450 
Closes #613 
Closes #612

## Checklist

- [x] migrate PRs from pypsa-eur-sec
- [x] harmonize solve_network @FabianHofmann 
- [x] CI
- [x] modularization of Snakefile
- [x] enable/disable progressbar 
- [x] documentation with instructions (how to run elec only etc.) @fneum 
- [x] release notes @fneum


fneum:
In my view, this is ready to be merged.

lisazeyen:
looks really great @fneum, @FabianHofmann . I had a look at the documentation and tried one electricity only, one sector-coupled and one myopic run, all worked nice and smoothly!

FabianHofmann:
According to our todo list, the only thing missing is the final releases of pypsa-eur(-sec) before merging. However looking at the changes in pypsa-eur since the release, none of it is functional (version fixes, tqdm, cluster wildcard). So I don't even know whether this is necessary. So, I am happy with merging as is anytime :)

fneum:
Alright, I agree that we don't need new independent releases before merging because 0.7.0 was very recent and because no functional changes were made since then.

Ok, since @lisazeyen, @FabianHofmann, @fneum are happy, and @nworbmot reviewed the docs, let's move ahead!

FabianHofmann:
Whoop!!

Issue #57, opened 2019-08-07
Include option to include planned lines from ENTSO-E's TYNDP 


nworbmot:
The option to include the planned HVDC lines from the TYNDP was already implemented. They are in a CSV file which can be included using an optional flag. Lines can be added, and existing lines can be replaced. Lines come with a start and finish lat/lon and are attached to the closest buses.
This should be extended also to planned HVAC lines from the TYNDP, along with expected commissioning years. Extracts from the ENTSO-E map may include some of this information already, but it's not clear how complete the map is compared to the TYNDP.
FIAS colleagues have already got a list of NEP (German TYNDP) projects for Germany.

fneum:
closed via #1085

PR #626, opened 2023-03-12
Integrate tyndp 2020

grecht:
Related to #57 .

- Added the option to include projects from the TYNDP 2020.
- Commissioning dates are included. Every non-TYNDP asset gets assigned the commissioning date 1970-01-01.
- The new configuration options include a possibility to set the desired project status and ignore selected projects.

In an external repository ["tyndp_to_pypsa"](https://github.com/grecht/tyndp_to_pypsa), coordinates were added to the TYNDP projects by hand based on asset names, the rudimentary maps on the TYNDP website, and OpenStreetMap. The data is converted there to the gridextract format so that it can be imported seamlessly into PyPSA-Eur. It is already split into asset type and "new" or "upgraded" based on the latest gridextract and `links_tyndp.csv` - but this means every time this data changes, the TYNDP files in `data/tyndp2020` would have to be rebuilt using `tyndp_to_pypsa`.

To test this, follow the usual config file workflow (remember to set `include: true` in the TYNDP2020 options).
I have not tested this beyond building the base network (see plot at bottom).

## Checklist

- [x] Check how capacities overall have changed as @martacki did [here](https://github.com/PyPSA/pypsa-eur/pull/308)
- [ ] The TYNDP project status is thrown away, and instead  `under_construction=True` is set in order to conform to the current format. Does this make sense?
- [ ] Is the way asset updates are applied problematic? In the current state when e.g. a line's capacity is upgraded, its voltage is updated but as it is a TYNDP project, `under_construction=True` is set. I do not know the implications of this, but the updated asset might not remain usable until the commissioning date of the update.
- [x] It is unclear how to represent consecutive upgrades of the same asset: Right now, the "best" upgrade is taken, but this is also the latest with a more uncertain status. Probably does not occur often, but I did not check.

base_network


config/config.default.yaml
```diff
@@ -257,6 +257,13 @@ conventional:
   nuclear:
     p_max_pu: "data/nuclear_p_max_pu.csv" # float of file name
 
+TYNDP2020:
+  include: false
+  allowed_statuses: ['in_permitting', 'under_construction']  # 'planned_not_yet_permitting', 'under_consideration'
```
pz-max, line 262: maybe you have to add these lines also in the test.
CI is failing now because some indexing or similar pandas error

pz-max, line 262: better -> `allowed_status`

scripts/base_network.py
```diff
@@ -680,6 +688,118 @@ def _adjust_capacities_of_under_construction_branches(n, config):
...
+    lines,
+    links,
+    new_buses,
+    new_lines,
+    new_links,
+    upg_buses,
+    upg_lines,
+    upg_links,
+    config,
+):
```
pz-max, line 702: Maybe add here the README tyndp as docstring

pz-max (changes requested):
CI is failing at the moment. Probably the test config needs adjustments.
Please test locally and add suggested fix.

config/config.default.yaml
martacki: no thats not the problem. the problam is somehow related to the new pandas version and trying to add data to an empty dataframe

scripts/base_network.py
```diff
@@ -680,6 +688,124 @@ def _adjust_capacities_of_under_construction_branches(n, config):
...
+    df = df.loc[df["tyndp_status"].isin(allowed_statuses)]
+    if ignore_projects:
+        df["project_id"] = _get_tyndp_project_id(df)
+        df = df.loc[~df.loc[:, "project_id"].isin(ignore_projects)]
+        df = df.drop("project_id", axis=1)
+
+    df = pd.concat(
+        [df, pd.DataFrame(index=df.index, columns=["under_construction"], data=True)],
+        axis=1,
+    )
```
FabianHofmann, line 790:
```suggestion
    df = df.assign(under_construction=True)
```
make hacky non-hacky

martacki:
Quick update (probably mostly for myself):
I tried running the whole workflow now, but resulting in non-intuitive behaviour, e.g. network can now be clustered to 42 nodes (not 37), where Germany and France now have 3 nodes (instead of 1), despite all nodes belonging to the same synchronous zone, which is strange.
norway `NO` now has 4 synchronous zones (instead of 1), and some additional buses that are not being aggregated but remain disconnected from the network (at a 42 node resolution), similarly for Sweden `SE` (2 synchronous zones now). Regardless, the aggregation connects all nodes and puts the extra 4 nodes to Germany `DE` and France `FR`, from my understanding.
The last additional nodes goes to Corsica/Sardinia, which from the algorithm perspective works as intended (because there are no links to be simplified). However, I believe that there is some errors with the new TYNDP links connecting the isles. In this case, the TYNDP states that the existing link is being upgraded, but there are 2 new links in this PR, while the existing ones are all being kept.
new links: `TYNDP2020_29` + `TYNDP2020_30`

p-glaum:
There was one mistake in the line filtering which kicked out many tyndp lines in the base_network script, which @martacki is fixing.
Below, we compare the pypsa-eur grid with tyndp 2020 to the project map from tyndp. There are still lines missing when comparing to tyndp 2020, but most of them are considered in this PR. 
Next to ToDo is to rerun the pypsa-eur workflow and do the plots from above again. 
map_plot
https://tyndp2020-project-platform.azurewebsites.net/projectsheets

fneum:
@martacki @p-glaum, what was the latest status for this PR? Do you think this is ready to be merged?

p-glaum:
not yet, still need to do some runs on the cluster to validate this PR

p-glaum:
Hi @grecht,
could you maybe also give me contributor rights that I can make adjustments to your code? Then I do not always have to ask @martacki.

PR #637, opened 2023-04-06
Multi horizon

lisazeyen:
Closes # (if applicable).

This PR should include perfect foresight for pathway optimisation.

Open ToDos:

- [x] add extra_functionalities for perfect foresight
- [ ] min_part_load for methanolisation for shipping and Fischer-Tropsch leads to infeasibilities
- [x] must run condition for existing gas boilers to avoid massive back up energy, add option to retrofit them to h2 boilers 
- [x] transmission expansion
- [x] carbon budget constraint
- [x] option for max growth for certain carriers

## Checklist


lisazeyen:
@FabianHofmann Some points still can be improved for the multi-decade optimisation, but I would be happy to pull a first beta version into the master branch and then work on the single features in separate smaller PR.

lisazeyen:
@FabianHofmann CI should run through once https://github.com/PyPSA/PyPSA/pull/721 is in. There is a problem if the optimisation is multi-decade and the list of extendable lines or DC links is empty when selecting the active assets of an empty list.

lisazeyen:
@FabianHofmann, what do you think, can we merge?

scripts/plot_summary.py
```diff
@@ -379,18 +385,24 @@ def historical_emissions(countries):
     e["waste management"] = "5 - Waste management"
     e["other"] = "6 - Other Sector"
     e["indirect"] = "ind_CO2 - Indirect CO2"
-    e["total wL"] = "Total (with LULUCF)"
-    e["total woL"] = "Total (without LULUCF)"
+    # e["total wL"] = "Total (with LULUCF)"
+    # e["total woL"] = "Total (without LULUCF)"
```
FabianHofmann, line 389: this is intended?

scripts/solve_network.py
```diff
@@ -47,6 +49,69 @@ def add_land_use_constraint(n, planning_horizons, config):
...
+    check_p_min_p_max(p_nom_max)
+    # drop multi entries in case p_nom_max stays constant in different periods
+    # p_nom_max = compress_series(p_nom_max)
+    # adjust name to fit syntax of nominal constraint per bus
+    df = p_nom_max.reset_index()
+    df["name"] = df.apply(
+        lambda row: f"nom_max_{row['carrier']}"
+        + (f"_{row['build_year']}" if row["build_year"] is not None else ""),
+        axis=1,
+    )
```
FabianHofmann, line 105: Why did you use the `n.buses["nom_max_{carrier}"]` scheme and not the `n.global_constraints`? I thought the latter is more flexible with the build_years.

scripts/solve_network.py
```diff
@@ -130,16 +194,146 @@ def add_co2_sequestration_limit(n, limit=200):
         limit = float(o[o.find("seq") + 3 :]) * 1e6
         break
 
-    n.add(
+    if config["foresight"] == "perfect":
```
FabianHofmann, line 197:
```suggestion
    if not n.investment_periods.empty:
```
This pattern is more object related. Could perhaps be used elsewhere. Then the `config` argument can also be removed.

scripts/solve_network.py
```diff
@@ -130,16 +194,146 @@ def add_co2_sequestration_limit(n, limit=200):
...
+        names,
         sense="<=",
         constant=limit,
         type="primary_energy",
         carrier_attribute="co2_absorptions",
+        investment_period=periods,
     )
 
 
+def add_carbon_constraint(n, snapshots):
```
FabianHofmann, line 215: Could you shortly point out why this one (and the one below) is necessary? It feels like this part of code should rather go into pypsa itself, but not sure...

config/config.perfect.yaml
```diff
@@ -0,0 +1,991 @@
+# SPDX-FileCopyrightText: : 2017-2023 The PyPSA-Eur Authors
```
FabianHofmann, line 1: I think we agreed on making this containing relative changes only, or? In VScode it is easy to see the relative change with the `Diff: File` command.

FabianHofmann (commented):
I would be much in favour to merge it. Here some last remaining comments and the CI should run through. Otherwise, it looks great!

scripts/plot_summary.py
lisazeyen: we can remove it for now. They changed the format in a newer version of the input data and then these two will be needed but I can remove it for now. We can update the data (which will effect also other scripts) in a separate PR.

scripts/solve_network.py
lisazeyen: Mhm, I think one of the main reasons I implemented it directly in PyPSA-Eur is that with this implementation it is solving faster because the constraint is just on the CO2 atmosphere store and not on all components which have emissions. But this is maybe rather PyPSA-Eur specific since I guess many people do not model carbon management.

scripts/solve_network.py
lisazeyen: I think it can be formulated in that way with a [max_relative_growth](https://github.com/PyPSA/PyPSA/blob/4514190afca73c8385b8fc50bbd9ad129b7d1c3c/pypsa/optimization/global_constraints.py#L170) so that the growth rate does not depend on an absolute number but on the installations in the previous period. We could think about adding an additional global constraint type for absolute growth numbers per investment period but I think we don't have this yet right?

lisazeyen:
@FabianHofmann thanks for your feedack. CI is running through with the latest PyPSA Version, only the building of the documentation fails.  I addressed your points above

PR #645, opened 2023-04-20
fix bugs in reserve constraint

lisazeyen:
Closes # (if applicable).

## Fix bugs in reserve constraint

This PR should fix to bugs in the operational reserve constraint.
1. bug: when formulating the lhs of the constraint it is summed over the time steps not the generators
2. bug: for the additional constraint that reserve + dispatch < capacity the sign is the wrong way around

This PR does **not** fix the issue that this constraint only works for the power sector.
## Checklist

### DIFF
--- a/scripts/solve_network.py
+++ b/scripts/solve_network.py
@@ -43,6 +43,7 @@ from vresutils.benchmark import memory_logger
 
 logger = logging.getLogger(__name__)
 pypsa.pf.logger.setLevel(logging.WARNING)
+from pypsa.descriptors import get_switchable_as_dense as get_as_dense
 
 
 def add_land_use_constraint(n, config):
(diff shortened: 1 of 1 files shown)


scripts/solve_network.py
```diff
@@ -441,17 +444,26 @@ def add_operational_reserve_margin(n, sns, config):
...
+
+    lhs = dispatch + reserve - capacity_variable * p_max_pu[ext_i]
+
+    rhs = (p_max_pu[fix_i] * capacity_fixed).reindex(columns=gen_i, fill_value=0)
 
-    lhs = n.model.constraints["Generator-ext-p-upper"].lhs
-    lhs = lhs + reserve.loc[:, lhs.coords["Generator-ext"]].drop("Generator")
-    rhs = n.model.constraints["Generator-ext-p-upper"].rhs
-    n.model.add_constraints(lhs >= rhs, name="Generator-ext-p-upper-reserve")
+    n.model.add_constraints(lhs <= rhs, name="updated_capacity_constraint")
```
FabianHofmann, line 466:
```suggestion
    n.model.add_constraints(lhs <= rhs, name="Generator-p-reserve-upper")
```

FabianHofmann (changes requested):
Looks good! I would just suggest renaming the constraint.

FabianHofmann (approved):

PR #647, opened 2023-04-22
add plain hydrogen turbine option for re-electrification

fneum:


scripts/prepare_sector_network.py
```diff
@@ -1067,18 +1067,37 @@ def add_storage_and_grids(n, costs):
...
+        logger.info("Adding hydrogen turbine for re-electrification.")
+
+        n.madd(
+            "Link",
+            nodes + " H2 turbine",
+            bus0=nodes + " H2",
+            bus1=nodes,
+            p_nom_extendable=True,
+            carrier="H2 turbine",
+            efficiency=costs.at["OCGT", "efficiency"],
```
FabianHofmann, line 1096: Intended that the parameters are taken from `OCGT`?

FabianHofmann: Can we add a comment that briefly points that out?

PR #659, opened 2023-05-08
Adapt equity constraint to sector-coupled networks

koen-vg:
This pull requests updates the equity constraint to include all energy carriers in sector-coupled models. For details on how this is done and which modelling and implementation choices were made, see the docstring and comments in the relevant function.

For example, here are two networks solved with a 30% and 95% country-level equity constraint, respectively:
Skjermbilde fra 2023-05-08 17-34-05
Skjermbilde fra 2023-05-08 17-34-09
Note the significant shift of solar investment to Germany, the Netherlands and Belgium in the higher equity network (the lower picture), as well as some smaller regional differences (e.g. Finland/Estonia). (However, these results were produced with a _weekly_ time resolution in order to test quickly, so don't take them too seriously!)

From my testing, the constraint starts to have a significant effect from an equity level of around 70-80%. All in all the results seem reasonable and I think having some level of equity can even be a good "default" or "baseline" for most modelling.

Unfortunately, since there are so many different components in the sector-coupled network, the implementation is on the long side, and I don't think it can be made much smaller. There is also potential for bugs or leaks where some new component is added in future versions and it's forgotten to take it into account in the equity constraint. Even for the current implementation, I tried quite hard to make sure that I caught everything but it would be great if something else could also have a look through it just to have a second pair of eyes.

Maybe it would also be an idea to add an automated test to make sure that this implementation at least keeps running? Checking that the constraint actually works is unfortunately quite a bit of work; about as much as writing the constraint in the first place in fact. But it would also be useful to write a post-processing step which calculates equity levels after the fact and checks that they are within bounds.

I think the only obvious problems with the current implementation are:
- Tested for electricity-only and sector coupled networks, but _only_ for overnight foresight. (But the original implementation was also only for overnight foresight.)
- Gas production (e.g. in the North Sea) is counted as imported since implementation-wise this is lumped together with pipeline and LNG imports. Too bad for Norway, the Netherlands and the UK: all their locally produced gas is not counted towards energy equity. Fixing this would require some changes "upstream" in the workflow.

I would document this better in the wildcards documentation, but whole sector-coupled network wildcards section of the documentation seems to be under construction. Anyway it's not so clear how to document options that can appear both in `{opts}` and `{sector_opts}` (such as EQ) but might have slightly different behaviour in the different cases.

## Checklist

### DIFF
--- a/doc/release_notes.rst
+++ b/doc/release_notes.rst
@@ -23,6 +23,8 @@ Upcoming Release
   hydrogen fuel cell. Add switches for both re-electrification options under
   ``sector: hydrogen_turbine:`` and ``sector: hydrogen_fuel_cell:``.
 
+* Adapt equity constraint option (EQ) to sector-coupled networks.
+
 PyPSA-Eur 0.8.0 (18th March 2023)
 =================================
 
--- a/scripts/solve_network.py
+++ b/scripts/solve_network.py
@@ -248,13 +248,66 @@ def add_CCL_constraints(n, config):
         )
 
 
-def add_EQ_constraints(n, o, scaling=1e-1):
+def add_EQ_constraints(n, level, by_country, config, scaling=1e-1):
     """
     Add equity constraints to the network.
 
-    Currently this is only implemented for the electricity sector only.
+    The equity option specifies a certain level x of equity (as in
+    EQx) where x is a number between 0 and 1. When this option is set,
+    each node in the network is required to produce at least a
+    fraction of x of its energy demand locally. For example, when
+    EQ0.7 is set, each node is required to produce at least 70% of its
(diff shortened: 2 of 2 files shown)


koen-vg:
Hi! Any updates on when this might be merged, or what might need to be done in order to get it merged? Do you need me to write tests first?

koen-vg:
Maybe after I resolve the merge conflicts, this could also be a candidate for v0.12, @fneum? I can do the necessary work to document this a little bit better and also add a configuration option to make this compatible with the new way of handling wildcards, scenarios etc. I still believe this is a valuable feature even if a bit complex; potential caveats will be documented.

PR #662, opened 2023-05-10
Remove vresutils

virio-andreyana:
Remove the depreciated vresutils from pypsa-eur.

Affected scripts
- add_electricity.py
- prepare_sector_network.py
- solve_network.py
- solve_operations_network.py

## Checklist

### DIFF
--- a/doc/release_notes.rst
+++ b/doc/release_notes.rst
@@ -23,6 +23,8 @@ Upcoming Release
   hydrogen fuel cell. Add switches for both re-electrification options under
   ``sector: hydrogen_turbine:`` and ``sector: hydrogen_fuel_cell:``.
 
+* Remove ``vresutils`` dependency.
+
 PyPSA-Eur 0.8.0 (18th March 2023)
 =================================
 
--- a/doc/requirements.txt
+++ b/doc/requirements.txt
@@ -9,7 +9,6 @@ sphinxcontrib-bibtex
 myst-parser  # recommark is deprecated, https://stackoverflow.com/a/71660856/13573820
 
 pypsa
-vresutils>=0.3.1
 powerplantmatching>=0.5.5
 atlite>=0.2.9
 dask[distributed]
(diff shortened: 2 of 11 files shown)


virio-andreyana:
The biggest challenge is finding a replacement for the memory_logger function in solve_network and solve_operations_network. The code from vresutils can be seen here https://github.com/FRESNA/vresutils/blob/master/vresutils/benchmark.py

fneum:
One could try a minimal memory_logger class that is put into `helpers.py`:

```py
import os
import time
import threading
import psutil

class MemoryLogger:
    def __init__(self, filename, interval):
        self.stop_event = threading.Event()
        self.thread = threading.Thread(target=self.start, args=(filename, interval))
        self.thread.start()

    def __exit__(self, *args):
        self.stop_event.set()
        self.thread.join()

    def start(self, filename, interval):
        with open(filename, 'w') as f:
            while not self.stop_event.is_set():
                total_memory = 0
                for proc in psutil.process_iter(['pid', 'memory_info']):
                    if proc.info['pid'] == os.getpid() or proc.parent() == os.getpid():
                        total_memory += proc.info['memory_info'].rss
                f.write(f"{time.time()} {total_memory / 10**6}\n")
                time.sleep(interval)
```

Requires `psutil`. Can be used:

```py
with MemoryLogger(filename='memory.log', interval=30.0):
    computationally_demanding_function()
```

FabianHofmann:
Or one could use the memory logger from snakemake (via benchmark). But in general, I think it would be good to allow to disable it, as memory logging can slow down processes significantly.

fneum:
Agree! The advantage of the vresutils memory logger is that it monitors and logs RAM usage continuously, but honestly I am not sure how important it is...

virio-andreyana:
Looking into snakemake versions of benchmark https://github.com/snakemake/snakemake/blob/main/snakemake/benchmark.py, I believe snakemake has already well documented its resident set size (RSS), virtual memory size and proportional set size (PSS) through its automatic benchmark log. Let's just remove vresutils.benchmark from the code without replacing it.

fneum:
I'm happy with this.

fneum:
Looks great! Finally `vresutils` is out! Just added a few finishing touches.

PR #694, opened 2023-07-13
Add test and option to disable online retrieve rules

euronion:

Minor improvement.

`snakemake` tries to access the internet for remote providers like `HTTP` while constructing the DAG. This prevents the execution of a workflow even when copies of all remote files are already locally available, as the DAG construction will simply fail.

This PR add's an option (`config[enable][retrieve] = {auto,True,False}`) and an automatic check (`auto`) on whether an internet connection can be made. If enabled/an internet connection is available, the `retrieve.smk` rules are included in the workflow. If disabled / no internet connection is available, the rules are not included, thus not part of the potential DAG and thus `snakemake` can run.

Encountered this problem locally when I had no internet connection and on a HPC cluster recently.

Interesting addition for PyPSA-EUR?

## Checklist

- [n/a] Changed dependencies are added to `envs/environment.yaml`.

### DIFF
--- a/Snakefile
+++ b/Snakefile
@@ -48,7 +48,10 @@ wildcard_constraints:
 
 include: "rules/common.smk"
 include: "rules/collect.smk"
-include: "rules/retrieve.smk"
+if (config["enable"].get("retrieve","auto") == "auto" and has_internet_access())or config["enable"]["retrieve"] is True:
+    include: "rules/retrieve.smk"
+else:
+    print("Datafile downloads disabled in config[retrieve] or no internet access.")
 include: "rules/build_electricity.smk"
 include: "rules/build_sector.smk"
 include: "rules/solve_electricity.smk"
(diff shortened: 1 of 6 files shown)


FabianHofmann:
Very nice. I would propose to move the if else statements to retrieve.smk file in order to have the main snakefile clean. There is already a lot of if statements in the .smk, which can be expanded by `retrieve_enabled`, defined through
```
retrieve_enabled = config["enable"]["retrieve"]
if retrieve_enabled ==  "auto":
    if has_internet_access():
        retrieve_enabled = True
    else:
        print("No internet access provided. Datafile downloads automatically disabled.")
        retrieve_enabled = False
```
   
E.g. the retrieve cutout function would be adjusted to

```

if config["enable"].get("retrieve_cutout", True) and retrieve_enabled:
...

```

euronion:
Thanks for the suggestion, RTR @FabianHofmann

FabianHofmann (approved):

FabianHofmann:
great!

PR #711, opened 2023-08-02
prepare_sector: add biomass transport costs for networks wo biomass network

FabianHofmann:
Add biomass transport cost to nodes in case no biomass transport network is considered.

### DIFF
--- a/rules/build_sector.smk
+++ b/rules/build_sector.smk
@@ -295,25 +295,34 @@ rule build_biomass_potentials:
         "../scripts/build_biomass_potentials.py"
 
 
-rule build_biomass_transport_costs:
-    input:
-        transport_cost_data=HTTP.remote(
-            "publications.jrc.ec.europa.eu/repository/bitstream/JRC98626/biomass potentials in europe_web rev.pdf",
-            keep_local=True,
-        ),
-    output:
-        biomass_transport_costs=RESOURCES + "biomass_transport_costs.csv",
-    threads: 1
-    resources:
-        mem_mb=1000,
-    log:
-        LOGS + "build_biomass_transport_costs.log",
-    benchmark:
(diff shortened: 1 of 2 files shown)


rules/build_sector.smk
```diff
@@ -782,7 +772,7 @@ rule prepare_sector_network:
         + "prenetworks/elec_s{simpl}_{clusters}_l{ll}_{opts}_{sector_opts}_{planning_horizons}.nc",
     threads: 1
     resources:
-        mem_mb=2000,
+        mem_mb=20000,
```
lisazeyen, line 775: that's quite a large increase, is that much memory necessary?

scripts/prepare_sector_network.py
```diff
@@ -2185,6 +2185,23 @@ def add_biomass(n, costs):
             carrier="solid biomass transport",
         )
 
+    elif options["biomass_spatial"]:
```
lisazeyen, line 2188: What if biomass is not spatially resolved (just one EU node for the biomass). Should the transport costs be added? If not one could make the rule to build the biomass transport costs conditional again. If yes, the generator should be added here as well.

lisazeyen (changes requested):

rules/build_sector.smk
FabianHofmann: that was actually not intended, thanks for pointing out

scripts/prepare_sector_network.py
FabianHofmann: I see, haven't thought about this option. Will make it conditional again

FabianHofmann:
thanks @lisazeyen, I fixed the things, please approve if you are fine with the changes

lisazeyen (approved):
Great, looks good!

PR #712, opened 2023-08-02
harmonize output of solving rules

FabianHofmann:
only copy active configuration to results folder
use copy config for all solve rules
remove copying of environment

## Checklist

### DIFF
--- a/rules/postprocess.smk
+++ b/rules/postprocess.smk
@@ -39,7 +39,7 @@ rule copy_config:
     params:
         RDIR=RDIR,
     output:
-        RESULTS + "config/config.yaml",
+        RESULTS + "config.yaml",
     threads: 1
     resources:
         mem_mb=1000,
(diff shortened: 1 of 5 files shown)


fneum (approved):
LGTM. We could think of adding a commit-hash to `n.meta`, like `abcd1234-dirty` to follow version of preparation and solving scripts.

FabianHofmann:
> LGTM. We could think of adding a commit-hash to `n.meta`, like `abcd1234-dirty` to follow version of preparation and solving scripts.

Indeed, and since we are storing the config in the network meta anyway, should we just get rid of the copy config for good?

fneum:
Hmm, you're right that there's some duplication but maybe a YAML is easier to inspect ad-hoc.

lisazeyen:
> 

I would favor keeping the .yaml file. It is also useful if you want to use it for a summary script of all scenarios since you do not have to read in a pypsa network to check the n.meta

PR #720, opened 2023-08-09
Fix in co2_emissions_year function and correspondingly in plot_summary.py and snakemake postprocess.smk

daniel-rdt:
fix in co2_emissions_year function and correspondingly in plot_summary and snakemake rule. `emissions_scope` and `report_year` were not properly assigned. Also input_co2 function parameter was added for usage of `co2_emissions_year` function outside of `prepare_sector_network` script.
Also changes to historical_emissions in plot_summary since some countries are not in eea dataset. Remove these countries from countries list.

## Checklist

### DIFF
--- a/scripts/plot_summary.py
+++ b/scripts/plot_summary.py
@@ -387,6 +387,19 @@ def historical_emissions(countries):
         countries.remove("GB")
         countries.append("UK")
 
+    # Albania (AL) and Bosnia Herzegovina (BA), Montenegro (ME), Macedonia (MK) and  Serbia (RS)
+    # not included in eea historical emission dataset
+    if "AL" in countries:
+        countries.remove("AL")
+    if "BA" in countries:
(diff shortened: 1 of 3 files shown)


scripts/plot_summary.py
```diff
@@ -387,6 +387,19 @@ def historical_emissions(countries):
...
+        countries.remove("AL")
+    if "BA" in countries:
+        countries.remove("BA")
+    if "ME" in countries:
+        countries.remove("ME")
+    if "MK" in countries:
+        countries.remove("MK")
+    if "RS" in countries:
+        countries.remove("RS")
+
```
FabianHofmann, line 402:
```suggestion
    # remove countries which are not included in eea historical emission dataset
    countries_to_remove = {"AL", "BA", "ME", "MK", "RS"}
    countries = list(set(countries) - countries_to_remove)
```

PR #724, opened 2023-08-15
Scenario management

FabianHofmann:
From release notes:

 PyPSA-EUR now supports the simultaneous execution of multiple scenarios. For this purpose, a scenarios.yaml file has been introduced which contains customizable scenario names with corresponding configuration overrides. To enable it, set the ``run: scenarios:`` key to ``True`` and define the scenario names to run under ``run: name:`` in the configuration file. The latter must be a subset of toplevel keys in the scenario file. 

## Checklist


FabianHofmann:
@fneum if you want you can already have a look, I will test it further locally for my project.

rules/common.smk
```diff
@@ -2,6 +2,57 @@
...
+        return get_config(keys, config, default)
+
+    def dynamic_getter(wildcards):
+        """Getter function for dynamic config values based on scenario."""
+        scenario_name = wildcards.run
+        if scenario_name not in scenarios:
+            raise ValueError(
+                f"Scenario {scenario_name} not found in file {config['scenariofile']}."
+            )
+        merged_config = merge_configs(config, scenarios[scenario_name])
```
coroa, line 48: That potentially looks like a hammered bit of code! Should there be an in-memory per-scenario cache?

FabianHofmann: Good idea!

FabianHofmann:
Note: Sweeping across years also works with this. For example using the scenario.yaml

```yaml
network2013:
  snapshots:
    start: "2013-01-01"
    end: "2014-01-01"
    inclusive: 'left'


network2019:
  snapshots:
    start: "2019-01-01"
    end: "2020-01-01"
    inclusive: 'left'

  renewable:
    onwind:
      cutout: europe-2019-era5
    offwind-ac:
      cutout: europe-2019-era5
    offwind-dc:
      cutout: europe-2019-era5
    solar:
      cutout: europe-2019-era5
    hydro:
      cutout: europe-2019-era5

```

will create two resource and result folders with distinct years.

fneum:
> Note: Sweeping across years also works with this. For example using the scenario.yaml

Oh, I am excited about this!

energyLS:
@FabianHofmann I have tested it, in my case two scenarios with varying `renewable_carriers` tested with `snakemake -call solve_sector_networks`. Works fine, I like the structure and introduction of `scenarios.yaml` which is quite intuitive to use.
Anyhow, snakemake complains when using `make_summary` rule:
```
WorkflowError:
Target rules may not contain wildcards. Please specify concrete files or a rule without wildcards at the command line, or have a rule without wildcards at the very top of your workflow (e.g. the typical "rule all" which just collects all results you want to generate in the end).
(pypsa-eur) energyls@ei-lan-346:/mnt/c/Users/scl38887/Documents/git/pypsa-eur$
```

even though the output of the rule does not include wildcards. At the first glance I assume the fetching of `config` has some issues here.
Regarding the discussion on the `shared_resources` implementation: For which use cases do you see the benefit of having a differentation of wildcards?
Feel free to reach out if you require more testing/contributions from my side :)

fneum:
Just one initial question: Is there a reason why the instances where the `config` object is used for workflow control (e.g. `if config["foresight"] == "overnight":`) is not using the `config_provider()` function?

FabianHofmann:
> Just one initial question: Is there a reason why the instances where the `config` object is used for workflow control (e.g. `if config["foresight"] == "overnight":`) is not using the `config_provider()` function?

Yes, the `config_provider` returns a function itself. This function takes wildcards as an argument to make the config scenario aware where possible. The `if` statements happen at a level where no wildcard is defined or passed by snakemake. So, these case differentiations have to rely on the base config instead of the scenario aware `config_provider` function.

rules/build_electricity.smk
```diff
@@ -20,7 +20,7 @@ if config["enable"].get("prepare_links_p_nom", False):
 
 rule build_electricity_demand:
     params:
-        snapshots={k: config["snapshots"][k] for k in ["start", "end", "inclusive"]},  # TODO: use config provider
+        snapshots={k: config_provider("snapshots", k) for k in ["start", "end", "inclusive"]},
```
FabianHofmann: for such cases, the function comes at its limits. This is likely not working, as the return type of the config_provider("snapshots", k) is a function. Probably, we would have to go for something like 
```
    snapshots=lambda w: {k: config_provider("snapshots", k)(w) for k in ["start", "end", "inclusive"]},
```
It looks like there are some more cases like that in this commit.

To make it more readable, we can also rename `config_provider` to `configuration` or `conf`

fneum:
Ok, so far I like the `config_provider`, but some challenging bits are these input expand dictionaries here:

```py
**{
    f"profile_{tech}": resources(f"profile_{tech}.nc")
    for tech in config_provider("electricity", "renewable_carriers")
},
**{
    f"conventional_{carrier}_{attr}": fn
    for carrier, d in config.get("conventional", {None: {}}).items()
    if carrier in config_provider("electricity", "conventional_carriers")
    for attr, fn in d.items()
    if str(fn).startswith("data/")
},
```

fneum:
> Ok, so far I like the `config_provider`, but some challenging bits are these input expand dictionaries here:
> 
> ```python
> **{
>     f"profile_{tech}": resources(f"profile_{tech}.nc")
>     for tech in config_provider("electricity", "renewable_carriers")
> },
> **{
>     f"conventional_{carrier}_{attr}": fn
>     for carrier, d in config.get("conventional", {None: {}}).items()
>     if carrier in config_provider("electricity", "conventional_carriers")
>     for attr, fn in d.items()
>     if str(fn).startswith("data/")
> },
> ```

Found the solution: https://snakemake.readthedocs.io/en/v7.32.3/snakefiles/rules.html#input-functions-and-unpack

```py
def input_profile_tech(w):
    return {f"profile_{tech}": resources(f"profile_{tech}.nc") for tech in config_provider("electricity", "renewable_carriers")(w)}

rule add_electricity:
    ...
    input:
        unpack(input_profile_tech),
    ...
```

config/create_scenarios.py
```diff
@@ -0,0 +1,37 @@
...
+scenario{scenario_number}:
+    config_section:
+        config_key: {config_value}
+    config_section2:
+        config_key2: {config_value2}
+"""
+
+# Define all possible combinations of config values.
+# This must define all config values that are used in the template.
+config_values = dict(config_value=["true", "false"], config_value2=[1, 2, 3, 4])
```
lindnemi, line 28: `config_value=` should probably be called `config_key=` in this line to avoid confusion

fneum:
merged via 184f3a0e03686ff621ee70c2bdb5dcdc039d9e01

PR #726, opened 2023-08-21
drop duplicates in prepare_hotmaps_database

lisazeyen:
Closes # (if applicable).

The script `build_industrial_distribution_keys` breaks for some spatial resolution since there can appear duplicates in the hotmaps database because of a `.sjoin` function with the regions. In this PR the duplicated rows should be checked against the country in the original dataframe and all the duplicated sites which are set to the wrong country should be dropped.

## Checklist


fneum (approved):
LGTM!

Just exchanged `pycountry` with `country_converter` as we should strive to use only one country conversion package.

Issue #727, opened 2023-08-22
fix p_nom_min for renewable generators in myopic configuration


martavp:
In the overnight configuration, p_nom_min for renewable generators is set based on existing historical capacities. This is fine. 

In the myopic configuration, when setting p_nom_max for renewable generators, the existing capacities are enforced over the maximum potential based on available land. The existing capacities have been written by prepare_network.py as p_nom_min in every expandable renewable generator). This is not correct.

A practical example, let's say that the existing solar capacity before 2020 in Spain is 5000 MW, and we want to run myopic for [2020,2030]
prepare_network.py creates two prenetworks with
generator.p_nom_min['ES solar-2020']=5000
generator.p_nom_min['ES solar-2030']=5000
but during the myopic run for 2020, the optimization yields generator.p_nom_opt['ES solar-2020']=maximum capacity based on available land. 

Then, what should happen is that generator.p_nom_min['ES solar-2030']=0

This is not what it does currently.

## Checklist
- [x ] I am using the current [`master`](https://github.com/PyPSA/pypsa-eur/tree/master) branch.
- [x ] I am running on an up-to-date [`pypsa-eur` environment](https://github.com/PyPSA/pypsa-eur/blob/master/envs/environment.yaml). Update via `conda env update -f envs/environment.yaml`.

`

PR #728, opened 2023-08-22
Fix min capacity myopic

martavp:
Closes https://github.com/PyPSA/pypsa-eur/issues/727

Fix the minimum capacity set for renewable generators when running the myopic configuration. 

I commented out the current code that imposes p_nom_min over p_nom_max. 
I think we don't need it anymore, but I'm not sure if I'm missing anything

## Checklist

### DIFF
--- a/scripts/solve_network.py
+++ b/scripts/solve_network.py
@@ -63,18 +63,18 @@ def _add_land_use_constraint(n):
         existing.index += " " + carrier + "-" + snakemake.wildcards.planning_horizons
         n.generators.loc[existing.index, "p_nom_max"] -= existing
 
-    # # check if existing capacities are larger than technical potential
-    # existing_large = n.generators[
-    #     n.generators["p_nom_min"] > n.generators["p_nom_max"]
-    # ].index
-    # if len(existing_large):
-    #     logger.warning(
-    #         f"Existing capacities larger than technical potential for {existing_large},\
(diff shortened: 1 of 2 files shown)


scripts/solve_network.py
```diff
@@ -60,18 +63,18 @@ def _add_land_use_constraint(n):
         existing.index += " " + carrier + "-" + snakemake.wildcards.planning_horizons
         n.generators.loc[existing.index, "p_nom_max"] -= existing
 
-    # check if existing capacities are larger than technical potential
```
lisazeyen, line 63: I am not sure if it is good to remove this part. I guess it can also happen that the existing capacities which are non-extendable are already larger than the technical potential which leads to infeasibilities. So it would be nice to keep the warning and the following part.

martavp: I think you are right, this is very unlikely but not impossible, so maybe we just keep it.

lisazeyen: I think I already got this warning especially with a higher spatial resolution, so I would favor to keep it

lisazeyen: I think I already got this warning especially with a higher spatial resolution, so I would favor to keep it since otherwise the network is infeasible to solve

martavp: we agree, I just reverted the commented lines

PR #730, opened 2023-08-24
apply protected() to zenodo downloads; only retrieve monthly fuel/CO2…

fneum:
… prices when necessary


lisazeyen (approved):
Nice!

euronion:
Out of curiosity: How does this change interact with the databundle for tutorial / non-tutorial? Will the proper files nevertheless still be downloaded if one switches from tutorial to non-tutorial? Or will the download fail?

PR #741, opened 2023-09-20
Move all rule to main Snakefile

chrstphtrs:


FabianHofmann:
We could even move rule `all` down to the other rules or?

fneum:
You mean before `rule purge`? Yes. One would just have to check if `snakemake -j8` would still trigger the `all` rule, or if it would go into one of the `include: ...` files

coroa:
`snakemake` triggers the first rule by default, the name `all` is convention.

chrstphtrs:
Snakemake does not trigger rules from the include block. That was the reason we had to move the all rule outside of there in the first place.So yes, should move the all rule down to start off the block of other rules.Am 21.09.2023 um 10:25 schrieb Jonas Hörsch ***@***.***>:﻿
snakemake triggers the first rule by default, the name all is convention.

PR #743, opened 2023-09-22
readthedocs: remove apt packages

FabianHofmann:
should fix the readthedocs built


fneum:
Would be nice if that worked! So far the graphviz elements in the documentation do not seem to be rendered:

https://pypsa-eur--743.org.readthedocs.build/en/743/tutorial.html#how-to-use-snakemake-rules

FabianHofmann:
Finally working :)

PR #757, opened 2023-10-11
Add option for SMR CC

tgi-climact:
This PR splits  the configuration to enable SMR and SMR CC. This is useful when we plan to avoid CC.

## Checklist

### DIFF
--- a/config/config.default.yaml
+++ b/config/config.default.yaml
@@ -452,7 +452,7 @@ sector:
   hydrogen_fuel_cell: true
   hydrogen_turbine: false
   SMR: true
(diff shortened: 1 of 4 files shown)


config/config.default.yaml
```diff
@@ -452,6 +452,7 @@ sector:
   hydrogen_fuel_cell: true
   hydrogen_turbine: false
   SMR: true
+  SMR_cc: false
```
fneum, line 455:
```suggestion
  SMR_cc: true
```
To keep current defaults.

fneum (approved):

PR #759, opened 2023-10-13
Update calculations of retrofit costs

ekatef:
The PR is aimed to update retrofit calculations to make the model work with `retro_endogen: false`. The main reasons why updates are needed is that data structure may have been altered and pandas version has changed.

-  replace pandas.append() was deprecated for a while and has been removed since pandas 2.0;
-  fix structure of area_tot (= total area) grouped dataframe to avoid messed concatenated columns;
-  fix a column name which contains subsector names for a building dataset: building_data.detail currently contains only [nan 'insulation' 'low-e'], while the script looks for "Totals" value which is currently contained in building_data.subsector;
-  deal with AssertionError: Length of new_levels (2) must be same as self.nlevels (3) which is currently arisen by reindex() function for prepare_building_topology( ) and map_tabula_to_hotmaps( ). That seems to be caused https://github.com/pandas-dev/pandas/issues/54909 for operation on insertion of smaller MultiIndex into larger MultiIndex and has been fixed recently with https://github.com/pandas-dev/pandas/pull/54885, which is one of milestones of pandas 2.1.1;
-  fix country codes in building_data to avoid messing-up "GB" and "UK";
-  remove duplicated country column in area_reordered;
-  fix call of difference( ) function;
-  fix computation of grouped means and total averages;
-  replace a level name with a number of the level.

Apart of the code fixes, the datakit structure should be ajusted a bit: `tabula-calculator-calcsetbuilding.csv` should be moved to `/pypsa-eur/data/retro`, where the workflow expects to find it.

## Checklist

### DIFF
--- a/envs/environment.yaml
+++ b/envs/environment.yaml
@@ -11,7 +11,7 @@ dependencies:
 - pip
 
 - atlite>=0.2.9
-- dask<=2023.9.1
+- dask
 
   # Dependencies of the workflow itself
 - xlrd
--- a/envs/environment.yaml
+++ b/envs/environment.yaml
@@ -11,7 +11,7 @@ dependencies:
 - pip
 
 - atlite>=0.2.9
-- dask
+- dask<=2023.9.1
 
   # Dependencies of the workflow itself
 - xlrd
@@ -27,7 +27,7 @@ dependencies:
 - numpy
 - pandas>=1.4
 - geopandas>=0.11.0
-- xarray
+- xarray<=2023.8.0
 - rioxarray
 - netcdf4
 - networkx
--- a/rules/build_sector.smk
+++ b/rules/build_sector.smk
@@ -609,7 +609,7 @@ if config["sector"]["retrofitting"]["retro_endogen"]:
             countries=config["countries"],
         input:
             building_stock="data/retro/data_building_stock.csv",
-            data_tabula="data/retro/tabula-calculator-calcsetbuilding.csv",
+            data_tabula="data/bundle-sector/retro/tabula-calculator-calcsetbuilding.csv",
             air_temperature=RESOURCES + "temp_air_total_elec_s{simpl}_{clusters}.nc",
             u_values_PL="data/retro/u_values_poland.csv",
             tax_w="data/retro/electricity_taxes_eu.csv",
--- a/scripts/build_retro_cost.py
+++ b/scripts/build_retro_cost.py
@@ -198,6 +198,8 @@ def prepare_building_stock_data():
         }
     )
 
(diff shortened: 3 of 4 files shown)


FabianHofmann:
Thanks, @ekatef for the PR. Looks good! CI is failing due to some internet connection problems, but let's wait till it's through. @lisazeyen, you approve?

martacki (commented):
Can confirm that the workflow runs with suggested data structure changes. But instead of moving the tabula data around, we could also change the respective path. Then the sector-bundle doesn't need updating (might be confusing if the files appears twice), i.e.
in `rules/build_sector.smk`, change

> data_tabula="data/retro/tabula-calculator-calcsetbuilding,csv",

to

> data_tabula="data/bundle-sector/retro/tabula-calculator-calcsetbuilding.csv",

Thoughts ?

envs/environment.yaml
```diff
@@ -11,7 +11,7 @@ dependencies:
 - pip
 
 - atlite>=0.2.9
-- dask
+- dask<=2023.9.1
```
martacki, line 14: For me, this doesn't need downgrading. Worked also with 2023.9.3.

ekatef:
> Can confirm that the workflow runs with suggested data structure changes. But instead of moving the tabula data around, we could also change the respective path. Then the sector-bundle doesn't need updating (might be confusing if the files appears twice), i.e. in `rules/build_sector.smk`, change
> 
> > data_tabula="data/retro/tabula-calculator-calcsetbuilding,csv",
> 
> to
> 
> > data_tabula="data/bundle-sector/retro/tabula-calculator-calcsetbuilding.csv",
> 
> Thoughts ?

Agree, fixed.

envs/environment.yaml
ekatef: > For me, this doesn't need downgrading. Worked also with 2023.9.3.

> For me, this doesn't need downgrading. Worked also with 2023.9.3.

Great news.  According to the recent reports for pypsa-earth-sec, constraining `atlite` is enough. Dropped `dask` constraint.

lisazeyen:
> Thanks, @ekatef for the PR. Looks good! CI is failing due to some internet connection problems, but let's wait till it's through. @lisazeyen, you approve?

Looks good to me, only the CI needs to run through

martacki (commented):
With the suggested changes, I was able to reproduce the retrofitting costs from "Mitigating Heat Demand Peaks" Study, Supplementary Material, Table 3. :tada: 

Just a quick note, in line 950, I need to un-comment renaming the column from `country` to `country_code`, otherwise I'll get an error (`country_code` does not exist). But could be that it's just me?

PR #767, opened 2023-10-25
Minor fixes for retrofit calculations

ekatef:

Fixing some minor discrepancies for buildings area calculations in `build_retro_cost` script. In particular:
1) U-values for Poland are represented in hotmaps dataset as `Construction features (U-value)` instead of `Construction features (U-values)`;
2) "Total" feature is contained in `subsector` column, not in `detail` one.

## Checklist

### DIFF
--- a/scripts/build_retro_cost.py
+++ b/scripts/build_retro_cost.py
@@ -203,7 +203,7 @@ def prepare_building_stock_data():
 
     # heated floor area ----------------------------------------------------------
     area = building_data[
-        (building_data.type == "Heated area [Mm²]") & (building_data.detail != "Total")
+        (building_data.type == "Heated area [Mm²]") & (building_data.subsector != "Total")
     ]
     area_tot = area[["country", "sector", "value"]].groupby(["country", "sector"]).sum()
     area = pd.concat(
--- a/scripts/build_retro_cost.py
+++ b/scripts/build_retro_cost.py
@@ -169,7 +169,7 @@ def prepare_building_stock_data():
             "Construction features (U-value)": "Construction features (U-values)",
         },
         inplace=True,
-    )     
+    )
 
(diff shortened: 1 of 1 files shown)


martacki:
The solar global radiation was somehow off, our updated values are taken from this [source](http://www.episcope.eu/fileadmin/tabula/public/docs/report/TABULA_CommonCalculationMethod.pdf) (which is also referenced in the script): http://www.episcope.eu/fileadmin/tabula/public/docs/report/TABULA_CommonCalculationMethod.pdf

martacki (commented):
With the suggested changes, I was able to reproduce the retrofitting costs from "Mitigating Heat Demand Peaks" Study, Supplementary Material, Table 3. :tada:  

Just a quick note, in line 950, I need to un-comment renaming the column from `country` to `country_code`, otherwise I'll get an error (`country_code` does not exist). But could be that it's just me?

fneum (approved):

fneum:
Could you confirm whether a renaming of country/country code is necessary?

Then, this can be merged.

yerbol-akhmetov:
Good day, I have also applied the changes and observed the produced retrofitting costs. The resulting file is quite similar to the aforementioned paper's results in terms of order of scale. I have observed some differences for various countries (AT, BA and etc). Regarding uncommenting line 950, I have first tried as it is (commented) and I did not see any error. When I have uncommented it, I have faced an error.

martacki:
@ekatef and I discussed the above mentioned issue, turns out it is (as suspected) on my end, so no need to commit my work-around.

martacki:
@ekatef and I have discussed the above mentioned issue about renaming, but it turns out that the error is on my end. Not sure what's wrong, but I think it's clean as is. Ready to merge.

Issue #367, opened 2022-05-17
Include "Nationally Designated Areas" (CDDA) in land eligibility


fneum:
Nationally designated areas (CDDA)

https://de.wikipedia.org/wiki/Common_Database_on_Designated_Areas

https://www.eea.europa.eu/data-and-maps/data/nationally-designated-areas-national-cdda-16

older version with UK exists

https://www.eea.europa.eu/data-and-maps/data/nationally-designated-areas-national-cdda-14

fneum:
https://sdi.eea.europa.eu/datashare/s/wC9kS3yBCATNH5f/download?path=%2FGPKG&files=CDDA_2023_v01_public.gpkg

PR #768, opened 2023-10-28
exclude forests from onshore wind eligible areas

fneum:
https://collections.sentinel-hub.com/corine-land-cover/readme.html

### DIFF
--- a/config/config.default.yaml
+++ b/config/config.default.yaml
@@ -161,7 +161,7 @@ renewable:
     capacity_per_sqkm: 3
     # correction_factor: 0.93
     corine:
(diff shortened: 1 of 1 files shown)


fneum:
On the other hand, allowing all forest areas as potential is also unrealistic.

https://www.agora-energiewende.de/daten-tools/photovoltaik-und-windflaechenrechner

Agora has a nice dashboard which allows certain shares of forest areas.

Generally, we have quite optimistic onshore wind potentials.

euronion:
Protected forest areas should be excluded via the natural protected areas exclusion (NATURA2000 or switch to WDPA like for Ukraine, which should cover more than NATURA2000 forests).

Didn't a master student of your's look into this?

fneum:
will add option to consider CDDA instead #367

PR #769, opened 2023-10-30
Fixing biomass transport cost

yerbol-akhmetov:
Good day. Here I propose changes to `build_biomass_transport_costs.py` file to fix the error that happens when `biomass_transport` is set true in the config. I have uploaded the error. It is related to utf-8 encoding, in particular, the error happens due to EUR sign. I run my code in the windows machine. The error is resolved by explicitly defining encoding as `cp1252` which is a windows compatible encoding and an extension of `latin-1` encoding. After I have fixed the code, I have checked the resulting csv. The resulting biomass transport costs coincides with the supplementary material in Liza's paper on "Mitigating heat demand ...". 

error utf-8
fix_biomass

## Checklist

### DIFF
--- a/scripts/prepare_sector_network.py
+++ b/scripts/prepare_sector_network.py
@@ -2708,9 +2708,8 @@ def add_industry(n, costs):
     )
 
     p_set_methanol = (
-            industrial_demand["methanol"]
-            .rename(lambda x: x + " industry methanol")
-            / nhours
+        industrial_demand["methanol"].rename(lambda x: x + " industry methanol")
+        / nhours
     )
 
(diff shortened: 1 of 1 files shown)


martacki (changes requested):
With the currently suggested changes, the script cannot be compiled on Linux. We're working on a fix.

yerbol-akhmetov:
Hello, @martacki. I have added encoding based on the OS. For Windows it is `cp1252`, while for Linux and MacOS it is None, because it works fine without it (while `cp1252` causes issue for Linux).

PR #772, opened 2023-10-31
madd oil generator to bus EU oil

martacki:
Add a oil generator to bus `EU oil`, when adding the transport sector. Without it, existing loads lead to a LHS RHS imbalance in the LP file and the isolated transport sector model cannot be solved (`Error: ValueError: Empty LHS with non-zero RHS in nodal balance constraint`).

## Checklist


fneum:
Thanks! We can directly use `add_carrier_buses` for this.

PR #780, opened 2023-11-16
Review endogenous transport

lisazeyen:
Closes # (if applicable).
This PR is a review of PR #734 

## Checklist


martavp:
My initial idea was to maintain the exogenous option as it is now (before this PR). Then, the land transport demand is always in units of MWh_electricity that are needed and the efficiencies take into account the type of vehicle (e.g. if it is an ICE you need to consume additional MWh_oil due to the low efficiency of the ICE). This would compensate for the [efficiency_gain](https://github.com/PyPSA/pypsa-eur/blob/54afa891cc4029545cbf4928b822ac7683bbd5e5/scripts/build_transport_demand.py#L50) but I overlooked the [temperature dependence correction](https://github.com/PyPSA/pypsa-eur/blob/54afa891cc4029545cbf4928b822ac7683bbd5e5/scripts/build_transport_demand.py#L58)

My summary of how I see it now.

**LAND TRANSPORT DEMAND EXPRESS IN MWh_elec**
[+] Exogenous code does not change, i.e. same calculation in build_transport_demand.py for exogenous and endogenous
[-] Efficiency for VE=1 but Efficiency for ICE and Fuel cell can be counterintuitive 

**LAND TRANSPORT DEMAND EXPRESS IN MWh_oil**
[+] Efficiency for VE=[efficiency_gain](https://github.com/PyPSA/pypsa-eur/blob/54afa891cc4029545cbf4928b822ac7683bbd5e5/scripts/build_transport_demand.py#L50), efficiency_for_ICE=1, efficiency_for_H2=efficiency_gain_for_H2 (to be calculated)
[- ] Different calculation in build_transport_demand.py for exogenous and endogenous. In the first case, it will be expressed in MWh_elec while in the second in MWh_oil

lisazeyen:
> My initial idea was to maintain the exogenous option as it is now (before this PR). Then, the land transport demand is always in units of MWh_electricity that are needed and the efficiencies take into account the type of vehicle (e.g. if it is an ICE you need to consume additional MWh_oil due to the low efficiency of the ICE). This would compensate for the [efficiency_gain](https://github.com/PyPSA/pypsa-eur/blob/54afa891cc4029545cbf4928b822ac7683bbd5e5/scripts/build_transport_demand.py#L50) but I overlooked the [temperature dependence correction](https://github.com/PyPSA/pypsa-eur/blob/54afa891cc4029545cbf4928b822ac7683bbd5e5/scripts/build_transport_demand.py#L58)
> 
> My summary of how I see it now.
> 
> **(A) LAND TRANSPORT DEMAND EXPRESS IN MWh_elec** [+] Exogenous code does not change, i.e. same calculation in build_transport_demand.py for exogenous and endogenous [-] Efficiency for VE=1 but Efficiency for ICE and Fuel cell can be counterintuitive
> 
> **(B) LAND TRANSPORT DEMAND EXPRESS IN MWh_oil** [+] Efficiency for VE=[efficiency_gain](https://github.com/PyPSA/pypsa-eur/blob/54afa891cc4029545cbf4928b822ac7683bbd5e5/scripts/build_transport_demand.py#L50), efficiency_for_ICE=1, efficiency_for_H2=efficiency_gain_for_H2 (to be calculated) [- ] Different calculation in build_transport_demand.py for exogenous and endogenous. In the first case, it will be expressed in MWh_elec while in the second in MWh_oil
> 
> At this point, after having realised that efficiency definitions are indeed counterintuitive and that the build_transport_demand needs to be modified anyway because the temperature dependence only applies to the EVs, I think option (b) is better.
> 
> What do you think @s8au ?

@martavp @s8au I agree option B is better. But I was rather thinking getting the actual energy demand by dividing the JRC-IDEES by the average fuel efficiency in the `build_transport_demand` script and calculating out the temperature correction (see last commit). This should be done for the exogenous and the endogenous case. The way it is done currently is simply wrong also for the exogenous case with the temperature correction if we have a share of combustion cars.

Then, in the `prepare_sector_network` take efficiencies of the corresponding car type (so ICE would be average fuel efficiency), assume correction for heating and cooling depending on the car type (I guess for fuel cells I would assume the same as for internal combustion cars, since the efficiency losses for heating are probably also smaller than for EVs since they also combust the fuel).

martavp:
> But I was rather thinking getting the actual energy demand by dividing the JRC-IDEES by the average fuel efficiency in the `build_transport_demand` script and calculating out the temperature correction (see last commit). This should be done for the exogenous and the endogenous case. 

I agree. Then, the land_transport_demand is always in  MWh of propulsion or mechanical energy, both for the endogenous and the exogenous

> The way it is done currently is simply wrong also for the exogenous case with the temperature correction if we have a share of combustion cars.

Yes, you are right.

lisazeyen:
I would just import the functions to calculate the temperature correction from the `build_transport_demand` script to `prepare_sector_network` and do the corrections there. Then it should be just one line getting the correction. You basically can adjust the efficiencies of the links depending on the temperature correction

PR #785, opened 2023-11-24
change coords name co2 limit per country

lisazeyen:
Closes # (if applicable).
This PR should check the constraint CO2 limit per country

- [x] rename coordinates of the constraint so that dual variables can be assigned correctly
- [ ] rework the constraint or the prepare_sector_network so that process emissions are accounted correctly
- [ ] adjust to decision made for oil carrier (if copper platted emissions should be accounted by index), if oil links are added between the nodes then nothing has to be changed in the constraint

## Checklist


lisazeyen:
I just wanted to open a draft PR which lists the open to-dos. @FabianHofmann can you check if the renaming of the constraint makes sense or if it would be better to do it in another way?

PR #808, opened 2023-12-20
select correct sector name in endogeneous retrofitting

martacki:
Turns out we're making this PR bigger and include many more fixes than originally intended:

- selection of the sector `sec` never chose `services`, thus costs of retrofitting services were highly overestimated (on average by a factor of 4). Fixed by correctly defining the corresponding variable `sec`
- introduced a more general node name that allows more than 9 buses per synchronous region
- avoiding deprecation of `.groupby(axis=1)` and not using `.loc[]`
- discriminate between district heating, decentral heating in urban areas or rural areas for `floor_area_node` calculations
- discriminate between gas boilers and resistive heaters in configuration (config.default.yaml), default settings align with previous settings
- remove gas generators in sector network if gas is a carrier (only gas store allowed)

## Checklist


scripts/prepare_sector_network.py
```diff
@@ -425,7 +425,7 @@ def update_wind_solar_costs(n, costs):
 
             logger.info(
                 "Added connection cost of {:0.0f}-{:0.0f} Eur/MW/a to {}".format(
-                    connection_cost[0].min(), connection_cost[0].max(), tech
+                    connection_cost.min(), connection_cost.max(), tech
```
martacki, line 428: wasn't sure with that one honestly, but using `connection_cost[0]` basically left with the min and max of a single value

also it was deprecated, with the following message: `FutureWarning: Series.__getitem__ treating keys as positions is deprecated. In a future version, integer keys will always be treated as labels (consistent with DataFrame behavior). To access a value by position, use ser.iloc[pos]`

fneum: ok, I would say.

scripts/prepare_sector_network.py
```diff
@@ -497,6 +497,9 @@ def add_carrier_buses(n, carrier, nodes=None):
         capital_cost=capital_cost,
     )
 
+    if carrier == "gas":
+        return
```
fneum, line 501:
```suggestion
```

This would prohibit the use of fossil gas, wouldn't it? 

Since the store is cyclic, where would the model source fossil gas?

fneum (changes requested):
Most changes look good. Thanks @martacki.

I am just unsure about skipping the gas generator. Wouldn't that mean the model can't use fossil gas anymore?

fneum:
I merged without removing the gas generators. Feel free to amend in another PR if necessary.

PR #823, opened 2023-12-29
[ARIADNE] additional_functionality: simplify and use params

fneum:


nworbmot:
I've changed it in https://github.com/PyPSA/pypsa-eur/pull/827 to allow snakemake itself to be passed.

PR #838, opened 2024-01-03
renewable_profiles: remove conservative potential estimation method

fneum:
This is to simplify the code by eliminating a feature that is not really used.

Opinions?


FabianHofmann:
I think it is sensible!

PR #844, opened 2024-01-04
Distinguish sequestered and stored CO2

fneum:
- stored CO2 can be used for CCU
- sequestered CO2 is stored for a long time and cannot be used for CCU

The link between stored and sequestered CO2 is a one-way street.

The cost of overground CO2 storage in tanks is more expensive than underground sequestration.


fneum:
Tested: The set sequestration limit is kept in both cases.

PR #853, opened 2024-01-09
Decorrelate electrical load year from weather year

tgi-climact:

I suggest to decorrelate the electrical load year from the weather year. Up to now, we are using the `snapshots` as reference configuration to choose the electrical load from OPSD data. But because this load is also used to determine the not modeled part of the electrical load, we need this value to be as accurate as possible. Therefor, we should use a more recent value than the default 2013 value for cutouts.

## Checklist

### DIFF
--- a/config/config.default.yaml
+++ b/config/config.default.yaml
@@ -283,6 +283,7 @@ transformers:
 
 # docs-load in https://pypsa-eur.readthedocs.io/en/latest/configuration.html#load
 load:
+  load_year: 2018
   power_statistics: true
   interpolate_limit: 3
   time_shift_for_large_gaps: 1w
--- a/config/config.default.yaml
+++ b/config/config.default.yaml
@@ -283,7 +283,7 @@ transformers:
 
 # docs-load in https://pypsa-eur.readthedocs.io/en/latest/configuration.html#load
 load:
-  load_year: 2018
+  load_year: false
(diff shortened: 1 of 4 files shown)


tgi-climact:
I made it optional to take into account thermal senstive loads.

martacki:
But there is a high correlation between the temperature and the electrical load, isn't there? See for example

tgi-climact:
Thanks @martacki for your reply. The data is indeed highly correlated but here is my problem I need more recent data for electrical load (2013 is way too old) and cutouts are only available for 2013. I have successfully built a `europe-2018-era5.nc` but computing `europe-2018-sarah.nc` is much more intensive That is why I suggested to decorrelate both dates. Would you suggest anything else ?

martacki:
I agree, it's hard to build newer solar profiles for other years than 2013 using SARAH-2 data, but you can also use the era5 data for your solar profiles (instead of the SARAH-2 data). If you do this, we suggest to use a correction factor. With this, the correlation of electricity to weather would be accounted for... if this is better than using SARAH-2 data, I cannot tell. Intuitively, I'd say yes, but I am not aware of any research and qualitative analysis.

tgi-climact:
Great PR there https://github.com/PyPSA/pypsa-eur/pull/204/ ! Closing this one.

PR #617, opened 2023-03-08
sector_opts: generate separate config.yaml for each scenario string

fneum:
https://github.com/PyPSA/pypsa-eur-sec/pull/250

PR #856, opened 2024-01-11
[WIP] add `{sector_opts}` wildcard flags to config file

fneum:
closes #617

### DIFF
--- a/config/config.default.yaml
+++ b/config/config.default.yaml
@@ -365,6 +365,11 @@ existing_capacities:
 
 # docs in https://pypsa-eur.readthedocs.io/en/latest/configuration.html#sector
 sector:
+  transport: true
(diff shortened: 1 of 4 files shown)


koen-vg:
I like this idea but (though I sure you might have noticed this), there are a few more places where `{sector_opts}` is directly checked for the sector flags, e.g. in `emission_sectors_from_opts` (used in building carbon budgets in `prepare_sector_network.py`) as well as in `historical_emissions` in `plot_summary.py`.

fneum:
Yes, I started working on it, and I plan to do it differently: I think I'm going for a function at the beginning of the script where settings from `{sector_opts}` will overwrite entries in `snakemake.config` and then you can use the config object everywhere else in the script.

PR #858, opened 2024-01-11
Add warning when negative bev availability profile values

tgi-climact:

Changing the `bev_avail_max` and `bev_avail_mean`  values could cause the profile to generate negative values. This leads to infeasibility. Raising a warning earlier in the process is useful to catch the configuration error sooner. This warning is raised in `build_transport_demand`.

## Checklist

### DIFF
--- a/doc/release_notes.rst
+++ b/doc/release_notes.rst
@@ -12,7 +12,8 @@ Upcoming Release
 
 * New configuration option ``everywhere_powerplants`` to build conventional powerplants everywhere, irrespective of existing powerplants locations, in the network (https://github.com/PyPSA/pypsa-eur/pull/850).
 
-* Remove option for wave energy as technology data is not maintained. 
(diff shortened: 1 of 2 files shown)


fneum:
It would be good to understand where the negative values could come from and fix the underlying issue.

tgi-climact:
> It would be good to understand where the negative values could come from and fix the underlying issue.

In my case, I was defining a too high value for `bev_avail_max` compared to `bev_avail_mean`. If so, `avail` becomes negative [here](https://github.com/PyPSA/pypsa-eur/blob/master/scripts/build_transport_demand.py#L129) : 

```
avail = avail_max - (avail_max - avail_mean) * (traffic - traffic.min()) / (
        traffic.mean() - traffic.min()
    )
```

Because it's configuration based, I'm afraid there's not that much to do other than raise a warning.

fneum (approved):

PR #859, opened 2024-01-11
Add technology specific renewable profiles for different planning horizons

p-glaum:
Closes # (if applicable).

## Checklist


p-glaum:
I am not so happy with the build renewable profiles, because this takes rather a long time. Suggestions on how to improve performance here without adding new wildcards are welcome.

scripts/build_renewable_profiles.py
```diff
@@ -335,6 +346,29 @@
         **resource,
     )
 
+    if year_dependent_techs and foresight != "overnight":
+        for key, techs in year_dependent_techs.items():
```
fneum, line 350: `year_dependent_techs` would always be a dictionary with one entry, right? In this case, it would be better to 
remove this for-loop.

Regarding performance, did you check that the number of iterations in this nested for-loop is the desired number? If yes, I don't think there is much you can gain in performance.

p-glaum: >year_dependent_techs would always be a dictionary with one entry, right? In this case, it would be better to
remove this for-loop.
>>correct, but the key can change depending whether we consider PV or wind to panel or turbine, respectively

>Regarding performance, did you check that the number of iterations in this nested for-loop is the desired number? 
>>yes, it only runs the function `func` once for every technology

fneum:
superseded by #912

PR #861, opened 2024-01-12
fix clustering bug due to missing countries in n.buses

p-glaum:
filter looking for empty values in n.buses.countries is not working anymore. This leads to errors later on in the `cluster_network` script. for 37 nodes, the error in the cluster script was that there are more combinations of subnetworks and countries than 37 and for higher cluster numbers this lead to a problem due to a division by 0.

The CI did not catch the mistake in the current master, because we only consider one country for the test CI 

## Checklist

### DIFF
--- a/scripts/base_network.py
+++ b/scripts/base_network.py
@@ -560,7 +560,7 @@ def _set_countries_and_substations(n, config, country_shapes, offshore_shapes):
         ~buses["under_construction"]
     )
 
-    c_nan_b = buses.country == "na"
+    c_nan_b = buses.country.isna()
     if c_nan_b.sum() > 0:
         c_tag = _get_country(buses.loc[c_nan_b])
         c_tag.loc[~c_tag.isin(countries)] = np.nan
--- a/doc/release_notes.rst
+++ b/doc/release_notes.rst
@@ -12,7 +12,7 @@ Upcoming Release
 
 * New configuration option ``everywhere_powerplants`` to build conventional powerplants everywhere, irrespective of existing powerplants locations, in the network (https://github.com/PyPSA/pypsa-eur/pull/850).
 
-* Remove option for wave energy as technology data is not maintained. 
+* Remove option for wave energy as technology data is not maintained.
 
 
 PyPSA-Eur 0.9.0 (5th January 2024)


martacki:
I just have changed this due to the new pandas version, where .isna() was not working anymore. Which pandas version are you using?

p-glaum:
I am now also working on the newest pandas version and there isna() is working. is the previous version with `== "na"` working for you?

FabianHofmann:
> I think to avoid any problems with pandas, we could just do `buses.loc[:, "country"] = ""` and then later filter for `c_nan_b = buses.country == ""`

I agree, but since the country column is of dtype string, we should rather track down where the `nan`s come from and make sure we use "" instead. Then, the filter would be `n.buses.country == ""`

p-glaum:
The nan values came from a previous pandas version. In the latest version, those entries are just empty. But we somehow need to filter them an neither `isna()` nor `isnull()` are working .

fneum:
If it only works for a particular pandas version, this should be reflected in the environment.yaml. Ideally, the minimum version should not be the very latest pandas version. We might need a workaround that works for more versions.

p-glaum:
The workaround could be the one I suggested. I.e. do `buses.loc[:, "country"] = ""`
and then later filter for `c_nan_b = buses.country == ""`

FabianHofmann:
or just add a `.fillna("na")` before the comparison

scripts/base_network.py
```diff
@@ -560,7 +560,7 @@ def prefer_voltage(x, which):
         ~buses["under_construction"]
     )
 
-    c_nan_b = buses.country == "na"
+    c_nan_b = buses.country.isna()
```
FabianHofmann, line 563:
```suggestion
    c_nan_b = buses.country.fillna("na") == "na"
```

p-glaum: did you check if this is working for multiple versions? because with some pandas versions, it also did not recognize the entries with `isna()`

fneum: I like this (hacky) solution @FabianHofmann

fneum (approved):

PR #862, opened 2024-01-12
Reformulate co2 constraint

lisazeyen:
Closes # (if applicable).

## Change CO2 emission constraint for sector coupled models
Instead of using the `primary_energy`global constraint formulation this PR puts a constraint on the last snapshot of the co2 atmosphere bus.

TODO: haven't tested yet if performance improves for overnight scenarios
## Checklist


fneum (approved):
Works like charm @lisazeyen!

I double-checked that the objective value and shadow prices are the same.

Performance was about the same for a 100-second model.

Issue #860, opened 2024-01-12
District heating share is not handled correctly with multiple nodes per country


nworbmot:
This is an ugly bug: for some reason `data/district_share.csv` is passed to `energy_totals.csv` first. Then it is distributed over the nodes to `population_weighted_energy_totals.csv` thereby reducing the share according to the number of nodes. E.g. a country with 10% DH share and 5 nodes with equal populations gets 2% DH share. Really the `district_share.csv` should be handled differently. It may need to be nodal in future.

nworbmot:
It is fixed in this commit (not yet merged): https://github.com/PyPSA/pypsa-eur/commit/6417c7a122954a08bd83af1fd639ea160510c864.

fneum:
closed via #884

PR #884, opened 2024-01-22
Merge recent ariadne advancements

fneum:
* Bugfix: Do not reduce district heat share when building population-weighted
  energy statistics. Previously the district heating share was being multiplied
  by the population weighting, reducing the DH share with multiple nodes.

* Move building of daily heat profile to its own rule
  :mod:`build_hourly_heat_demand` from :mod:`prepare_sector_network`.

* In :mod:`build_energy_totals`, district heating shares are now reported in a
  separate file.

* Move calculation of district heating share to its own rule
  :mod:`build_district_heat_share`.

* Move building of distribution of existing heating to own rule
  :mod:`build_existing_heating_distribution`. This makes the distribution of
  existing heating to urban/rural, residential/services and spatially more
  transparent.


fneum:
closes #860

Issue #886, opened 2024-01-23
Regression caused by auto-formatting in 58b012e


koen-vg:
## Checklist

- [x] I am using the current [`master`](https://github.com/PyPSA/pypsa-eur/tree/master) branch or the latest [release](https://github.com/PyPSA/pypsa-eur/releases). Please indicate.
- [x] I am running on an up-to-date [`pypsa-eur` environment](https://github.com/PyPSA/pypsa-eur/blob/master/envs/environment.yaml). Update via `conda env update -f envs/environment.yaml`.

## Describe the Bug

The auto-formatting in 58b012e has caused the snakemake workflow to crash, at least for me, with the following error:
```
SyntaxError in file /home/koen/Dokument/UiT/Research/Modelling/eu-hydrogen/rules/solve_overnight.smk, line 32:
Expecting rule keyword, comment or docstrings inside a rule definition. (solve_overnight.smk, line 32)
```
This seems to be related to https://github.com/snakemake/snakefmt/issues/208.

Would it be possible to revert the commit in question?

PR #888, opened 2024-01-23
Make data retrieval rules use their own minimal conda env

koen-vg:
I usually use the default snakemake options for when to re-run rules; including when the software environment changes. This means, however, that even data-retrieval rules would re-run when changing some dependency version in `environment.yaml`. Such re-runs have become especially painful now that the output files from these data retrieval rules are being marked as protected: this prevents the rule from re-running, and I have to `sudo rm -rf` the databundles! Not great.

The present PR just creates a separate, minimal conda env for the retrieval rules. Just a quality-of-life improvement.

(And it seems some snakefmt fixed have sneaked their way in.)

## Checklist

### DIFF
--- a/envs/retrieve.yaml
+++ b/envs/retrieve.yaml
@@ -1,0 +1,13 @@
+# SPDX-FileCopyrightText: : 2017-2024 The PyPSA-Eur Authors
+#
+# SPDX-License-Identifier: MIT
+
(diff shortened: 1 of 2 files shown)


fneum:
Excellent, thanks @koen-vg! I will also take care of the #886 now.

PR #890, opened 2024-01-23
add air-sourced heat pumps to rural areas

fneum:
previously only ground-sourced heat pumps were considered for rural areas


lisazeyen:
@fneum looks good! I was just wondering if we should make this optional in the config because I would expect that air heat pumps in rural areas are not going to be used. So would be nice to be able to disable the adding of them to reduce the network size...

PR #891, opened 2024-01-24
Don't include triple solar potential option by default

koen-vg:

A trivial change in default `{sector_opts}` wildcard. I know that it's very much the intention that users will inform themselves about each wildcard and option and change them to fit their needs. However, defaults do carry some weight, and arguably the `solar+p3` option was a sensible default. After a0f43aee, however, this is no longer the case! I would remove the option from the default `{sector_opts}` wildcard to avoid confusion.

## Checklist

### DIFF
--- a/config/config.default.yaml
+++ b/config/config.default.yaml
@@ -44,7 +44,7 @@ scenario:
   opts:
   - ''
   sector_opts:
-  - Co2L0-3H-T-H-B-I-A-solar+p3-dist1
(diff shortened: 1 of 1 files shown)


fneum:
Agreed, I missed this. Thanks!

koen-vg:
Actually I do now see that this was maybe a bit rushed and there are more references to `solar+p3` in other configs as well as mock snakemake sections. Would you want me to remove those as well for consistency?

fneum:
I can do it, no worries

PR #893, opened 2024-01-24
Avoid the use use of `workflow.source_path`

koen-vg:
The above function returns a path to a cached source file, which changes for every run, causing the resulting parameter to change and the solve_network rule to re-run every time. The problem was introduced in https://github.com/PyPSA/pypsa-eur/commit/d145758fb7ff4bf126ddada8eec6d8f942c93f4f

I have tested the solution, and also tested to make sure this works when pypsa-eur is used as a module.

## Checklist

### DIFF
--- a/rules/common.smk
+++ b/rules/common.smk
@@ -43,7 +43,7 @@ def memory(w):
 def input_custom_extra_functionality(w):
     path = config["solving"]["options"].get("custom_extra_functionality", False)
     if path:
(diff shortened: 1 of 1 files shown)


fneum (approved):
Oh, interesting! This is a big downside of using `workflow.source_path()` for any place (input, param) and probably not intended by the `snakemake` developers. Thanks for finding the issue and providing the fix!

PR #898, opened 2024-01-25
Add several map projection options

koen-vg:
It's quite petty but being based in the north of Norway I can't really deal with the default EqualEarth projections and its wild distortions. I say give the user some (limited) choice in projection without them having to do surgery on the `plot_network.py` script. Do feel free to discard (I will just keep it for personal use) if this kind of "feature creep" isn't really desired.

Added a `config["plotting"]["projection"]` option with a few different preset choices. Falls back to the old default of EqualEarth if not specified.

## Checklist

### DIFF
--- a/config/config.default.yaml
+++ b/config/config.default.yaml
@@ -773,7 +773,13 @@ plotting:
     color_geomap:
       ocean: white
       land: white
-  projection: "EqualEarth"
+  projection:
+    name: "EqualEarth"
+    # See https://scitools.org.uk/cartopy/docs/latest/reference/projections.html for alternatives, for example:
+    # name: "LambertConformal"
+    # central_longitude: 10.
+    # central_latitude: 50.
+    # standard_parallels: [35, 65]
   eu_node_location:
     x: -5.5
     y: 46.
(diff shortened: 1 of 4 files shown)


fneum:
I like it!

I was thinking whether one could give even more flexibility to the user rather than just 4 options:

```yaml
plotting:
  projection:
    name: Orthographic
    # kwargs
    central_longitude: 10.
    central_latitude: 50.
```

```py
proj_kwargs = snakemake.params.plotting.get("projection", dict(name="EqualEarth"))
proj_func = getattr(ccrs, proj_kwargs.pop("name"))
proj = proj_func(**proj_dict)
```

The `config.yaml` would default to:

```yaml
plotting:
  projection:
    name: EqualEarth
```

Then, one could just document a few recommended options that are currently explicit in the code.

koen-vg:
Sure, I like it! I just haven't really gotten comfortable with the whole `getattr` thing yet, so I didn't really consider this :)

For now I've given one reasonable alternative example (commented) in config.default.yaml, along with a link to the complete list of projections. I feel like this should provide a useful starting point for the user, but feel free to move things or let me know if you'd for instance rather document alternatives in `configtables/plotting.csv` or so instead.

fneum (approved):
Great!

Issue #896, opened 2024-01-24
Error plotting CH4 network due to link_width argument containing more than gas pipeline links


koen-vg:
## Checklist

- [x] I am using the current [`master`](https://github.com/PyPSA/pypsa-eur/tree/master) branch or the latest [release](https://github.com/PyPSA/pypsa-eur/releases). Please indicate.
- [x] I am running on an up-to-date [`pypsa-eur` environment](https://github.com/PyPSA/pypsa-eur/blob/master/envs/environment.yaml). Update via `conda env update -f envs/environment.yaml`.

## Describe the Bug

I've been having this fairly fleeting issue; somehow I can make it happen sometime, but not always? I'm quite confused about this. See error message below.

Anyway; the following does seem to fix it:

```diff
modified   scripts/plot_network.py
@@ -544,7 +544,7 @@ def plot_ch4_map(network):
     link_widths_orig = n.links.p_nom / linewidth_factor
     link_widths_orig[n.links.p_nom < line_lower_threshold] = 0.0
 
-    max_usage = n.links_t.p0.abs().max(axis=0)
+    max_usage = n.links_t.p0[n.links.index].abs().max(axis=0)
     link_widths_used = max_usage / linewidth_factor
     link_widths_used[max_usage < line_lower_threshold] = 0.0
 
```
The problem is that non-gas pipeline links are dropped from `n.links` earlier, but not from `n.links_t`. Has anyone had this before? Maybe it's worth just putting the above change in?

## Error Message

```
[Wed Jan 24 16:00:18 2024]
rule plot_network:
    input: results/[...]/postnetworks/elec_s_50_lv1.5__Co2L0_2050.nc, resources/[...]/regions_onshore_elec_s_50.geojson
    output: results/[...]/maps/elec_s_50_lv1.5__Co2L0-costs-all_2050.pdf, results/[...]/maps/elec_s_50_lv1.5__Co2L0_2050-today.pdf
    jobid: 61
    benchmark: benchmarks/[...]/plot_network/elec_s_50_lv1.5__Co2L0_2050
    reason: Code has changed since last execution
    wildcards: simpl=, clusters=50, ll=v1.5, opts=, sector_opts=Co2L0, planning_horizons=2050
    threads: 2
    resources: tmpdir=/tmp, mem_mb=10000, mem_mib=9537

Activating conda environment: .snakemake/conda/0d7d55fa83204a765b0aba3c9ff3a1a3_
[...]

PR #900, opened 2024-01-25
Split rule `plot_network` into power, hydrogen, gas

fneum:
This makes the code more modular and simplifies the code.

The long script `plot_network.py` became increasingly hard to handle.


fneum:
closes #896 via b73c614f

PR #901, opened 2024-01-26
prepare_sector_network: reorder buses for dac components

FabianHofmann:
Change the order of buses (bus0, bus1, ...) for DAC components to meet the convention of the other components: `bus0` refers to the electricity bus (input), `bus1` to the heat bus (input), 'bus2' to the CO2 atmosphere bus (input), and `bus3` to the CO2 storage bus (output).

Why is this helpful? 

For many routines in the pypsa statistics module, a convention where the `bus0` points to the location (in order to extract local balances) is assumed. With this change, the location of the DAC facilities is properly processed.  

## Checklist


scripts/prepare_sector_network.py
```diff
@@ -710,27 +710,27 @@ def add_dac(n, costs):
...
+        bus2="co2 atmosphere",
+        bus3=spatial.co2.df.loc[locations, "nodes"].values,
         carrier="DAC",
         capital_cost=costs.at["direct air capture", "fixed"],
-        efficiency=1.0,
-        efficiency2=efficiency2,
-        efficiency3=efficiency3,
+        efficiency=-heat_input / electricity_input,
+        efficiency2=-electricity_input,
+        efficiency3=electricity_input,
```
fneum, line 733: Shouldn't this be `1 / electricity_input`? The unit should be tCO2 / MWh_el.
```suggestion
        efficiency2=1 / -electricity_input,
        efficiency3=1 / electricity_input,
```

fneum (changes requested):
I think there is one mistake with the units / efficiencies. Could you double check?

scripts/prepare_sector_network.py
FabianHofmann: you are totally right!

FabianHofmann:
I made a small test case locally, this works as expected.

fneum (approved):

PR #909, opened 2024-02-01
fix plotting with H2_retrofit option

JulianGeis:
plot_network.py is currently not working with H2_retrofit option

- tested this fix for myopic with and without H2_retrofit


fneum:
superceded by #937

PR #916, opened 2024-02-06
Pandas deprecation fixes

koen-vg:
Just fixing some deprecation and future warnings as I go along :)

## Checklist

### DIFF
--- a/scripts/make_summary_perfect.py
+++ b/scripts/make_summary_perfect.py
@@ -246,8 +246,9 @@ def calculate_energy(n, label, energy):
                 .groupby(level=0)
                 .sum()
                 .multiply(c.df.sign)
-                .groupby(c.df.carrier, axis=1)
+                .T.groupby(c.df.carrier)
                 .sum()
+                .T
             )
         else:
             c_energies = pd.DataFrame(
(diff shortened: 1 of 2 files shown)


koen-vg:
Ah, this time the tests saved me; I was a little too eager with just one of the fixes.

scripts/make_summary_perfect.py
```diff
@@ -514,9 +508,7 @@ def calculate_weighted_prices(n, label, weighted_prices):
             if names.empty:
                 continue
 
-            load += (
-                n.links_t.p0[names].groupby(n.links.loc[names, "bus0"], axis=1).sum()
-            )
+            load += n.links_t.p0[names].T.groupby(n.links.loc[names, "bus0"]).sum()
```
fneum, line 511: Just double-checking the transposes here:

So far, in most cases `.groupby(..., axis=1).sum()` is replaced by `.T.groupby(...).sum().T`

scripts/make_summary_perfect.py
```diff
@@ -652,7 +644,7 @@ def calculate_co2_emissions(n, label, df):
         emitted = n.generators_t.p[gens.index].mul(em_pu)
 
         emitted_grouped = (
-            emitted.groupby(level=0).sum().groupby(n.generators.carrier, axis=1).sum().T
+            emitted.groupby(level=0).sum().T.groupby(n.generators.carrier).sum().T
```
fneum, line 647: similar here, given the additional transpose

fneum (commented):
Looks good! Thanks! Just unsure about two cases that I couldn't verify without testing it myself.

koen-vg:
Yes you were right! At least the one I was able to test, in the weighted price calculation. It turned out the code path wasn't used before due to an unrelated bug, which I now also fixed. I will say that the whole function seems pretty fragile...

For the CO2 emissions calculations, the code path you pointed out is also never taken! Hence why I didn't find the mistake. It turns out that in sector-coupled perfect foresight networks there aren't any co2-emitting generators anyway; all emissions go the the co2 atmosphere store.

fneum:
Great! Thanks for looking into it and sorry that the script is so buggy...

fneum:
test failures seem unrelated, I will merge given previous commit CI ran through.

koen-vg:
No worries, happy to help and anyway it's so much better to have anything at all to get started with than having to do everything scratch!

PR #922, opened 2024-02-09
create a bus for every unique coordinate, not only substations

### LINKED ISSUE #699: Ensure assignment of all regions without substation
## Checklist

- [x] I am using the latest [release](https://github.com/PyPSA/pypsa-eur/releases)

## Describe the Bug

While building bus regions in `build_bus_regions` some onshore bus regions are filtered out for not having a substation. Fore some clusterings this leads to missing regions later on. They are, however, referenced for bus mapping since they still appear in the network.
This might be the cause for some issues with focus_weights.

Detail:
Here some regions for clustered onshore regions disappear:
https://github.com/PyPSA/pypsa-eur/blob/03e1ed6e7914e9e1c5055b7efc06d64425e2bb49/scripts/cluster_network.py#L437-L443

this is due to some regions that don't have a substation being filtered out here:
https://github.com/PyPSA/pypsa-eur/blob/03e1ed6e7914e9e1c5055b7efc06d64425e2bb49/scripts/build_bus_regions.py#L138

fneum:
closes #699

### DIFF
--- a/doc/release_notes.rst
+++ b/doc/release_notes.rst
@@ -10,6 +10,11 @@ Release Notes
 Upcoming Release
 ================
 
+* Regions are assigned to all buses with unique coordinates in the network with
+  a preference given to substations. Previously, only substations had assigned
+  regions, but this could lead to issues when a high spatial resolution was
+  applied.
+
 * The default configuration ``config/config.default.yaml`` is now automatically
   used as a base configuration file and no longer copied to
   ``config/config.yaml`` on first use. The file ``config/config.yaml`` should be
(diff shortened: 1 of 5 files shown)


martacki (approved):
Nice, also did some plots at `elec.nc`, and compared to before, now every (visible) node has a Voronoi region.

Sometimes (in case of the whole EUR network without UA/MD, it happens for 6 cases), locations that are intended to have the same geo-coordinates (but have a very small deviation in 4th decimal digit or so) get two separate regions. But this is resolved after `simplify_network` (after mapping the network to one voltage level where these nodes are aggregated). So not really worth to add some rounding accuracy, I'd say. After `simplify_network`, this case does not appear anymore.

PR #928, opened 2024-02-13
cluster nice names as well

lindnemi:
I noticed that in the myopic optimisation for the pypsa-ariadne project, the carriers returned by 
```python
n.statistics.withdrawal(
        bus_carrier="rural heat", 
        groupby=n.statistics.groupers.get_name_bus_and_carrier,
    ).filter(
        like="DE",
        axis=0,
    ).groupby("carrier").sum()
 ```
will differ between the 2020 network and the 2030 network. In the 2020 baseyear a carrier will, e.g., be named `residential urban decentral heat` , whereas in the 2030 network it will  simply be ` urban decentral heat `. In both cases all links in the network actually have `urban decentral heat` as carrier - it's just the names that are different.

I traced the bug to the `n.carriers.nice_name` column, on which n.statistics relies. 

image

It's clearly visible that the name's didn't get updated correctly alongside the carrier. It's a bit strange that this occurs only for 2020 and not for 2030, and is probably related to how `cluster_heat_buses` is called. It happens either in `prepare_sector_network` (for later years) or in `add_existing_baseyear` (for 2020). In all cases the `sanitize_carriers` function is called, but fails to correctly rename the clustered heat carriers. Instead of looking into the details of why `sanitized_carriers` fails, the easy fix that i propose is to rename the `nice_name` column directly in `cluster_heat_buses`.

rename the `nice_name` column directly in `cluster_heat_buses`

## Checklist

### DIFF
--- a/rules/postprocess.smk
+++ b/rules/postprocess.smk
@@ -45,7 +45,7 @@ if config["foresight"] != "perfect":
             (
                 LOGS
                 + "plot_power_network/elec_s{simpl}_{clusters}_l{ll}_{opts}_{sector_opts}_{planning_horizons}.log"
-            )
+            ),
         benchmark:
             (
                 BENCHMARKS
@@ -74,7 +74,7 @@ if config["foresight"] != "perfect":
             (
                 LOGS
                 + "plot_hydrogen_network/elec_s{simpl}_{clusters}_l{ll}_{opts}_{sector_opts}_{planning_horizons}.log"
-            )
+            ),
         benchmark:
             (
                 BENCHMARKS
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

**Comment by lindnemi** [comment]
@lisazeyen @p-glaum

**Comment by p-glaum** [comment]
Hi @lindnemi, to fix this, try to use 
```
n.statistics.withdrawal(
        bus_carrier="rural heat", 
        groupby=n.statistics.groupers.get_name_bus_and_carrier,
        **nice_names=False**
    )
```

lisazeyen:
We still can modify the `cluster_heat_buses` function as @lindnemi suggested to get the correct `nice_names`. @lindnemi can you make sure that only the function is changed in the PR and not other files (like e.g. `build_ammonia_production.py`)? Then we can pull it in.

lindnemi:
> We still can modify the `cluster_heat_buses` function as @lindnemi suggested to get the correct `nice_names`. @lindnemi can you make sure that only the function is changed in the PR and not other files (like e.g. `build_ammonia_production.py`)? Then we can pull it in.

The other changes are by the pre-commit.ci... Do I just revert them, or should we keep them?

PR #933, opened 2024-02-15
update h2 pipe losses

lisazeyen:
Closes # (if applicable).
- in the German Kernnetz we have planned pipe capacities ranging between 1.2-17 GW_H2 with 70 bars
- [DEA](https://ens.dk/en/our-services/projections-and-models/technology-data/technology-data-transport-energy) (sheet name H2 70) assumes total pipe losses for this pipe size of 1.3-2.2%/1000 km (losses are higher with higher operating pressure).

I think that includes the energy demand of the compressors, which is in DEA data ranging between 0.7-0.9% for pipes operating at 70 bar.

I would suggest taking the mean value, so 0.8% energy demand for the compressors and 1.75%/1000 km total pipe losses. I subtracted in the PR now for the pipe losses without the compressors 100-(1.75-0.8)= 99.05 -> 0.9905
Do we assume that the compressors are installed

## Checklist

### DIFF
--- a/config/config.default.yaml
+++ b/config/config.default.yaml
@@ -538,8 +538,8 @@ sector:
       efficiency_static: 0.98
       efficiency_per_1000km: 0.977
     H2 pipeline:
-      efficiency_per_1000km: 1 # 0.979
-      compression_per_1000km: 0.019
+      efficiency_per_1000km: 0.996
+      compression_per_1000km: 0.018
     gas pipeline:
       efficiency_per_1000km: 1 #0.977
       compression_per_1000km: 0.01
--- a/config/config.default.yaml
+++ b/config/config.default.yaml
@@ -538,8 +538,8 @@ sector:
       efficiency_static: 0.98
       efficiency_per_1000km: 0.977
     H2 pipeline:
-      efficiency_per_1000km: 0.9905
-      compression_per_1000km: 0.008
+      efficiency_per_1000km: 1 # 0.982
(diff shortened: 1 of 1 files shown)


fneum:
I think all of it refers to the compressor:

image

https://ens.dk/sites/ens.dk/files/Analyser/technology_data_for_energy_transport.pdf page 133

So maybe just 1 - 0.018 = 0.982 for the compression demand.

fneum:
Ah, and the "Energy demand, compressor filling station %" is only for the entry into the pipeline network. I think we should ignore this one because pipelines can pass through regions. Proper modelling would distinguish H2 bus and H2 pipeline bus with filling compressor link connecting the two.

We assume compressor cost in the investment cost of the pipelines.

config/config.default.yaml
```diff
@@ -538,8 +538,8 @@ sector:
       efficiency_static: 0.98
       efficiency_per_1000km: 0.977
     H2 pipeline:
-      efficiency_per_1000km: 1 # 0.979
-      compression_per_1000km: 0.019
+      efficiency_per_1000km: 0.9905
+      compression_per_1000km: 0.008
```
fneum, line 542:
```suggestion
      efficiency_per_1000km: 1 # 0.982
      compression_per_1000km: 0.018
```

fneum: Yes, I did exactly that. 1.8%

PR #947, opened 2024-02-23
Eurostat 2023 data for energy totals

toniseibold:

Reads in the latest eurostat report from 2023 and uses it for building energy totals. 
Since the JRC IDEES data is only available for 2015, the energy balances are scaled by the factor of eurostat[report_year]/eurostat[2015] data as well.
The latest [swiss energy totals](https://www.bfe.admin.ch/bfe/de/home/versorgung/statistik-und-geodaten/energiestatistiken/energieverbrauch-nach-verwendungszweck.html/) are added to ensure consistent energy totals.

To complement the PR, please add the [eurostat data from 2023](https://ec.europa.eu/eurostat/web/energy/database/additional-data#Energy%20balances) to the zenodo data.

Missing is only the passenger car efficiency and the number of passenger cars that are initially in the JRC IDEES but not in the eurostat data set.

Comparison between 2023 and 2016 reports for the year 2013 show no difference except for Norway and Switzerland.
[energy_totals_comparison.csv](https://github.com/PyPSA/pypsa-eur/files/14419679/energy_totals_comparison.csv)

## Checklist


lisazeyen:
@toniseibold great! Just a few remarks:

1. in the long-term we would like to move away from the Zenodo sector-bundle repository. Can you add to the `retrieve_sector_databundle.py` some lines which download the necessary files from Eurostat directly?
2. The new data set is still also containing the previous years and the UK, so there is no need to keep the old ones, right? This simplifies the script, you can use as an input only the new one an remove the `if-else` statement depending on which Eurostat input you use.
3. Can you do some comparison of the output files between previous Eurostat and the new?

scripts/build_energy_totals.py
```diff
@@ -669,31 +750,53 @@ def build_eurostat_co2(input_eurostat, countries, report_year, year=1990):
...
-            "international navigation": (ct, "-", "Bunkers"),
-            "domestic aviation": (ct, "+", "+", "Domestic aviation"),
-            "international aviation": (ct, "+", "+", "International aviation"),
-            # does not include industrial process emissions or fuel processing/refining
-            "industrial non-elec": (ct, "+", "Industry"),
-            # does not include non-energy emissions
-            "agriculture": (eurostat_co2.index.get_level_values(0) == ct)
-            & eurostat_co2.index.isin(["Agriculture / Forestry", "Fishing"], level=3),
-        }
+        if report_year != 2023:
```
FabianHofmann, line 757: perhaps avoid the code duplication here? only adjust/overwrite differing keys in one if-clause

scripts/build_energy_totals.py
```diff
@@ -737,6 +840,140 @@ def build_transport_data(countries, population, idees):
...
+
+        avi_d = ratio.loc[(country, slice(None), "Domestic aviation")]
+        avi_i = ratio.loc[(country, "International aviation", slice(None))]
+        energy.loc[country, avia_inter] *= avi_i[["total"]].iloc[0, 0]
+        energy.loc[country, avia_domestic] *= avi_d[["total"]].iloc[0, 0]
+
+        nav = ratio.loc[(country, slice(None), "Domestic Navigation")]
+        energy.loc[country, navigation] *= nav[["total"]].iloc[0, 0]
+
+    return energy
```
FabianHofmann, line 974: If you would want to have this more compact, you could also use such a structure:

```python
def rescale_energy(idees_countries, energy, eurostat):
    # ... eurostat calculations ...  

    country_mapping = {"GB": "UK", "GR": "EL"}
    idees_countries.replace(country_mapping, inplace=True)

    mappings = {  # Define your mappings here
        "residential": {
           "total": ["total residential space", ...],
           "ele": ["electricity residential space", ...]
        }, 
        "services": { ... },
        # ...
    }

    for country in idees_countries:
        for sector, mapping in mappings.items():
            sector_ratio = ratio.loc[(country, slice(None), sector)]

            # Vectorized scaling
            energy.loc[country, mapping['total']] *= sector_ratio['total'].iloc[0, 0]
            energy.loc[country, mapping['ele']] *= sector_ratio['ele'].iloc[0, 0]

    return energy
```

FabianHofmann (commented):
just some stylistic comments

FabianHofmann:
> 1. in the long-term we would like to move away from the Zenodo sector-bundle repository. Can you add to the `retrieve_sector_databundle.py` some lines which download the necessary files from Eurostat directly?

I think, a separate rule `retrieve_eurostat` would be better to have a clear distinction

scripts/build_energy_totals.py
toniseibold: Since now only the eurostat data from 2023 is used, the whole if-clause is obsolete

scripts/build_energy_totals.py
```diff
@@ -759,12 +996,24 @@ def build_transport_data(countries, population, idees):
     input_eurostat = snakemake.input.eurostat
     eurostat = build_eurostat(input_eurostat, countries, report_year, data_year)
     swiss = build_swiss(data_year)
-    idees = build_idees(idees_countries, data_year)
+    # data from idees only exists for 2015
+    if data_year > 2015:
+        # read in latest data and rescale later
+        idees = build_idees(idees_countries, 2015)
+    else:
+        idees = build_idees(idees_countries, data_year)
```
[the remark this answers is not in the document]
toniseibold: Most energy system optimizations are running for the base year 2013. Since there is JRC IDEES data for that specific year, I want to keep it

scripts/build_energy_totals.py
toniseibold: Thanks for the tip :)

scripts/build_energy_totals.py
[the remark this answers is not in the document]
FabianHofmann: but then the comment `data from idees only exists for 2015` is wrong or?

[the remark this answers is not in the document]
toniseibold: Yes it's misleading
The report is from 2015 and they are not continuing the reporting
However the report contains data from 2000-2015

lisazeyen:
@toniseibold
 Have you done some comparison of the output files between previous Eurostat and the new? Can you plot it here or confirm that the outputs are similar? The CI is failing for the script `build_energy_totals.py` can you make sure that it is running through? It seems to be connected with the swiss data and the year `2019`

fneum (approved):

fneum:
Checked output of `energy_totals.csv` before and after and value changes are in an acceptable range.

PR #952, opened 2024-02-28
Add check for turning off transmission expnasion if limit reached

koen-vg:

I recently ran into an issue with myopic optimisations where if the transmission expansion limit has been reached in one time horizon, this could run into model infeasibility when solving at the next time horizon. What happened was that the minimum transmission network capacity given by `s_nom_min` and `p_nom_min` combined equalled the global limit set in `n.global_constraints` up to 12 significant digits. But with the lower limit through `*_nom_min` so close to the upper limit given by the global constraint, the model ran into numerical problems and was declared infeasible.

I think the "morally correct" approach is to check of transmission is already at the limit, and if so, fix the capacities at those given by `*_nom_min`, make them non-extendable and remove the global limit. That's what I've implemented in this PR.

The implementation seems to be working fine on my side, though I haven't gotten around to testing it yet for the volume limit, only the cost limit. However, do let me know if anyone can think of any potential issues or better ways of dealing with this.

## Checklist


fneum:
Sounds familiar. We also came across this issue for the volume limit recently and fixed it very similarly in `add_brownfield.py`.

https://github.com/PyPSA/pypsa-eur/blob/16e2286188d25e40056296fa979c59564ae45887/scripts/add_brownfield.py#L127-L149

I think this is the slightly better place to handle this since it only affects the myopic optimization.

Maybe you can expand this function to also cover the cost limit case? It should just be adding a `"capital_cost * "` prefix to the calculation if I am not wrong. What I liked about your PR is that it goes by global constraint type, not just the index name.

koen-vg:
Right, somehow I missed that this already existed but just for the volume limit! I've just force-pushed a sort of combination; you can see if you like it. Probably the for loop over global constraints of a particular type is slightly unnecessary but I don't think it hurts just in case. I think when checking where two things are numerically "close" it's better to normalise first so that this works for models of different scales.

fneum (approved):

PR #956, opened 2024-03-02
adding regions to network.shapes

drifter089:
- Data has been transferred from `cluster_network.py` to `n.shapes` in a GeoPandas DataFrame.

The copied data appears as follows: 
Screenshot from 2024-03-02 21-56-35

For the file: 
Screenshot from 2024-03-02 21-58-07

The rule `build_bus_regions` uses `build_bus_regions.py` to produce `regions_offshore.geojson`, which is subsequently transformed into:

- `regions_offshore_elec_s{simpl}.geojson`
- `regions_offshore_elec_s{simpl}_{clusters}.geojson`

using `cluster_network.py`.

- I have copied data into `cluster_network.py`.
- Should we consider transferring data into `build_bus_regions.py` or earlier, remove `.geojson` files, and perform all subsequent operations in the `n.shapes` GeoPandas DataFrame?
- Or, should we maintain the functionality of transferring data from `regions_offshore_elec_s{simpl}_{clusters}.geojson` separately?
- Additionally, we need to determine if, when approach 1 is implemented, the data from `regions_offshore_elec_s{simpl}.geojson` is required by any other rule once the subsequent file (e.g., `regions_offshore_elec_s{simpl}_{clusters}.geojson`) is created. If operations are conducted in a GeoPandas DataFrame, the preceding information will not be available after transformations.

Please suggest how we should proceed.

### DIFF
--- a/scripts/cluster_network.py
+++ b/scripts/cluster_network.py
@@ -431,7 +431,7 @@ def clustering_for_n_clusters(
 def cluster_regions(busmaps, input=None, output=None):
     busmap = reduce(lambda x, y: x.map(y), busmaps[1:], busmaps[0])
 
-    clustered_regions=[]
+    clustered_regions = []
     for which in ("regions_onshore", "regions_offshore"):
         regions = gpd.read_file(getattr(input, which))
         regions = regions.reindex(columns=["name", "geometry"]).set_index("name")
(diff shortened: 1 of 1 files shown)


drifter089:
@pz-max , @FabianHofmann 
please review :smile:

FabianHofmann:
@drifter089 do you mind if push to this PR directly? There is still some wrong assignments

fneum:
superseded by #1012

PR #958, opened 2024-03-05
compute lifetime after grouping DateIn

lindnemi:
Closes # (if applicable).

The problem is described in https://github.com/PyPSA/pypsa-ariadne/pull/29.

The lifetime is computed as DateOut - DateIn + 1. At the moment, the model considers not DateIn as start of the lifetime but the grouping year. The grouping year may be up to 5 years later than DateIn, and thus some power plants live up to 5 years longer than they should. 

Fixes the issue by computing lifetime based on the grouping year.

## Checklist

### DIFF
--- a/config/config.default.yaml
+++ b/config/config.default.yaml
@@ -354,7 +354,7 @@ solar_thermal:
 
 # docs in https://pypsa-eur.readthedocs.io/en/latest/configuration.html#existing-capacities
 existing_capacities:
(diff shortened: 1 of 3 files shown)


lindnemi:
these may be considered two separate things, but i also added grouping years 1960-1980 now. There are still some powerplants in operation from that time.

fneum (approved):

Issue #966, opened 2024-03-13
Correctly enforce existing capacities if build year >= base year


lindnemi:
## Describe the Bug

The baseyear in pypsa-ariadne is 2020. We want to enforce capacities that have been build in 2020 or afterwards.

At the moment this is done in `add_existing_baseyear.py`, just as if the build year was before 2020, by setting p_nom. 
However, after the optimisation, for some carriers, the optimal capacity is smaller than p_nom. This means that the already existing assets are not used by the model. Why?

1.  A constraint for exisiting capacities is added at https://github.com/PyPSA/pypsa-eur/blob/0a0a35e4a4d8f891be81a05648145a9d14c86dae/scripts/add_existing_baseyear.py#L248-L250

However, for renewable carriers  `p_nom_min` gets reset to `0`  in solve_network.py.  I would consider this a bug, since that way existing capacities get disregarded. The responsible lines are https://github.com/PyPSA/pypsa-eur/blob/0a0a35e4a4d8f891be81a05648145a9d14c86dae/scripts/solve_network.py#L100-L101 and https://github.com/PyPSA/pypsa-eur/blob/0a0a35e4a4d8f891be81a05648145a9d14c86dae/scripts/solve_network.py#L126-L127

Apparently the behaviour was first introduced in https://github.com/PyPSA/pypsa-eur/pull/728 to limit the land use of renewables.

2. When i revert [this PR](https://github.com/PyPSA/pypsa-eur/pull/728) the p_nom_lims are correctly carried forward. However,  then `p_nom_min` is still smaller than the `p_nom` for most carriers, which i understand to be indicative of the existing capacities. And thus, once again `p_nom_opt` is lower than it should be.

3. For offwind-ac the situation is actually the opposite and p_nom is smaller than p_nom_min. This might be related to the issue that existing wind capacities get added without a suffix and are thus overlooked

## How to Fix

Basically two approaches could work:

1. set p_nom_min = p_nom

2. differentiate between enforced assets and optimized assets. then `carrier-year` would exist twice if build year >= base year.

(and also assign all already existing offwind capacities to offwind-ac and offwind-dc)

I am a bit reluctant to simply go for 1. since it might break other peoples code (??)

lindnemi:
[...]

PR #967, opened 2024-03-14
Reform existing grouping year logic

lindnemi:

At the moment, the grouping year for existing capacities is always bigger or equal to the build year. E.g., solar-2005 collects all solar built between 2000 and 2005. This is exactly the opposite of what's happening during the optimization. Here solar-2020 would collect all capacities built between 2020 and 2025. This PR changes the grouping year logic of existing capacities to match  those of newly built capacities.

Partially addresses https://github.com/PyPSA/pypsa-eur/issues/966.  To distinguish existing capacities in pypsa-ariadne, a simple fix would be to use the grouping years 2015 and 2021. This should avoid clashes between newly build capacities (e.g., solar-2020) and existing capacities added in and after the baseyear (then solar-2015 and solar-2021)

## Checklist


JulianGeis (approved):
Code looks good. I tested the outcome against the ariadne2 branch. No more addition of capacities from the future. The capacities are reduced for the generators.

lindnemi:
IMO the capacities should match. we need to understand why they do not match before merging.

for later years capacities may be different, because for generators the lifetime is not averaged between grouping years. so whereas before some generators lived longer than their actual lifetime, now they live short. however, this is another issue that should be adressed separately

lindnemi:
https://github.com/PyPSA/pypsa-eur/blob/60fe95ddf7c47e30b28f2cac3061e5a6f9c59163/scripts/add_existing_baseyear.py#L298
Instead of reading always the same number from the costs data this should be a reasonable average, such as

https://github.com/PyPSA/pypsa-eur/blob/60fe95ddf7c47e30b28f2cac3061e5a6f9c59163/scripts/add_existing_baseyear.py#L351

fneum:
@lisazeyen could you have a final look?

lindnemi:
it would be nice to understand why the capacities before and after my changes  do not match before merging.

lindnemi:
> https://github.com/PyPSA/pypsa-eur/blob/60fe95ddf7c47e30b28f2cac3061e5a6f9c59163/scripts/add_existing_baseyear.py#L301
> 
> Instead of reading always the same number from the costs data this should be a reasonable average, such as
> 
> https://github.com/PyPSA/pypsa-eur/blob/60fe95ddf7c47e30b28f2cac3061e5a6f9c59163/scripts/add_existing_baseyear.py#L351

The lifetime should be weighted by capacity when aggregating.

lindnemi:
So.... whereas added heating capacities for e.g. gas boilers were about 2 % lower than existing heating capacities before this PR, now they end up 3% higher. Why?

lindnemi:
> So.... whereas added heating capacities for e.g. gas boilers were about 2 % lower than existing heating capacities before this PR, now they end up 3% higher. Why?

ok i get it now. For central gas boilers the efficiency is 1.03. So 100% of existing capacities gets translated into 1.03% of installed capacitiy. The 98% of the earlier version were wrong, with roughly 5% missing due to the incorrect implementation of `ratio`. So everything seems to work as it should with this PR.

scripts/add_existing_baseyear.py
```diff
@@ -434,12 +443,23 @@ def add_heating_capacities_installed_before_baseyear(
...
             efficiency = costs.at[costs_name, "efficiency"]
 
-        for i, grouping_year in enumerate(grouping_years):
-            if int(grouping_year) + default_lifetime <= int(baseyear):
-                continue
+        valid_grouping_years = pd.Series(
+            [
+                int(gy)
+                for gy in grouping_years
+                if int(baseyear) - default_lifetime <= int(gy) < int(baseyear)
```
lisazeyen: This changes the definition of the valid grouping years, e.g.
you have `baseyear=2020`, `default_lifetime=20`, `grouping_year=gy=2000`

Then before 2000 + 20 <= 2020 -> True -> would not be considered

Now: 2020-20 <= 2000 -> True would be considered

scripts/add_existing_baseyear.py
```diff
@@ -434,12 +443,24 @@ def add_heating_capacities_installed_before_baseyear(
...
             efficiency = costs.at[costs_name, "efficiency"]
 
-        for i, grouping_year in enumerate(grouping_years):
-            if int(grouping_year) + default_lifetime <= int(baseyear):
-                continue
+        valid_grouping_years = pd.Series(
+            [
+                int(grouping_year)
+                for grouping_year in grouping_years
+                if int(grouping_year) + default_lifetime > int(baseyear)
```
lindnemi, line 450: this should be >= imo

scripts/add_existing_baseyear.py
```diff
@@ -434,12 +443,22 @@ def add_heating_capacities_installed_before_baseyear(
...
+                and int(grouping_year) < int(baseyear)
+            ]
+        )
+        
+        # get number of years of each interval
+        _years = (valid_grouping_years.diff().shift(-1)
+                  .fillna(baseyear-valid_grouping_years.iloc[-1]))
+        # Installation is assumed to be linear for the past
+        ratios = _years / _years.sum()
```
lisazeyen, line 459: I think this definition is correct but changes compared to the previous definition of ratio using the lifetime
`ratio = (int(grouping_year) - int(grouping_years[i - 1])) / default_lifetime`

scripts/add_existing_baseyear.py
lisazeyen: to be consistent with previous definition, I think it shoud not.

Let's say `grouping_year=2000`, `baseyear=2020`, `default_lifetime=20`
Then before
`int(grouping_year) + default_lifetime <= int(baseyear)`
 2020 + 20 <= 2020 -> True -> not considered
Now with your sign ">="
2020 + 20 >= 2020 -> True -> considered as valid grouping year

So it should be `>`. This is also consistent with our definition that operation of an asset starts in 2020, with a lifetime of 20 years it should end operation then add the end of 2019.

scripts/add_existing_baseyear.py
lindnemi: i agree that this is correct. with the old definition only 95% of existing heating was added, which can be seen form the prenetwork-brownfield

PR #973, opened 2024-03-18
Fill in missing eurostat data on domestic aviation energy demand

koen-vg:

This fixes a nasty bug where `rescale_idees_from_eurostat` would result in infinite kerosene demand upon division by zero (obviously leading to infeasibility of the model, but no earlier error). I think filling in zeroes this way seems like the most reasonable way of dealing with the problem, but not 100% sure. At least it does work now.

## Checklist


fneum (approved):
Thanks! Great catch, great fix!

PR #978, opened 2024-03-19
snakefile: move copy_default_files and process_run_config to helpers

FabianHofmann:
In the current master, the CI on windows fails due to some weird syntax error (probably something wrong in snakemake >= 8, but hard to track). Took the opportunity and moved some snippets from the Snakefile header into dedicated functions.

### DIFF
--- a/Snakefile
+++ b/Snakefile
@@ -4,8 +4,8 @@
 
 from pathlib import Path
 import yaml
-from os.path import normpath
-from shutil import move, rmtree
+from os.path import normpath, exists
+from shutil import copyfile, move, rmtree
 from snakemake.utils import min_version
 
 min_version("8.5")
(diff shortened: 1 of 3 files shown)


Snakefile
```diff
@@ -2,44 +2,25 @@
 #
 # SPDX-License-Identifier: MIT
 
-from os.path import normpath, exists
-from shutil import copyfile, move, rmtree
```
fneum, line 6: needs to be imported in _helpers.py

FabianHofmann: it was the `move` function which is needed in some smk files (and the rmtree is used as well). Reverted this

FabianHofmann:
annoying, it all works but does not resolve the original problem with the indentation error on windows machines... still we can merge this

FabianHofmann:
I really don't know what's wrong with these windows machines... For a temporal fix, I removed windows from the CI requirement list. l

PR #980, opened 2024-03-19
Miscellaneous bugfixes

koen-vg:
I really do try to test my PRs, but somehow _two_ typos slipped through in the most recent ones; I'm sorry! Here is the fix.

To make up for it I found and fixed a mistake in the logic of filling in missing years of total hydro production data. I've tested this and it works :) Used to throw an error (cannot cast Series to boolean) but also the calculation of which years were missing was the wrong way around.

Hope it's fine it package these two fixes in one PR just to avoid overrunning you with PRs.

## Checklist

### DIFF
--- a/rules/retrieve.smk
+++ b/rules/retrieve.smk
@@ -115,7 +115,7 @@ if config["enable"]["retrieve"] and config["enable"].get(
             mem_mb=5000,
         retries: 2
         run:
-            move(input[0], output[0])
+            copyfile(input[0], output[0])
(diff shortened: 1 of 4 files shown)


koen-vg:
If you will allow me, I tacked one more one-line bugfix to this PR; the natura.tiff file needs to be copied to resources instead of moved since otherwise multiple processes can try to move the same file away, and all but one will fail.

fneum (approved):
No worries @koen-vg! I didn't catch it in review either.

PR #981, opened 2024-03-19
Use raw strings to avoid illegal backslash warnings python 12

koen-vg:
I swear this will be the last of these small boring PRs for a little while, sorry for the barrage.

It seems like snakemake and the whole workflow now works fine with python 12, although it's up to condas mood whether python 11 or 12 is installed. But python 12 does complain about the current use of backslashes in strings in the Snakefile; marking the relevant strings as raw strings removes the warnings.

## Checklist


fneum (approved):

FabianHofmann:
@koen-vg improvements are always welcome, even when they are small

PR #993, opened 2024-03-26
Check if scenario file exists before attempting to load

koen-vg:
A very simple quality of life improvement. Useful when running the `create_scenarios` rule the first time.

When wishing to generate a new scenarios files (say, "scenarios2.yaml"), one could change the corresponding config option (`run/scenarios/file`) and run the `create_scenarios` rule. However, if say "scenarios2.yaml" doesn't exist yet, the `get_scenarios` function fails. An awkward solution is to first create "scenarios2.yaml" manually (and put in some mock scenarios, otherwise `scenarios.keys()` fails!), and then run `create_scenarios`. With this check, that's not necessary anymore.

## Checklist


FabianHofmann:
makes sense!

PR #1012, opened 2024-04-10
Write shapes to .nc-files

cpschau:
In addition to the .geojson-files that are stored in the resources-directory, the shape files are stored inside the pypsa.Network-object and consequently stored as .nc-files.


FabianHofmann:
replaced by #1013

PR #1015, opened 2024-04-11
Increase default costs for gas storage and allam cycle

tgi-climact:

Increase the default capital cost for gas storage and the default marginal cost for allam cycle in `prepare_sector_network`.

For gas storage, the low capital cost of gas storage  leads to a huge amount of stores. To avoid this, an increase is proposed.

For allam cycle, low marginal cost is extremely in favour of allam cycle. To avoid undesirable behaviour, the use of the gas value is suggested.

## Checklist

### DIFF
--- a/scripts/prepare_sector_network.py
+++ b/scripts/prepare_sector_network.py
@@ -520,7 +520,8 @@ def add_carrier_buses(n, carrier, nodes=None):
 
     unit = "MWh_LHV" if carrier == "gas" else "MWh_th"
     # preliminary value for non-gas carriers to avoid zeros
-    capital_cost = costs.at["gas storage", "fixed"] if carrier == "gas" else 0.02
+    capital_cost = costs.at["gas storage", "fixed"] if carrier == "gas"\
+        else 200 * costs.at[carrier, "discount rate"]
 
     n.madd("Bus", nodes, location=location, carrier=carrier, unit=unit)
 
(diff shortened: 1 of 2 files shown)


scripts/prepare_sector_network.py
```diff
@@ -747,7 +748,7 @@ def add_allam(n, costs):
         p_nom_extendable=True,
         # TODO: add costs to technology-data
         capital_cost=0.6 * 1.5e6 * 0.1,  # efficiency * EUR/MW * annuity
-        marginal_cost=2,
```
fneum, line 750: Drawing from the gas bus already covers the gas fuel cost. In general, power plants implemented as links do not include fuel cost in their `marginal_cost`. The 2 €/MWhth is only the VOM.

scripts/prepare_sector_network.py
```diff
@@ -520,7 +520,8 @@ def add_carrier_buses(n, carrier, nodes=None):
 
     unit = "MWh_LHV" if carrier == "gas" else "MWh_th"
     # preliminary value for non-gas carriers to avoid zeros
-    capital_cost = costs.at["gas storage", "fixed"] if carrier == "gas" else 0.02
```
fneum, line 523: Not sure I understand the connection to the discount rate here, but 14 €/MWh/a is much too high for liquid fuels.

fneum (changes requested):

scripts/prepare_sector_network.py
tgi-climact: Makes perfect sense. Thank you !

PR #756, opened 2023-10-11
Add rule to update IRENA renewables capacities

tgi-climact:
This PR add an automated rule to retrieve data from IRENA. This was useful to get updated values for renewables capacities. This PR has updated values up to 2022 (previously 2018).

## Checklist

- [x] I tested my contribution locally and it seems to work fine.
- [x] Code and workflow changes are sufficiently documented.
- [x] Changed dependencies are added to `envs/environment.yaml`.
- [x] Changes in configuration options are added in all of `config.default.yaml`.
- [x] Changes in configuration options are also documented in `doc/configtables/*.csv`.
- [x] A release note `doc/release_notes.rst` is added.

PR #1018, opened 2024-04-11
Use powerplantmatching IRENASTAT for renewable capacities in `add_existing_baseyear`

koen-vg:

I wanted to update existing renewable capacities to 2023 values recently made available by IRENA. But there seems to be some duplicated effort on this front within the PyPSA umbrella: powerplantmatching has information of renewable capacities from IRENA, and the same data is downloaded through a separate path in `retrieve_irena.py`.

As nice of an improvement https://github.com/PyPSA/pypsa-eur/pull/756 is, I would propose cutting out `retrieve_irena.py` altogether and using the IRENA data retrieved by powerplantmatching directly. This is implemented in the current PR, achieving the same functionality as before with much less code all in all.

To avoid a regression in terms of data for recent years, this PR should only be merged once https://github.com/PyPSA/powerplantmatching/pull/157 is merged, and then should be combined with a version bump of the powerplantmatching dependency.

In the future, data on existing renewable capacities can be updated simply by updating powerplantmatching and bumping the version in pypsa-eur. This will then transparently filter through to _both_ places where existing renewable capacities are added to network in pypsa-eur, namely in `add_electricity.py` _and_ in `add_existing_baseyear.py`. Having these two scripts depend on different sources for the same data seems like a source of confusion and potential problems.

## Checklist


fneum:
I agree, @koen-vg. The link in `retrieve_irena.py` also seems to be broken, so this is actually quite timely.

fneum:
https://github.com/PyPSA/powerplantmatching/pull/157 is merged and included in v0.5.13 of powerplantmatching

fneum (approved):
LGTM, only requires powerplantmatching version bump in environment.yaml

koen-vg:
Great, easy enough to bump the version :)

koen-vg:
Guess it wasn't that easy, haha, powerplantmatching 0.5.13 will need a little bit of time until it's installable I guess.

koen-vg:
It does now seem to work.

PR #1023, opened 2024-04-12
Improve agg_p_nom_limits configuration

tgi-climact:

Improved the behaviour of `agg_p_nom_limits`:

  - Moved the associated configuration to `solving`. This allows *Snakemake* to correctly decide which rules to run when the configuration changes.

  - Added the ability to enable aggregation of all *offwind* types (*offwind-ac* and *offwind-dc*) when writing the constraint.

  - Added the possibility to take existing capacities into account when writing the constraint.

  - Added the possibility to have a different file for each planning horizon.

## Checklist


config/config.default.yaml
```diff
@@ -758,6 +757,14 @@ solving:
     linearized_unit_commitment: true
     horizon: 365
 
+  agg_p_nom_limits:
+    agg_offwind: false
+    include_existing: false
+    years:
+      2030: data/agg_p_nom_minmax.csv
+      2040: data/agg_p_nom_minmax.csv
+      2050: data/agg_p_nom_minmax.csv
```
FabianHofmann, line 766: thanks for the contribution. I am but I see that the functionality is needed. perhaps we can add columns to the data file for different years instead?

tgi-climact: @FabianHofmann I've pushed an update.

tgi-climact:
@FabianHofmann Would it be useful to add some code to extend this feature to links ? e.g. write constraint into `agg_p_nom_minmax.csv`for nuclear

fneum (approved):

fneum:
@tgi-climact all looks good!

Yes, I think it makes sense to extend this feature to links. I'll merge this PR first, and then we can extend.

tgi-climact:
> @tgi-climact all looks good!
> 
> Yes, I think it makes sense to extend this feature to links. I'll merge this PR first, and then we can extend.

Great, I will propose a new PR when I have some time.

PR #1025, opened 2024-04-14
Update energy balance for residential based on new Eurostat data

yerbol-akhmetov:
Good day. Here, I propose to update the residential part of the `energy_totals` based on disaggregated Eurostat's household data. The provided function updated `total residential`, `total residential space`, `total residential water`, and `total residential cooking` based on the data in https://ec.europa.eu/eurostat/databrowser/view/nrg_d_hhq/default/table?lang=en.

## Checklist


fneum:
Thanks @yerbol-akhmetov! Let us know when this is ready to review.

In the meantime. Would it be possible to do this with the currently used eurostat April 2023 version or alternatively combine it with an update of the whole workflow to eurostat February 2024 version? I'd avoid mixing the versions.

yerbol-akhmetov:
> Thanks @yerbol-akhmetov! Let us know when this is ready to review.
> 
> In the meantime. Would it be possible to do this with the currently used eurostat April 2023 version or alternatively combine it with an update of the whole workflow to eurostat February 2024 version? I'd avoid mixing the versions.

Thanks for the comment, @fneum. Maybe I interpret it wrongly, there is no versioning in February 2024 data. It was just lastly updated at that time. However, I also found the Eurostat March 2024 version for energy balances (the same format as in April 2023). But does it make sense to update 2023 data to 2024, since majority of the data are at past and the same? The only difference is that 2024 data contains 2022's energy balances.

yerbol-akhmetov:
> Thanks @yerbol-akhmetov! Let us know when this is ready to review.
> 
> In the meantime. Would it be possible to do this with the currently used eurostat April 2023 version or alternatively combine it with an update of the whole workflow to eurostat February 2024 version? I'd avoid mixing the versions.

Otherwise, it would be great if you can review it.

fneum:
Thanks, @yerbol-akhmetov! It looked mostly good. I took the liberty of extending this to the electricity shares as well to ensure consistency between total and electricity and simplify the code a bit.

yerbol-akhmetov:
> Thanks, @yerbol-akhmetov! It looked mostly good. I took the liberty of extending this to the electricity shares as well to ensure consistency between total and electricity and simplify the code a bit.
> 
> see 703b206e

Super, thanks for handling it.

Issue #843, opened 2024-01-03
`remove_stubs: false` not supported


koen-vg:
## Checklist

- [x] I am using the current [`master`](https://github.com/PyPSA/pypsa-eur/tree/master) branch or the latest [release](https://github.com/PyPSA/pypsa-eur/releases). Please indicate.
- [x] I am running on an up-to-date [`pypsa-eur` environment](https://github.com/PyPSA/pypsa-eur/blob/master/envs/environment.yaml). Update via `conda env update -f envs/environment.yaml`.

## Describe the Bug

It must have a been a while ago that the `clustering["simplify_network"]["remove_stubs"]` configuration option actually worked. Right now, when set to false, the workflow actually crashes because the `connection_costs` output of the `simplify_network` rule is never produced. (This output is written by the `_adjust_capital_costs_using_connection_costs` function, which, slightly indirectly, is only run when the `remove_stubs` option is enabled.)

Moreover, when I was playing around with the option, I found several problems later in the workflow too for sector-coupled networks, apparently mostly related to DC offshore wind buses that are stubs and usually merged with onshore AC buses, but lead to problems when `remove_stubs: false`. This is the case only for certain countries (Germany, Sweden, possibly more).

One option would be to remove the choice altogether and just remove the configuration option.

On the other hand, not removing stubs actually helps quite a bit with the spatial representation of Northern Norway, which is the reason for me to look into the option in the first place. Right now I sort of hacked together a solution where stubs are removed everywhere but in Norway, and this basically works for me.

I'm not exactly sure what the best solution for the project overall is. If nobody else is interested in having networks with stubs, it's probably best to just remove the configuration option. If it's a clear ambition that stubs should be supported, however, the lack of functionality should at least be mentioned in the documentation.

koen-vg:
Just to illustrate what I mean, here the topology of a network I'm currently working with. When stubs are removed, every node north of Lofoten (i.e. all the upper "tree-like" network extensions) are merged into more southern nodes, unfortunately removing any kind of spatial resolution in the representation of wind capacity factors, etc.
![bilde](https://github.com/PyPSA/pypsa-eur/assets/74298901/5efc91cf-ba45-4bcd-b32f-5dc55a7bb77e)

PR #1031, opened 2024-04-16
Remove connection costs output

martacki:

Suggest to remove the snakemake output for connection costs, because it creates bugs if a selection of `countries` is taken that does not contain links, and `remove_stubs` is set to `false`.

Also, `UA` is not part of the synthetic load data, so removing `UA` there. I noticed the `connection_costs` bug when running `UA` only which is why I'm combining those two bugfixes here.

workflow tested also with config.default (all countries) for power and all other sectors.

I'm not aware of any need of the output in other rules.

## Checklist


koen-vg:
I guess this actually fixes https://github.com/PyPSA/pypsa-eur/issues/843 too?

martacki:
> I guess this actually fixes #843 too?

awesome!

PR #1034, opened 2024-04-20
set OCGT capacities from generators to links

yerbol-akhmetov:
Good day. Here I propose to set `p_nom` and `p_nom_min` for OCGT in n.links from `p_nom` n.generators (before OCGT is removed using `remove_elec_base_techs` in `prepare_sector_networks`. 

Currently `p_nom` of n.links for OCGT is 0 after `prepare_sector_network`. I helps to set capacities back, when OCGT is moved from generators to links.

## Checklist

### DIFF
--- a/scripts/prepare_sector_network.py
+++ b/scripts/prepare_sector_network.py
@@ -3690,7 +3690,8 @@ def lossy_bidirectional_links(n, carrier, efficiencies={}):
 
 def get_capacities_from_elec(n, carrier, component):
     """
-    Gets capacities for {carrier} in n.{component} that were previously assigned in add_electricity
+    Gets capacities for {carrier} in n.{component} that were previously
+    assigned in add_electricity.
     """
     component_list = ["generators", "storage_units", "links", "stores"]
     component_dict = {name: getattr(n, name) for name in component_list}
--- a/scripts/prepare_sector_network.py
+++ b/scripts/prepare_sector_network.py
@@ -3711,20 +3711,24 @@ def lossy_bidirectional_links(n, carrier, efficiencies={}):
 
 
 def get_nominal_capacities(n, carrier, component):
-    """Gets capacities for {carrier} in n.{component} """
+    """
+    Gets capacities for {carrier} in n.{component}
+    """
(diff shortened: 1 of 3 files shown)


scripts/prepare_sector_network.py
```diff
@@ -3732,6 +3751,8 @@ def lossy_bidirectional_links(n, carrier, efficiencies={}):
     )
     pop_weighted_energy_totals.update(pop_weighted_heat_totals)
 
+    capacities_OCGT, efficiencies_OCGT = get_capacities(n, carrier="OCGT", component="generators")
```
martacki: maybe we shouldn't make it a default, but make it configurable? i.e. in config.default define a new entry in `sector`: `keep_OCGT: False` which is False by default and then something like
```suggestion
    if options["keep_OCGT"]:
        capacities_OCGT, efficiencies_OCGT = get_capacities(n, carrier="OCGT", component="generators")
    else:
        capacities_OCGT, efficiencies_OCGT = 0, None
```

(please test before comitting, I just made it up). This would ensure that the current workflow isn't changed

scripts/prepare_sector_network.py
```diff
@@ -3732,6 +3752,10 @@ def lossy_bidirectional_links(n, carrier, efficiencies={}):
     )
     pop_weighted_energy_totals.update(pop_weighted_heat_totals)
 
+    capacities_OCGT, efficiencies_OCGT = get_capacities(
+        n, carrier="OCGT", component="generators"
+    )
```
martacki, line 3757: maybe we shouldn't make it a default, but make it configurable? i.e. in `config.default` define a new entry in `sector`: `keep_OCGT: False` which is `False` by default and then something like
```suggestion
    if options.get("keep_OCGT", False):
        capacities_OCGT, efficiencies_OCGT = get_capacities(
            n, carrier="OCGT", component="generators"
        )
    else:
        capacities_OCGT, efficiencies_OCGT = 0, None
```

(please test before comitting, I just made it up). This would ensure that the current workflow isn't changed

martacki (approved):

PR #1037, opened 2024-04-24
Add coal generators

yerbol-akhmetov:
Good day.Here I propose to update the `prepare_sector_networks.py` script to facilitate addition of coal-based electricity generators attached to AC buses when coal and lignite is added to configs as follows:
image

`EU coal` is considered to supply `coal for industry` as previous. Therefore we make sure that `EU coal` is created even though `coal` carrier is already there (from eariler created coal generators). Also carrier for coal is not redefined again, because it is already there. This code works both when adding `coal` and `lignite` to config and without them being added.

## Checklist

### DIFF
--- a/scripts/prepare_sector_network.py
+++ b/scripts/prepare_sector_network.py
@@ -3689,12 +3689,12 @@ if __name__ == "__main__":
 
         snakemake = mock_snakemake(
             "prepare_sector_network",
-            configfiles="../../configs/EEE_study/config.rigid-industry_2030.yaml",
+            configfiles="test/config.overnight.yaml",
             simpl="",
             opts="",
-            clusters="48",
-            ll="v1.15",
-            sector_opts="CO2L0.45-100H-T-H-B-I",
+            clusters="37",
+            ll="v1.0",
+            sector_opts="CO2L0-24H-T-H-B-I-A-dist1",
             planning_horizons="2030",
         )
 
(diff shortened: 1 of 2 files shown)


scripts/prepare_sector_network.py
```diff
@@ -510,14 +510,15 @@ def add_carrier_buses(n, carrier, nodes=None):
...
+    # skip if carrier already exists (continue with coal because it can be already added by in generators)
+    if carrier in n.carriers.index and not carrier == "coal":
         return
 
     if not isinstance(nodes, pd.Index):
         nodes = pd.Index(nodes)
 
-    n.add("Carrier", carrier)
+    if not carrier == "coal":
+        n.add("Carrier", carrier)
```
martacki, line 521: Why is `coal` already in `carriers`, but there are no coal generators & stores? Is it because the generator and store are `coal for industry`?

FabianHofmann:
Thanks @yerbol-akhmetov for the PR. It feels that this can be further generalized. Perhaps, instead of `carrier == 'coal'` perhaps we can use `carrier in snakemake.config['pypsa-eur']['Generator']`?

yerbol-akhmetov:
> Thanks @yerbol-akhmetov for the PR. It feels that this can be further generalized. Perhaps, instead of `carrier == 'coal'` perhaps we can use `carrier in snakemake.config['pypsa-eur']['Generator']`?

Agree, @FabianHofmann, thanks for noting.

yerbol-akhmetov:
> snakemake.config['pypsa-eur']['Generator']

Only thing is `carrier in snakemake.config['pypsa_eur']['Generator']` make it create store and generator for not only `coal`, but also `lignite` if it is added to `pypsa_eur` parameter in config file. But is this desirable? It was thought of created coal and lignite generators to be attached to AC, not sure about lignite store. What do you think?

martacki:
Closed in favor of #1038

PR #1039, opened 2024-04-25
Fix double space in existing capacities

koen-vg:
A very minor fix, but even this will surely avoid some confusion in the long run.

The `name_suffix` variable already includes a space after the bus name, so before this there used to be a double space after the bus name in the added capacities.

## Checklist

### DIFF
--- a/rules/build_electricity.smk
+++ b/rules/build_electricity.smk
@@ -433,7 +433,7 @@ rule add_electricity:
             else resources("networks/base.nc")
         ),
         tech_costs=lambda w: resources(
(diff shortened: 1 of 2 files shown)


koen-vg:
Pretty sure that this pre-commit snakefmt commit is incorrect; a bit annoying cause I see snakefmt apply this change all the time. Maybe if you're quick you can still remove this from master?

koen-vg:
Ah then I learned something new today! Anyway the CI is running just fine so it's in fact not a problem as you say.

PR #1054, opened 2024-05-06
Fix hydropower and load bugs

joph:
### Bugfix for `approximate_missing_eia_stats()`

If ES or PT were not present in the list of optimized countries, the IEA hydropower approximation would fail, i.e. `approximate_missing_eia_stats()`. Therefore runs, where  the option `eia_approximate_missing` was activated would fail for weather years not part of the IEA data. The functoin would throw a key error:
`KeyError: "None of [Index(['ES', 'PT'], dtype='object', name='name')] are in the [index]"`

The reason is that `approximate_missing_eia_stats()` overwrites anomalous years for 'ES' and 'PT'. By checking if the countries are present before overwriting the data, the bug can be fixed:
`if ("ES") in runoff:`
        `    runoff.loc[1978, ["ES"]] = runoff.loc[1979, ["ES"]]`
`if ("PT") in runoff:`
        `    runoff.loc[1978, ["PT"]] = runoff.loc[1979, ["PT"]]`

### Bugfix for `manual_adjustment()`

The `copy_timeslice()` procedure will add countries to the load data which are not present in the countries list, if a `fn_load` object is provided. If `manual_adjustment()` is called, which uses `copy_timeslice()`, this will partly add data for some countries such as GB. Later on, this causes an `AssertionError: Load data contains nans.`

## Checklist

- [X] I tested my contribution locally and it seems to work fine.

### DIFF
--- a/scripts/build_hydro_profile.py
+++ b/scripts/build_hydro_profile.py
@@ -140,9 +140,9 @@ def approximate_missing_eia_stats(eia_stats, runoff_fn, countries):
 
     # fix outliers; exceptional floods in 1977-1979 in ES & PT
     if "ES" in runoff:
-        runoff.loc[1978, ["ES"]] = runoff.loc[1979, ["ES"]]
+        runoff.loc[1978, "ES"] = runoff.loc[1979, "ES"]
     if "PT" in runoff:
-        runoff.loc[1978, ["PT"]] = runoff.loc[1979, ["PT"]]
+        runoff.loc[1978, "PT"] = runoff.loc[1979, "PT"]
 
     runoff_eia = runoff.loc[eia_stats.index]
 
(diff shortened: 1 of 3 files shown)


scripts/build_hydro_profile.py
```diff
@@ -139,7 +139,10 @@ def approximate_missing_eia_stats(eia_stats, runoff_fn, countries):
     runoff.index = runoff.index.astype(int)
 
     # fix outliers; exceptional floods in 1977-1979 in ES & PT
-    runoff.loc[1978, ["ES", "PT"]] = runoff.loc[1979, ["ES", "PT"]]
+    if "ES" in runoff:
+        runoff.loc[1978, ["ES"]] = runoff.loc[1979, ["ES"]]
+    if "PT" in runoff:
+        runoff.loc[1978, ["PT"]] = runoff.loc[1979, ["PT"]]
```
fneum, line 145:
```suggestion
    if "ES" in runoff:
        runoff.loc[1978, "ES"] = runoff.loc[1979, "ES"]
    if "PT" in runoff:
        runoff.loc[1978, "PT"] = runoff.loc[1979, "PT"]
```

fneum (approved):
Thanks @joph! Looks good with a minor style adjustment.

PR #1065, opened 2024-05-15
Split out time aggregation to its own rule

koen-vg:
As noted, build year aggregation as in https://github.com/PyPSA/pypsa-eur/pull/1056 doesn't work as implemented when combined with time series segmentation because the segmentation could (and often is) different for different planning horizons.

The present PR splits time aggregation off into its own rule which is run only once for all planning horizons. This rule is what constructs the new subset of snapshots and their weights; the aggregation of time-dependent data is performed in `prepare_sector_network.py`

I thought this isn't a terrible thing even independently of build year aggregation; there might be other times when you want to directly compare the same network at different planning horizons snapshot-to-snapshot. I also imagine it might be a little easier to implement new time aggregation features as this step is now a little more isolated and easier to comprehend.

Do let me know if this seems like a reasonable idea or not, or if anyone could imagine a different but better way of achieving the same objective.

## Checklist


fneum (approved):
Excellent! This is a very good idea!

PR #1067, opened 2024-05-17
fix_DC_cluster_issue

### LINKED ISSUE #996: Clustering problem seemingly caused by addition of new TYNDP projects
## Checklist

- [x] I am using the current [`master`](https://github.com/PyPSA/pypsa-eur/tree/master) branch or the latest [release](https://github.com/PyPSA/pypsa-eur/releases). Please indicate.

## Describe the Bug

Commit c5026aaf9b24a4ec17049e03cf2f6953b7059f02 seems to have caused a strange kind of regression. It causes an additional sub_network with a single node to be added to Germany, but one without any load, meaning that that node isn't clustered correctly (since no nodes are allocated to it in the clustering).

The problem goes away after removing the following line from `data/links_tynpd.csv`:
````
Germany-UK Hybrid Interconnector,Fetteresso (UK),Emden (DE),800,,2000,under consideration,,https://eepublicdownloads.blob.core.windows.net/public-cdn-container/tyndp-documents/TYNDP2024/240220_TYNDP2024_project_portfolio.xlsx,-2.383,56.991,7.207,53.376
````

The node in question has a `generator` column with the value `onwind`, so I guess that the mechanism of the problem is that some onshore wind farm happens to be very close to one of the above HVDC terminals, and gets assigned to the corresponding bus. But the carrier is `DC` for this node, and so with this carrier and a generator attached, it doesn't get simplified/merged into the German AC grid.

The above is kind of conjecture; I didn't really have time to dig deep into the details. I'm not super familiar with the building and simplification of the base network, so I'm not 100% where to fix the problem either.

For reference, the problem only actually surfaced a while later in the workflow for the sector-coupled network when clustering the gas network.

p-glaum:
Closes #996 

This PR fixes the bug described in #996. It guarantees that new links defined in `links_tyndp.csv` are only connected to AC buses. Before, this caused an error because we obtained a "meshed" DC grid in the model, which was not cleaned in the simplify_network script and led to wrong clustering.

## Checklist

### DIFF
--- a/scripts/base_network.py
+++ b/scripts/base_network.py
@@ -272,14 +272,15 @@ def _add_links_from_tyndp(buses, links, links_tyndp, europe_shape):
         if links_tyndp.empty:
             return buses, links
 
-    tree = spatial.KDTree(buses[["x", "y"]])
+    tree_buses = buses.query("carrier=='AC'")
+    tree = spatial.KDTree(tree_buses[["x", "y"]])
     _, ind0 = tree.query(links_tyndp[["x1", "y1"]])
-    ind0_b = ind0 < len(buses)
-    links_tyndp.loc[ind0_b, "bus0"] = buses.index[ind0[ind0_b]]
+    ind0_b = ind0 < len(tree_buses)
+    links_tyndp.loc[ind0_b, "bus0"] = tree_buses.index[ind0[ind0_b]]
 
     _, ind1 = tree.query(links_tyndp[["x2", "y2"]])
-    ind1_b = ind1 < len(buses)
-    links_tyndp.loc[ind1_b, "bus1"] = buses.index[ind1[ind1_b]]
+    ind1_b = ind1 < len(tree_buses)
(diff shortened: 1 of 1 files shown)


tgi-climact:
Thanks for the PR @p-glaum . When I looked up into the issue, I saw that filtering only AC buses resulted in different coordinates for the tyndp links. Have you checked this ? Is this a problem ? Thanks

p-glaum:
In my opinion this should not make a difference, because so far as I know, in links_tyndp the geo coordinates are just taken roughly from Open Street Maps. Additionally, the exact location of the TYNDP links isn't clear anyway. We want to switch to OSM for the base network anyway, and then we will also have some new functionalities which can better handle those cases from #996. For example, if there is a new DC link for an offshore node which currently does not exist in the base network, `base_network.py` will add a new bus at the given location.

Have you found any big differences? Because usually, we aggregate the network anyway and do not solve it in high resolution.

tgi-climact:
No big differences founded. I was just sharing my observations and related questions. Glad to read that it should not be a problem. Then the fix is quite simple.

PR #1068, opened 2024-05-17
Define methanol energy demand for industry

tgi-climact:

The industrial energy demand of methanol was methane and electricity. I suggest to use the low carbon route (already used for shipping methanol) defined by [DECHEMA](https://dechema.de/dechema_media/Downloads/Positionspapiere/Technology_study_Low_carbon_energy_and_feedstock_for_the_European_chemical_industry.pdf). This route uses hydrogenation of $CO_2$ to produce methanol, which requires hydrogen, carbon dioxide and electricity.

## Checklist


fneum:
Neat suggestion and nicely prepared! With intermediate industry sector ratios, there is now an exogenous transition from methane-based methanol production to the low-carbon route. An extension of this would be to directly add links for the methane route to the model and endogenize this.

tgi-climact:
@fneum Very little detail here. I realise it would be better to have `#467b8c` (or something else) as the colour for industry methanol (https://github.com/PyPSA/pypsa-eur/blob/master/config/config.default.yaml#L1156) rather than the same colour as shipping (`#468c8b`). Do you want a PR for this ?

PR #1091, opened 2024-06-06
Fix grouping logic again

lindnemi:
#967 takes the point of view, that the capacities added by the model in 2025 correspond to the time interval [2025,2030]. The logic of the grouping years for existing capacities was different and solar-2020 would correspond to capacities installed in [2015,2020].  #967 aligned the way the grouping years work, s.t. solar-2015 would correspond to the interval [2015,2020].

However, as pointed out in https://github.com/PyPSA/pypsa-ariadne/issues/101 the interpretation that solar-2025 corresponds to [2025,2030] might be flawed. In pypsa-ariadne we impose capacities and co2 limits corresponding to actual laws. These limits apply to the year 2025,  thus any capacities extensions of the solar-2025 generator required to achieve these limit have to be built before 2025. Hence the more correct point of view is that the capacities expanded by the model in 2025 correspond to the time interval [2020,2025]. This is how the grouping years of existing capacities worked before #967, and thus this PR mostly reverts #967.

Open issues:

- [x] grouping years for heating may not extend 2020. Otherwise infeasibilities occur. I have no idea why, but i would prefer if this would be handled in the code and not in the config

- [x] The renewable capacity added in the base year is neglected because of the land use constraint in solve_network. it would be desirable to have separate Generators for extendable and existing renewable capacities.

### DIFF
--- a/scripts/add_existing_baseyear.py
+++ b/scripts/add_existing_baseyear.py
@@ -224,30 +224,21 @@ def add_power_capacities_installed_before_baseyear(n, grouping_years, costs, bas
     phased_out = df_agg[df_agg["DateOut"] < baseyear].index
     df_agg.drop(phased_out, inplace=True)
 
-    older_assets = (df_agg.DateIn < min(grouping_years)).sum()
-    if older_assets:
+
+    newer_assets = (df_agg.DateIn > max(grouping_years)).sum()
+    if newer_assets:
         logger.warning(
-            f"There are {older_assets} assets with build year "
-            f"before first power grouping year {min(grouping_years)}. "
+            f"There are {newer_assets} assets with build year "
+            f"after last power grouping year {max(grouping_years)}. "
             "These assets are dropped and not considered."
             "Consider to redefine the grouping years to keep them."
         )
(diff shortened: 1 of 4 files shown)


scripts/add_existing_baseyear.py
```diff
@@ -224,30 +224,19 @@ def add_power_capacities_installed_before_baseyear(n, grouping_years, costs, bas
     phased_out = df_agg[df_agg["DateOut"] < baseyear].index
     df_agg.drop(phased_out, inplace=True)
 
-    older_assets = (df_agg.DateIn < min(grouping_years)).sum()
-    if older_assets:
+    newer_assets = (df_agg.DateIn > max(grouping_years)).sum()
```
lisazeyen, line 227: shouldn't it be `df_agg.DateIn >= max(grouping_years)`?

lindnemi: Good question... at the moment the 2020 generator is for 2016-2020. So 2020 will be included and everything >2020 will be excluded.

The opposite case of excluding 2020 might be reasonable as well, though...

PR #1099, opened 2024-06-12
Fix and simplify solar potential constraint

koen-vg:
The right hand side of equation was incorrect; the `solar-hsat` capacities shouldn't be subtracted! The assumption is that the rhs of the equation is the _total_ solar potential, and that total potential is stored in `p_nom_max` of the "regular" solar generators.

I have indeed run into infeasibility caused by this mistake.

I also took the opportunity to clean up some of the surrounding code.

Finally, I'll note that the tests are pretty much guaranteed to succeed whether or not this fix is correct (you'll have to trust my personal judgement), because there are no extendable renewables in the test configs! I'm sure this is not actually intentional; maybe from back when renewables were automatically extendable? Anyway I would suggest an update of the test configs :)

## Checklist

### DIFF
--- a/doc/release_notes.rst
+++ b/doc/release_notes.rst
@@ -10,6 +10,8 @@ Release Notes
 Upcoming Release
 ================
 
+* Bugfix for previously incorrect total solar potential constraint (only relevant when ``solar-hsat`` was enabled).
+
 * Set p_nom = p_nom_min for generators with baseyear == grouping_year in add_existing_baseyear. This has no effect on the optimization but helps n.statistics to correctly report already installed capacities.
 
 * Reverted outdated hotfix for doubled renewable capacity in myopic optimization.
--- a/scripts/solve_network.py
+++ b/scripts/solve_network.py
@@ -224,7 +224,6 @@ def add_solar_potential_constraints(n, config):
         "solar-hsat": config["renewable"]["solar"]["capacity_per_sqkm"]
         / config["renewable"]["solar-hsat"]["capacity_per_sqkm"],
     }
-    rename = {"Generator-ext": "Generator"}
 
     solar_carriers = ["solar", "solar-hsat"]
     solar = n.generators[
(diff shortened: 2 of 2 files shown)


lisazeyen:
FYI @Parisra

lisazeyen:
@koen-vg I fully agree with adjusting the test configs.

I have one question regarding your PR, if you run the code myopic, shouldn't you subtract from your `p_nom_max` of your "regular" solar generators the already build capacities of the `hsat` generators?

koen-vg:
Ah, you know what, I think you might actually be right and I just got confused.

I think I might actually just have ran into an issue with this constraint in another branch (the build year aggregation) that I guess was actually messing with this, and got me confused about the logic.

Just to test that I understand this correctly, `p_nom_max` of the solar generators is adjusted downwards by previously installed capacities in myopic optimisations, but is _not_ adjusted down by the previously installed solar-hsat capacities, right? It would be equivalent to subtract the previously installed solar-hsat capacities here, or to subtract them from `p_nom_max` of the solar generators in `_add_land_use_constraint`.

If I now understand this correctly, then this PR should just be ignored and the most productive thing that came out of it was a realisation that the tests are probably not quite working as we want :)

lisazeyen:
Yes, exactly. Maybe it would be cleaner to subtract the already build `solar-hst` in `_add_land_use_constraint` but then the constraint should currently work as intended and I close this PR.

koen-vg:
Thanks for your time, sorry for the mistaken PR!

Parisra:
> @koen-vg I fully agree with adjusting the test configs.
> 
> I have one question regarding your PR, if you run the code myopic, shouldn't you subtract from your `p_nom_max` of your "regular" solar generators the already build capacities of the `hsat` generators?

Thanks @lisazeyen, this is exactly the purpose of the constraint (hence the use of 'p_nom_opt' for solar-hsat). To clarify a little here: 
- No capacity would be subtracted for an overnight optimisation. 
- For a myopic optimisation, two things will happen 
           - as @koen-vg mentioned,  p_nom_max of the solar generators is adjusted downwards by previously installed capacities in _add_land_use_constraint. This also happens separately for solar-hsat generators.
           - The constraint mandates that the total installed capacity of solar and solar-hsat generators during each investment period is less than the remaining potential. The remaining potential is calculated by taking the solar potential (already adjusted to current horizon) and subtracting the previously installed solar-hsat. 

I know this implementation is a little complicated, but the issue here is we have two potentials (one for solar and one for solar-hsat), but both represent the solar potential of the region. Therefore, we have to do a calculation each time to find how much potential we actually have left.

PR #1103, opened 2024-06-14
non-zero capital_cost for methanol stores

lindnemi:
For pypsa-ariadne @toniseibold added a DE specific methanol store with non-zero capital cost. For pypsa-eur the capital cost is zero, which leads to unrealistically large e_nom_opt values. 

I directly added capital_cost in `madd`. For the other stores the capital cost is added in `add_carrier_buses`. Maybe methanol should be handled via this function as well? @fneum

### DIFF
--- a/doc/release_notes.rst
+++ b/doc/release_notes.rst
@@ -10,6 +10,8 @@ Release Notes
 Upcoming Release
 ================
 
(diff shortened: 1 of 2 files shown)


lindnemi:
https://github.com/PyPSA/pypsa-ariadne/pull/108 related

fneum (approved):

PR #1104, opened 2024-06-14
dirtiest hottest fix for oil demand in transport sector

toniseibold:
Closing https://github.com/PyPSA/pypsa-ariadne/issues/102 for now.

Problem is, that in `build_energy_totals.py` the efficiencies of cars are taken from JRC IDEES.
Also the `config["sector"]["transport_ice_efficiency"] is only valid for cars.
So far, from the energy demand a transport demand in kilometers is calculated neglecting the demand in road freight transport.
This leads to a significant difference in oil demand between PyPSA-Eur and reality.
Until this is fixed, this very hot and dirty fix ensures that the demand in the transport sector is correct.
It takes an average fuel efficiency across all road vehicles.

Additionally in the config is changed to the following values:
```
  transport_ice_efficiency: 10.637 # 1 MWh_oil = 16.0712 * 100 km
```

This leads to the following demand in transport sector:
final_energy_detailed
primary_energy

### DIFF
--- a/scripts/build_energy_totals.py
+++ b/scripts/build_energy_totals.py
@@ -305,8 +305,8 @@ def idees_per_country(ct, base_dir):
     row = "Heavy duty vehicles (Diesel oil incl. biofuels)"
     ct_totals["total heavy duty road freight"] = df.loc[row]
 
-    assert df.index[61] == "Passenger cars"
-    ct_totals["passenger car efficiency"] = df.iloc[61]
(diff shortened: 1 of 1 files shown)


toniseibold:
@lindnemi: note that we are underestimating the electricity demand for transport in 2045 since the `config[sector][transport_electric_efficiency]` is also just for private vehicles. However, I don't have robust data for a mixed pool of vehicles from JRC IDEES yet.

PR #1110, opened 2024-06-20
fix sector coupeled model for isolated non EU28 countries, e.g. UA

martacki:
contents:
Just making the sec model run by filling missing sectoral data with defaults, average EU values or zeros, if nothing else is available. Thus, it currently has no real sectoral data, but allows the execution of the power model only in the sector coupeled style (i.e. adding distribution grids, solar thermal, etc.)

other comments:
experimenting to use git cherry-pick from OET/pypsa-eur fork to align our master branch with upstream ... (was this PR in OET fork https://github.com/open-energy-transition/pypsa-eur/pull/1)... unfortunately used the wrong git commands, so that the history was lost

## Checklist

### DIFF
--- a/scripts/build_energy_totals.py
+++ b/scripts/build_energy_totals.py
@@ -1401,7 +1401,40 @@ if __name__ == "__main__":
         disable_progressbar=snakemake.config["run"].get("disable_progressbar", False),
     )
     swiss = build_swiss()
-    idees = build_idees(idees_countries)
+    if len(idees_countries) > 0:
+        idees = build_idees(idees_countries)
+    else:
+        # e.g. UA and MD
+        logger.info(f"No IDEES data available for {countries} and years 2000-2015. Filling with zeros.")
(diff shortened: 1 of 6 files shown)


scripts/build_energy_totals.py
```diff
@@ -1401,7 +1401,40 @@ def update_residential_from_eurostat(energy: pd.DataFrame) -> pd.DataFrame:
...
+                "electricity rail freight",
+                "total domestic aviation passenger",
+                "total domestic aviation freight",
+                "total international aviation passenger",
+                "total international aviation freight",
+                "derived heat residential",
+                "derived heat services",
+                "thermal uses residential",
+                "thermal uses services",
+            ],
```
martacki, line 1439: didn't find a better way of doing this, suggestions appreciated

(maybe sth like 
idees = build_idees(['DE'])
and then replacing all values by 0 or so could be better?)

PR #1114, opened 2024-06-24
Quick fix of missing country but in add_existing_baseyear.py

p-glaum:
I guess it either has to do with Pandas version or powerplantmatching.
Could not reproduce the bug since currently I do not have it but maybe check whether missing countries are in
```
import powerplantmatching as pm
irena=pm.data.IRENASTAT().powerplant.convert_country_to_alpha2()
irena.Country.unique()
```
@toniseibold checked and the coutnries were inside `irena.Country.unique()` in her environment which caused the error

Closes # (if applicable).

## Checklist

### DIFF
--- a/scripts/add_existing_baseyear.py
+++ b/scripts/add_existing_baseyear.py
@@ -104,7 +104,7 @@ def add_existing_renewables(df_agg, costs):
                 n.generators.loc[gens, "bus"]
             ).sum()
 
-        nodal_df = df.reindex(n.buses.loc[elec_buses, "country"])
+        nodal_df = df.reindex(n.buses.loc[elec_buses, "country"], fill_value=0)
         nodal_df.index = elec_buses
         nodal_df = nodal_df.multiply(nodal_fraction, axis=0)
 
--- a/scripts/add_existing_baseyear.py
+++ b/scripts/add_existing_baseyear.py
@@ -104,7 +104,7 @@ def add_existing_renewables(df_agg, costs):
                 n.generators.loc[gens, "bus"]
             ).sum()
 
-        nodal_df = df.loc[n.buses.loc[elec_buses, "country"]]
+        nodal_df = df.reindex(n.buses.loc[elec_buses, "country"])
         nodal_df.index = elec_buses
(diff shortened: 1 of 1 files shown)


toniseibold (approved):

lindnemi:
I am wondering why the countries are missing in the first place. Any ideas? @p-glaum 

The way the code is now we at least get a warning that they are missing...

p-glaum:
Toni was checking yesterday to see if the missing countries are in the database from powerplantmatching and there they were existing. So, I guess it is somehow connected with the panda's functions being used, which leads to a dataframe without the countries in the index.

scripts/add_existing_baseyear.py
```diff
@@ -104,7 +104,7 @@ def add_existing_renewables(df_agg, costs):
                 n.generators.loc[gens, "bus"]
             ).sum()
 
-        nodal_df = df.loc[n.buses.loc[elec_buses, "country"]]
```
fneum, line 107: How are the NaN values handled here? Shouldn't they be set to zero?

fneum (commented):
Would be good to understand the problem rather than taking measures against the symptoms.

scripts/add_existing_baseyear.py
p-glaum: I agree that this is only a quick fix. So this is not meant to pull into the master. I do not get the error anymore, so I cannot investigate the error.

p-glaum: > How are the NaN values handled here? Shouldn't they be set to zero?

True. I added fill_value=0 to reindex()

toniseibold:
@p-glaum 
```
irena[irena.Technology.str.contains("Offshore")].Country.unique()
['AT' 'BE' 'CH' 'CZ' 'DE' 'DK' 'FR' 'GB' 'LU' 'NL' 'NO' 'PL' 'SE']
```

fneum:
> So this is not meant to pull into the master.

So no pull request?

toniseibold:
@lindnemi is opposed to merging something without knowing what exactly is going on. So no pull request.

PR #1116, opened 2024-06-25
Rename ev battery master

lindnemi:
The old name is confusing. Only caveat: With this PR the Store has the exact same name as the bus. I don't think that's a problem but maybe there is a convention not to do that?

## Checklist


lindnemi:
Related https://github.com/PyPSA/pypsa-ariadne/pull/127

fneum (approved):
Breaking change.

I think there are several instances where store has same carrier as bus.

lindnemi:
> I think there are several instances where store has same carrier as bus.

yes, i saw that, but i was not worried about the carriers but the names (since i changed the suffixes to be the same)

PR #1141, opened 2024-07-09
FT and electrolysis waste heat for DH as float

cpschau:
The waste heat of Fischer-Tropsch synthesis and electrolyzers is dominating the urban central heat mix for planning horizons > 2030 (https://github.com/PyPSA/pypsa-ariadne/issues/114). This PR allows for the multiplication of the link efficiency by a float in order to make more conservative assumptions regarding the usage of this waste heat.

### DIFF
--- a/config/config.default.yaml
+++ b/config/config.default.yaml
@@ -571,12 +571,12 @@ sector:
   min_part_load_fischer_tropsch: 0.5
   min_part_load_methanolisation: 0.3
   min_part_load_methanation: 0.3
-  use_fischer_tropsch_waste_heat: true
+  use_fischer_tropsch_waste_heat: 0.2
   use_haber_bosch_waste_heat: true
   use_methanolisation_waste_heat: true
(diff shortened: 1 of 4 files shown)


lindnemi:
Thanks for getting started with this so quickly. A few minor points

- Why did you settle for 0.2? Tom suggested 0.25-0.5 as far as i remember.
- A release_note might be helpful
- Methanolisation, Sabatier and H2 Fuel Cells also connect to central heat and should probably get their own multipliers

lindnemi (approved):
image

I already merged it into ariadne2 and it seems to do it's job. :clap:

fneum:
Great! Just leaving a note on backwards compatibility for next time: The change in the config file was purely stylistic but leads to old configs not being compatible with the new version.

lindnemi:
ups, i didn't notice that in the review :grimacing: should we revert to the old config style before the next release?

fneum:
Yes, let's do it!

PR #1143, opened 2024-07-10
Config for Industry Steam Production

toniseibold:
Lays the foundation to close https://github.com/PyPSA/pypsa-ariadne/issues/121

So far all industry steam is assumed to be produced from biomass which leads to high biomass and biomass CC volumes in that sector. Here, the industry config `steam_biomass_fraction`, `steam_hydrogen_fraction` and `steam_electricity_fraction` is added. 
Those configs are taking effect in `build_industry_sector_ratios.py`. Except for chemical steam processing which is assumed to be at > 1000°C the steam is provided by the fractions given in the config.
Following the current approach, the efficiency of hydrogen steam provision is expected to be as high as the efficiency from natural gas steam provision. The provision of steam by electricity is expected to have an efficiency of 100 %.

## Checklist


scripts/build_industry_sector_ratios.py
```diff
@@ -1471,6 +1607,23 @@ def other_industrial_sectors():
 
     params = snakemake.params.industry
 
+    if (
```
lindnemi, line 1610: This should probably throw an error, i prefer if the fractions can not directly be used to fiddle with industry demand

lindnemi (approved):

lindnemi:
@fneum might make sense to integrate this into master and also to change the default assumption that all steam is provided by biomass

PR #1151, opened 2024-07-12
chore: ignore all `pre-commit` commits in blame

lkstrp:
Add all auto-fixing `pre-commit` commits to git blame ignore.

To activate locally run:

```
git config blame.ignoreRevsFile .git-blame-ignore-revs
```

GitHub should show those changes immediatly.

Note: 
`pre-commit` will still show up as the author of any lines which have been created by it, since there is no one else to blame.

@Irieo Let me know if this solves your issue, then this could be updated automatically via a workflow.


Irieo:
yes, so convenient now! @lkstrp

PR #1163, opened 2024-07-18
enable EQ constraint on locations

martacki:
Closes # (if applicable).

enable the EQ constraint to operate on locations, particularly useful for sector coupled models

## Checklist


tgi-climact:
Hi @martacki , I am also interested by this feature. If you intend to make it work for sector-coupled models, it seems to me that we should also add links related to power supply into the constraint. Am I missing something ?

martacki:
Hi @tgi-climact thanks for your interest in the feature and remark! You're absolutely right. Needs some filtering though, I think, so that we always take the right pX stream (i.e. p1, p2, ...), if applicable. Need to have a close look into the generators, too.

I didn't have any generating units in my network that were attached by links, so I'll need to find a new test-case for this

koen-vg:
Related to https://github.com/PyPSA/pypsa-eur/pull/659? Unfortunately I believe implementing the EQ constraint for sector-coupled networks can be a bit complicated (see the docstring of the rewritten EQ constraint function in the aforementioned PR).

(P.S. I do hope to find the time to revisit https://github.com/PyPSA/pypsa-eur/pull/659 and see if I can finally get it merged. Looking back at it now it does seem like maybe it's more complicated than it needs to be so I'll be on the look-out for simplifications.)

PR #1172, opened 2024-07-21
base_network: use GeoSeries.voronoi_polygons instead of custom solution

fneum:
- [x] could perhaps be put into function
- [x] before and after shapes compared and identical


FabianHofmann:
nice one :)

PR #1175, opened 2024-07-22
Approximate district heating COPs via Jensen et al. 2018

### LINKED ISSUE #857: Check COP calculation of heat pumps
## Compare current COP calculation with Lorenz COP calculation of DEA

- DEA COP calculations p.285 [here](https://ens.dk/sites/ens.dk/files/Analyser/technology_data_catalogue_for_el_and_dh.pdf)

amosschle:
Closes #857.

with respect to central heating.

- Introduce separate forward and return temperatures for district (central) heating.
- The rule `build_cop_profiles` builds separate heat pump COPs for decentral and central heating.
- The COP approximation for central heating is based on [Jensen et al. 2018](https://backend.orbit.dtu.dk/ws/portalfiles/portal/151965635/MAIN_Final.pdf).
- Additional config settings for the approximation are added, with default values partly based on [Pieper et al. 2019](https://www.sciencedirect.com/science/article/pii/S0360544220311014)

## Checklist

Jupyter notebook and config files are attached below (`.txt` must be changed accordingly).

## Impact on results
A quick test run (DE only, 2 nodes, 2050) suggests the resulting COPs after changing the approximation method **and** forward/return temperatures in district heating to be somewhat similar as before.

Updating only the approximation method but setting an (unrealistic) forward temperature of 55C (the current default) results in significantly higher central heat pump COPs and higher heat pump investments.

image
image
image

[jupyter_notebook.txt](https://github.com/user-attachments/files/16378347/jupyter_notebook.txt)

[config.txt](https://github.com/user-attachments/files/16378356/config.txt)

[scenarios.txt](https://github.com/user-attachments/files/16378358/scenarios.txt)


doc/configtables/sector.csv
```diff
@@ -5,9 +5,17 @@ biomass,--,"{true, false}",Flag to include biomass sector.
 industry,--,"{true, false}",Flag to include industry sector.
 agriculture,--,"{true, false}",Flag to include agriculture sector.
 district_heating,--,,`prepare_sector_network.py <https://github.com/PyPSA/pypsa-eur-sec/blob/master/scripts/prepare_sector_network.py>`_
--- potential,--,float,maximum fraction of urban demand which can be supplied by district heating. Ignored where below current fraction.
--- progress,--,Dictionary with planning horizons as keys., Increase of today's district heating demand to potential maximum district heating share. Progress = 0 means today's district heating share. Progress = 1 means maximum fraction of urban demand is supplied by district heating
--- district_heating_loss,--,float,Share increase in district heat demand in urban central due to heat losses
+#NAME?,--,float,maximum fraction of urban demand which can be supplied by district heating. Ignored where below current fraction.
```
fneum, line 8: @amosschle reviewing now, something went wrong with the configtables (`#Name?`). Could you revert this?

amosschle: My bad, thanks for spotting thiis. Should be fixed now.

config/config.default.yaml
```diff
@@ -408,6 +408,15 @@ sector:
       2045: 0.8
       2050: 1.0
     district_heating_loss: 0.15
+    # check these numbers!
```
fneum, line 411:
```suggestion
```
to be removed for final version

doc/release_notes.rst
```diff
@@ -10,6 +10,8 @@ Release Notes
 Upcoming Release
 ================
 
+* Changed heat pump COP approximation for central heating to be based on Jensen et al. 2018 and a default forward temperature of 90C. This is more realistic for district heating than the previously used approximation method.
```
fneum, line 13: perhaps add link to paper.

doc/configtables/sector.csv
```diff
@@ -6,6 +6,14 @@ industry,--,"{true, false}",Flag to include industry sector.
...
 district_heating,--,,`prepare_sector_network.py <https://github.com/PyPSA/pypsa-eur-sec/blob/master/scripts/prepare_sector_network.py>`_
 -- potential,--,float,maximum fraction of urban demand which can be supplied by district heating. Ignored where below current fraction.
+-- forward_temperature,°C,float,Forward temperature in district heating
+-- return_temperature,°C,float,Return temperature in district heating. Must be lower than forward temperature
+-- heat_source_cooling,K,float,Cooling of heat source for heat pumps
+-- heat_pump_cop_approximation,,,
+-- refrigerant,--,"{ammonia, isobutane}",Heat pump refrigerant assumed for COP approximation
+-- heat_exchanger_pinch_point_temperature_difference,K,float,Heat pump pinch point temperature difference in heat exchangers assumed for approximation.
+-- isentropic_compressor_efficiency,--,float,Isentropic efficiency of heat pump compressor assumed for approximation. Must be between 0 and 1.
+-- heat_loss,--,float,Heat pump heat loss assumed for approximation. Must be between 0 and 1.
```
fneum, line 16:
```suggestion
-- -- refrigerant,--,"{ammonia, isobutane}",Heat pump refrigerant assumed for COP approximation
-- -- heat_exchanger_pinch_point_temperature_difference,K,float,Heat pump pinch point temperature difference in heat exchangers assumed for approximation.
-- -- isentropic_compressor_efficiency,--,float,Isentropic efficiency of heat pump compressor assumed for approximation. Must be between 0 and 1.
-- -- heat_loss,--,float,Heat pump heat loss assumed for approximation. Must be between 0 and 1.
```
to mark subgroup of config entries

scripts/prepare_sector_network.py
```diff
@@ -1826,10 +1826,18 @@ def add_heat(n, costs):
...
+        )
+        .to_pandas()
+        .reindex(index=n.snapshots),
+        "air central": xr.open_dataarray(snakemake.input.cop_air_central_heating)
+        .to_pandas()
+        .reindex(index=n.snapshots),
+        "ground central": xr.open_dataarray(snakemake.input.cop_soil_central_heating)
         .to_pandas()
         .reindex(index=n.snapshots),
     }
```
fneum, line 1843: just a thought, but maybe one can build a dictionary with filenames first and then do:

```py
cop = {k: xr.open_dataarray(fn).to_pandas().reindex(index=n.snapshots) for k, vn in cop.items()}
```

which would reduce code length

scripts/add_existing_baseyear.py
```diff
@@ -639,29 +638,43 @@ def add_heating_capacities_installed_before_baseyear(
...
+                    snakemake.input.cop_air_central_heating
+                )
+                .to_pandas()
+                .reindex(index=n.snapshots),
+                "ground central": xr.open_dataarray(
+                    snakemake.input.cop_soil_central_heating
+                )
+                .to_pandas()
+                .reindex(index=n.snapshots),
+            },
```
fneum, line 667: see other similar comment to make this more compact through list comprehension

scripts/add_existing_baseyear.py
```diff
@@ -639,29 +638,43 @@ def add_heating_capacities_installed_before_baseyear(
...
         add_heating_capacities_installed_before_baseyear(
-            n,
-            baseyear,
-            grouping_years_heat,
-            ashp_cop,
-            gshp_cop,
-            time_dep_hp_cop,
-            costs,
-            default_lifetime,
+            n=n,
```
fneum, line 643: style suggestion: to make code less nested, perhaps created variables `cop` and `existing_heating` before function call and just pass object?

scripts/build_cop_profiles/BaseCopApproximator.py
```diff
@@ -0,0 +1,82 @@
...
+        Approximate heat pump coefficient of performance (COP).
+
+        Returns:
+        -------
+        Union[xr.DataArray, np.array]
+            The calculated COP values.
+        """
+        pass
+
+    def celsius_to_kelvin(
```
fneum, line 48: Is it necessary to duplicate the functions, once with `@staticmethod` decorator?

fneum (changes requested):
Looks mostly good to me. I like the structure!

One thing I am wondering about is whether this new COP calculation method should rather be integrated into `atlite` (i.e. here: https://github.com/PyPSA/atlite/blob/16c02fd4aa1816f73d96ed5ae52b1be69f671b18/atlite/convert.py#L259-L323). 

Bit unfortunate that I missed to tell you about this feature of `atlite` earlier. I can understand if you don't want to reimplement this now, but it would have some advantages, mainly that it would be available outside PyPSA-Eur.

scripts/build_cop_profiles/BaseCopApproximator.py
amosschle: It very much isn't.

amosschle:
> Looks mostly good to me. I like the structure, and it's very well documented!
> 
> One thing I am wondering about is whether this new COP calculation method should rather be integrated into `atlite` (i.e. here: https://github.com/PyPSA/atlite/blob/16c02fd4aa1816f73d96ed5ae52b1be69f671b18/atlite/convert.py#L259-L323).
> 
> Bit unfortunate that I missed to tell you about this feature of `atlite` earlier. I can understand if you don't want to reimplement this now, but it would have some advantages, mainly that it would be available outside PyPSA-Eur.

I agree. I'd suggest to merge it to `PyPSA-Eur` and move it to `atlite` later. Happy to open an issue for that.

fneum:
> I agree. I'd suggest to merge it to PyPSA-Eur and move it to atlite later. Happy to open an issue for that.

Yes, let's do that. Let me know when it's ready to merge. Read the Docs builds I will fix separately.

scripts/add_existing_baseyear.py
amosschle: Hm, can do but that would create unnecessary memory load, wouldn't it?

scripts/prepare_sector_network.py
amosschle: I agree. We could replace `soil` with `ground` everywhere (or vice versa), define something like `heat_pump_sources = ["air", "ground"]`, `heat_system_type=["central", "decentral"]`, maybe in `config.default`, and use dict comprehensions over `heat_pump_sources` & `heat_system_type` wherever possible. What do you think?

scripts/add_existing_baseyear.py
fneum: Also fine to leave as is, but I don't think it would create a memory overhead since no copy is made.

scripts/prepare_sector_network.py
fneum: Yes! I would vote for keeping `ground` and replacing `soil` since it changes less in the config files users are exposed to.

scripts/add_existing_baseyear.py
amosschle: This should be fixed now with the new structure using only one file for COPs

amosschle:
@fneum Thanks for the review! This is good to go from my side.

amosschle:
Ready for re-review. 
- I've made some structural changes that beautify the handling of different kinds of heat systems, system types etc. as enums and moved that to a `definitions` module, as suggested by @lkstrp. Hope that's fine.
- I'm getting slightly different results compared to a few days ago but I'm seeing that there were some changes in `master`, which I've merged into the feature that might be the reason (e.g. the clustering changed and COPs as a consequence)
- I've also adjusted the scripts for perfect foresight and myopic mode.
- Docs are not building but the error message is a little cryptic. Maybe @fneum, you can help?

lisazeyen:
> Ready for re-review.
> 
> * I've made some structural changes that beautify the handling of different kinds of heat systems, system types etc. as enums and moved that to a `definitions` module, as suggested by @lkstrp. Hope that's fine.
> * I'm getting slightly different results compared to a few days ago but I'm seeing that there were some changes in `master`, which I've merged into the feature that might be the reason (e.g. the clustering changed and COPs as a consequence)
> * I've also adjusted the scripts for perfect foresight and myopic mode.
> * Docs are not building but the error message is a little cryptic. Maybe @fneum, you can help?

- the Docs were not building through a changed syntax for the intersphinx mapping, which unrelated to changes in your PR, should be fixed in the master branch now

lisazeyen (approved):

lisazeyen:
@amosschle really nice, code looks super clean and well documented! We should use the same approach for other sectors as well.

amosschle:
**Note for future reference:**
- #857 suggests using Lorenz COPs. [Pieper et al. 2020](https://www.sciencedirect.com/science/article/abs/pii/S0360544220311014) suggests an approximation via Jensen et al. (implemented in this PR) to be superior (the authors test both and other methods against a thermodynamic simulation model for a Danish district heating system
- This PR does not touch household heat pump COPs. This is still approximated via [Staffell et al. 2012](https://pubs.rsc.org/en/content/articlelanding/2012/ee/c2ee22653g/unauth). An alternative would be using values/method from the [When2Heat dataset](https://www.nature.com/articles/s41597-019-0199-y). My preliminary analysis (to be investigated further) suggests somewhat similar COP values.

scripts/add_existing_baseyear.py
```diff
@@ -24,6 +24,10 @@
 from add_electricity import sanitize_carriers
 from prepare_sector_network import cluster_heat_buses, define_spatial, prepare_costs
 
+from scripts.definitions.heat_sector import HeatSector
+from scripts.definitions.heat_system import HeatSystem
+from scripts.definitions.heat_system_type import HeatSystemType
```
fneum, line 29: small thing, but I think one could just `from definitions.heat_sector import HeatSector`

amosschle: Thanks for spotting. Happy to change it in #1180

PR #1177, opened 2024-07-22
aggregate curtailment into single curtailment generator per bus

fneum:
Potential performance improvement.

Fixes the profile of VRE generators and adds a "curtailment generator" (flexible fictive load) analogous to load-shedding generator.

The model can no longer distinguish what carrier was curtailed.


fneum:
Testing indicates that this may help but not necessarily in all cases.

PR #1180, opened 2024-07-26
Add option of country-specific district heating supply temperatures

amos-schledorn:
Closes # (if applicable).

Allow country-specific forward and return temperatures in district heating:
- Make corresponding settings a dictionary with country codes as optional keys and `default` a necessary key. The `default` temperature is chosen for those countries not included in the dictionary.
- Update `build_cop_profiles.run` accordingly. Forward/return temperatures that enter COP approximations are changed from floats to data arrays with dimensions `time` and `name` (i.e. nodes).

## Checklist

## Notes
I've made a test run with the following settings (different from the proposed default settings):
```
scenario:
  clusters:
  - 4
  planning_horizons:
  - 2050


countries: ['DE', 'DK', 'PL']

sector:
  district_heating:
    forward_temperature:
      DE: 90
      DK: 70
      default: 80
    return_temperature:
      DE: 50
      DK: 40
      default: 45
```

As expected, this leads to a slightly higher overall heat pump share in district heating:
image

Plotting heat pump COPs for one node per country indicates that DE-COPs stay constant (same temperature as before), PL-COPs increase somewhat (slightly lower `default` supply temperatures) and DK-COPs increase substantially (much lower temperatures than before):
image

The proposed default supply temperatures are 70/40C in Scandinavia and 90/50C elsewhere. Admittedly, this is a guesstimate (also as a compromise between expected current and future temperatures), since open data seems very difficult to come by. I would suggest to open another issue for researching this data but allowing the option to use country-specific temperatures already now.


scripts/_helpers.py
```diff
@@ -785,3 +785,7 @@ def get_snapshots(snapshots, drop_leap_day=False, freq="h", **kwargs):
         time = time[~((time.month == 2) & (time.day == 29))]
 
     return time
+
+
+def get_country_from_node_name(node_name: str) -> str:
+    return node_name[:2]
```
lisazeyen, line 791: you can use the function [`assign_locations`](https://github.com/PyPSA/pypsa-eur/blob/f8d0efbe992413aac522851939dd7445f6084595/scripts/make_summary.py#L29)

scripts/build_cop_profiles/run.py
```diff
@@ -68,6 +147,23 @@ def get_cop(
 
     set_scenario_config(snakemake)
 
+    # map forward and return temperatures specified on country-level to onshore regions
+    regions_onshore = gpd.read_file(snakemake.input.regions_onshore)["name"]
```
lisazeyen, line 151: it is normally better to keep the snakemake object outside the function, so that one can call the function also from a different script

lisazeyen (changes requested):
Only some small comments. One more general question, is it a good approximation to use the same temperatures for different planning horizons and an eventually increasing district heating share? Or could one expect that temperatures could be reduced due to new generation DH systems? This does not need to be tackled in this PR.

scripts/build_cop_profiles/run.py
amosschle: Hm, not sure I can follow here. Isn't that what I'm doing? (`regions_onshore` is read from the snakmake input in the `__main__` block)

amosschle:
> Only some small comments. One more general question, is it a good approximation to use the same temperatures for different planning horizons and an eventually increasing district heating share? Or could one expect that temperatures could be reduced due to new generation DH systems? This does not need to be tackled in this PR.

Thanks for the review! I agree, we should probably (at least add the option to) decrease temperatures over time. I was a little hesitant to do that here because it introduces more structural changes (COPs would become year-specific). 
The current default temperatures are a compromise between what I believe to be current and future supply temperatures. 
Should I add the feature here on open a separate issue?

scripts/_helpers.py
amosschle: Yes, `assign_locations` seems to do what I need. Though it looks like it requires a `pypsa.Network` as input. Which network would I pass?

amosschle:
@lisazeyen, this is ready to go from my side. 
- I've marked your comment on the `assign_locations` method as resolved.
- I've merged the master branch into the feature and results seem fine.
- See my response to your open comment.

lisazeyen (approved):

PR #342, opened 2022-03-24
Automatically retrieve data directly from source

fneum:
Co-authored-by: euronion <42553970+euronion@users.noreply.github.com>

Closes # (if applicable).

## Changes proposed in this Pull Request

- [x] restore Windows compatibility at end
- [ ] delete zips because they are only temporary
- [ ] split Snakefile -> `workflow/rules/downloads.smk`

## Checklist

- [ ] I tested my contribution locally and it seems to work fine.
- [ ] Code and workflow changes are sufficiently documented.
- [ ] Newly introduced dependencies are added to `envs/environment.yaml` and `envs/environment.docs.yaml`.
- [ ] Changes in configuration options are added in all of `config.default.yaml`, `config.tutorial.yaml`, and `test/config.test1.yaml`.
- [ ] Changes in configuration options are also documented in `doc/configtables/*.csv` and line references are adjusted in `doc/configuration.rst` and `doc/tutorial.rst`.
- [ ] A note for the release notes `doc/release_notes.rst` is amended in the format of previous release notes.

PR #1190, opened 2024-08-02
naturalearth: automatically download and remove from data bundle

fneum:
related #342 

The data from Naturalearth misses a lot of smaller islands, even at 10m resolution. We should consider switching to GADM.


lisazeyen (approved):
Great! One thing less in the data bundle!

PR #1206, opened 2024-08-09
Dynamic central heating temperatures

amos-schledorn:
This PR introduces dynamic forward temperatures in central heating. It also adds the skeleton to use dynamic return temperatures, which are assumed constant for now.
The method is based on a reference curve from [Pieper et al. 2019](https://www.sciencedirect.com/science/article/pii/S0360544219305857?via%3Dihub), where for ambient temperatures below 0C, the highest possible forward temperature is assumed and vice versa for temperatures above 10C. Between these threshold levels, forward temperatures are linearly interpolated.
Temperature levels are increased for non-Scandinavian countries.

- Add a `build_central_heating_temperature_profiles` package with a `CentralHeatingTemperatureApproximator` class that takes ambient temperature profiles, lower and upper ambient temperature levels (0C/10C), max./min. forward temperatures and fixed return temperatures as inputs and returns dynamic forward and return temperatures
- Add `run` module in the same package to process this data and call the approximator class.
- Adjust the `build_cop_profiles` package accordingly
- Add necessary config settings

## Checklist

## Notes
Below config was used for testing:
```
scenario:
  clusters:
  - 3
  planning_horizons:
  - 2050


countries: ['DE', 'NL', 'PL']
```

Plots for temperature levels and COP profile for Germany seem reasonable:
Picture 1

Picture 1


amos-schledorn:
Based on feedback, I have introduced ambient temperature smoothing (by default in a rolling average of 72h window size). This better corresponds to real-world district heating system operation.

image

rules/build_sector.smk
```diff
@@ -212,17 +212,72 @@ rule build_temperature_profiles:
...
+            "district_heating",
+            "supply_temperature_approximation",
+            "upper_threshold_ambient_temperature",
+        ),
+        rolling_window_ambient_temperature=config_provider(
+            "sector",
+            "district_heating",
+            "supply_temperature_approximation",
+            "rolling_window_ambient_temperature",
+        ),
```
fneum, line 253: This is perfectly alright, but one could shorten this by using a whole config block and retrieving the individual settings from `snakemake.params.supply_temperature_approximation[attr]`. Not necessary to adjust.
```suggestion
        supply_temperature_approximation=config_provider(
            "sector",
            "district_heating",
            "supply_temperature_approximation",
        ),
        snapshots=config_provider("snapshots"),
```

fneum (approved):
Looks good! Also tested locally.

PR #1223, opened 2024-08-16
Ariadne2

lindnemi:
This is just to get a view of how much ariadne2 has diverged.

Eventually we will have to merge it.

## Checklist


lindnemi:
ah thanks, very intersting, i see.

However, if merged master into ariadne2 only 3 weeks ago, so shouldn't that be the common ancestor? there are many commits from even earlier showing up in the diff

lkstrp:
But earlier commits from `ariadne2`, not from the master

PR #1244, opened 2024-08-29
add option to vary parameter

lisazeyen:
.## Changes proposed in this Pull Request
Add option to modify any parameter by a setting in the config called `vary`
following the syntax `component` `carrier` `parameter` one can modify any static values of a component. This is helpful for creating different scenarios and also would allow to remove all quite a few settings in the config (e.g. `co2_network_cost_factor`, `aviation_demand_factor`, `gas_distribution_grid_cost_factor`,...)

## Checklist


fneum:
Doesn't that exist more or less?

https://pypsa-eur.readthedocs.io/en/latest/configuration.html#adjustments

lisazeyen:
I think like this it is more flexible, because you can vary everything, also the demand for example

koen-vg:
In my humble opinion I think it would a little confusing to have both `adjustments` and `vary` config sections - arguably there is already an intimidating number of different ways of adjusting the configuration.

lisazeyen:
Thanks a lot for your suggestions @koen-vg  and @fneum . I still think it is useful to be able to more flexibly modify the components and attributes. For example, with the current implementation in the master branch the load cannot be modified and only specific attributes. Also it is not possible to have a modification depending on the investment period, which can be very useful. I have adapted now the existing function `maybe_adjust_costs_and_potentials` to work more flexibly, including the option to vary by a factor or to set an absolute value. Let me know, what you think about this!

koen-vg:
I definitely see the appeal of being able to adjust any attribute, also depending on the investment horizon (and choosing relative or absolute adjustments)! I think it made a lot of sense to build this into the `adjustments` config.

If you ask me, I would at this point just get rid of the `maybe_adjust_costs_and_potentials` function (calling `modify_attribute` directly), since it really only serves as thin scaffolding to `modify_attribute`. Can just be less confusing to people that don't know the history of the `maybe_adjust_costs_and_potentials` function. But even without this change the PR looks good to me :)

fneum (approved):

PR #1247, opened 2024-08-30
prepare_sector_network.py: add oil bus whenever industry is added

p-glaum:
I got the warning that many links like "naphtha for industry" or "Fischer-Tropsch" had missing bus "EU oil" because it was not declared. Usually, it is always declared because in shipping it was added and also for ICE vehicles. However, when I do an overnight optimization with the time horizon 2050, then we do not have oil demand for shipping and ICE, and the oil bus is not added which led to the warning.

## Checklist


fneum (commented):
Great, could you move this to the first location where the oil bus is used in `prepare_sector_network`? Trying to stick to the convention that buses should be defined before using them.

p-glaum:
> Great, could you move this to the first location where the oil bus is used in `prepare_sector_network`? Trying to stick to the convention that buses should be defined before using them.

So you mean I add the oil bus in the transport sector, even if ICEs are not considered?

p-glaum:
what one could also do is restructure:
```python
  if snakemake.params.foresight in ["myopic", "perfect"]:
      add_lifetime_wind_solar(n, costs)

      conventional = snakemake.params.conventional_carriers
      for carrier in conventional:
          add_carrier_buses(n, carrier)
```
to 
```python
    if snakemake.params.foresight in ["myopic", "perfect"]:
        add_lifetime_wind_solar(n, costs)

      conventional = snakemake.params.conventional_carriers
      for carrier in conventional:
          add_carrier_buses(n, carrier)
```

p-glaum:
If I do it with a condition whether an option like "biomass_to_liquid" is considered, I would need to add three if clauses in add_biomass

fneum:
Maybe then at top of `add_industry()` we need to rework this anyway, but adding the oil bus in between the shipping is a bit too hidden.

p-glaum:
I moved oil and methnaol to the top of add_industry, because they are always needed. To avoid the warning that bus does not exist in biomass, I added add_carrier_buses to all options.

p-glaum:
> Maybe then at top of `add_industry()` we need to rework this anyway, but adding the oil bus in between the shipping is a bit too hidden.

I agree that we need to rework this in a smart way

fneum (approved):

PR #1250, opened 2024-08-30
prepare_sector_network: fix municipal waste transport links

p-glaum:
Closes # (if applicable).

The problem was that the municipal waste transports links had the same index as solid biomass, and they were not added. With the new index for the municipal waste transports links, one can only take the values of the biomass_transport df, because the link indices changed.

## Checklist


fneum:
Thanks for fixing!

PR #1139, opened 2024-07-08
Addition of unsustainable biomass potentials

cpschau:
This PR adds unsustainable solid, gaseous, and liquid biomass sources to the model. It is assumed that unsustainable sources are currently used almost exclusively. Therefore, the potential is derived from Eurostat data on biomass primary production. The unsustainable potentials are gradually phased out until 2035, while the formerly used sustainable biomass potentials from the ENSPRESO database are phased in simultaneously.

unsustainable_biomass

## Checklist

- [x] I tested my contribution locally and it seems to work fine.
- [x] Code and workflow changes are sufficiently documented.
- [x] Changed dependencies are added to `envs/environment.yaml`.
- [x] Changes in configuration options are added in all of `config.default.yaml`.
- [x] Changes in configuration options are also documented in `doc/configtables/*.csv`.
- [x] A release note `doc/release_notes.rst` is added.

PR #1254, opened 2024-09-02
Follow-Up: Unsustainable Biomass

cpschau:
1. Using the unsustainable biomass potentials introduced in #1139, the model becomes infeasible for setups with spatially resolved biomass demand because no statistics are provided for Switzerland in the Eurostat data. Consequently, the country's unsustainable biomass potential cannot be determined and is set to zero. As a result, for networks with deactivated solid biomass transport and zero sustainable biomass potentials, there is no source to cover the unavoidable solid biomass load ("solid biomass for industry"). With this PR, the sustainable biomass potentials for Switzerland are retained in build_biomass_potentials, independent of the configuration setting biomass:share_sustainable_potential_available.
2. The global biomass constraint, which sets the biomass potentials as the upper bound for solid biomass usage in the model within prepare_sector_network, has been adjusted to include the unsustainable potentials.
3. In prepare_sector_network, oil buses are added when incorporating unsustainable bioliquids. Otherwise, there is a risk that the bioliquid-to-oil link may use undefined buses.
4. For overnight scenarios, the biomass potentials are adjusted to include only sustainable sources in build_biomass_potentials.

### DIFF
--- a/scripts/prepare_sector_network.py
+++ b/scripts/prepare_sector_network.py
@@ -2465,20 +2465,37 @@ def add_biomass(n, costs):
             e_max_pu=e_max_pu,
         )
 
-        e_max_pu = pd.DataFrame(1, index=n.snapshots, columns=spatial.biomass.nodes)
-        e_max_pu.iloc[-1] = 0
+        n.madd(
+            "Bus",
+            spatial.biomass.nodes,
+            suffix=" unsustainable",
+            location=spatial.biomass.locations,
+            carrier="unsustainable solid biomass",
+            unit="MWh_LHV",
+        )
 
         n.madd(
             "Store",
             spatial.biomass.nodes,
             suffix=" unsustainable",
-            bus=spatial.biomass.nodes,
+            bus=spatial.biomass.nodes + " unsustainable",
             carrier="unsustainable solid biomass",
             e_nom=unsustainable_solid_biomass_potentials_spatial,
             marginal_cost=costs.at["fuelwood", "fuel"],
             e_initial=unsustainable_solid_biomass_potentials_spatial,
             e_nom_extendable=False,
             e_max_pu=e_max_pu,
+        )
+
+        n.madd(
+            "Link",
+            spatial.biomass.nodes,
+            suffix=" unsustainable",
+            bus0=spatial.biomass.nodes + " unsustainable",
+            bus1=spatial.biomass.nodes,
+            carrier="unsustainable solid biomass",
+            efficiency=1,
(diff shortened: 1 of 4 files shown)


lisazeyen:
Great @cpschau!

I think the unsustainable biomass potential should not depend on the foresight and only on your planning horizon. So also for an overnight scenario, it would be good, to take the unsustainable biomass potential defined in the config.

Can you add a plot of the unsustainable potential of each country? Are all other countries, like e.g. Norway and Great Britain covered?

You have a separate carrier unsustainable biomass so you also need a separate Global Constraint for the operational limit I suppose [here](https://github.com/PyPSA/pypsa-eur/blob/06dc60db11357e81f548fa7e813d288697682789/scripts/prepare_sector_network.py#L2630-L2637)

cpschau:
> Great @cpschau!
> 
> I think the unsustainable biomass potential should not depend on the foresight and only on your planning horizon. So also for an overnight scenario, it would be good, to take the unsustainable biomass potential defined in the config.
> 
> Can you add a plot of the unsustainable potential of each country? Are all other countries, like e.g. Norway and Great Britain covered?
> 
> You have a separate carrier unsustainable biomass so you also need a separate Global Constraint for the operational limit I suppose [here](https://github.com/PyPSA/pypsa-eur/blob/06dc60db11357e81f548fa7e813d288697682789/scripts/prepare_sector_network.py#L2630-L2637)

Thanks for the quick review @lisazeyen! Here are the potentials for 2020 (only unsustainable except for Switzerland) and 2050 (only sustainable). Norway and UK are covered in the data.

2020:
biomass_potentials_unsus
2050:
biomass_potentials_sus

I will remove the distinction of the overnight potentials, I do not remember what I was thinking when I added it tbh :D.

Additionally, I'll introduce a global constraint for the operational limit on unsustainable biomass. I believe it's also necessary to add an extra generator for unsustainable biomass in case we decide to spatially resolve the biomass demand without allowing for biomass transport.

scripts/prepare_sector_network.py
```diff
@@ -2631,7 +2648,7 @@ def add_biomass(n, costs):
             "GlobalConstraint",
             "biomass limit",
             carrier_attribute="solid biomass",
-            sense="<=",
+            sense="==",
```
lindnemi, line 2651: this stays

lindnemi:
bitte noch `unsustainable solid biomass: '#998622'` zu 'tech_colors:' hinzufügen

fneum (approved):

Issue #540, opened 2020-11-24
Switch building heating demand from final energy consumption to thermal energy services in JRC-IDEES


nworbmot:
Since we calculate our own losses for e.g. gas boilers, heat pumps and district heating, we should use the services demand, not final energy from JRC-IDEES. But I'm not sure what factor to use for pipe losses inside the buildings.

PR #1255, opened 2024-09-02
Use JRC-IDEES thermal energy service instead of FE for buildings heating demand

nworbmot:
This change consists of:

- reading the final energy (FE) to thermal energy service (TES) efficiency for each each country from JRC-IDEES, separately for gas and oil (this efficiency represents e.g. the losses in older non-condensing boilers, averaging around 20% in 2020)

- the efficiencies are stored in `resources/heating_efficiencies.csv` for different years and countries

- using TES instead of FE for the heat demand in n.loads_t, so that it is pure energy service instead of including losses in legacy equipment

- using the stored boiler efficiencies for baseyear equipment in `add_existing_baseyear.py`

In the baseyear (e.g. 2020) this should have no effect on FE, since the reduction to TES is compensated by the lower efficiencies of existing equipment.

In the future (e.g. 2050) this leads to a reduction in heating demand, since new equipment is more efficient than existing.

The following documents changes to results for a 45-node 3H simulation with the myopic code and investment years 2020 and 2050.

The change reduces the heat demand by 19% in 2020 and 20% in 2050:

image

The fraction of heat provided by gas boilers actually goes down, since some is substituted by biomass boilers and heat pumps:

image

image

Gas boiler average efficiency drops from 97% to 78%, while oil boiler average efficiency drops from 90% to 62%.

This means that the effects of lower demand by 19% but including boiler losses better doesn't quite cancel out, because some of the gas demand is eaten by biomass boilers and heat pumps. Here is the gas boiler gas consumption:

image

The change reduced the objective function in 2020 by 2.3% and in 2050 by 9.1%.

Closes [#540](https://github.com/PyPSA/pypsa-eur/issues/540).

## Checklist


lisazeyen:
one thought @nworbmot : in `prepare_sector_network` the thermal energy service (TES) is added as the heating demand. This heating demand is split into individual and district heating. In `add_existing_baseyear` the efficiencies for individual heating for existing assets are used, while for district heating the efficiencies of the technology data are used. This lead for district heating networks in e.g. 2020 that final energy consumption from JRC-IDEES is unequal to our assumptions.

nworbmot:
I agree this will lead to a ~10% underestimation of district heating demand, which we should fix at a later point, but we should pull this PR now so we get the benefit for individual heating (~20% reduction of FE in 2050). 20% is a bigger effect that 10%*10% (error times 10% DH share)  = 1%. I'm trying to think what the best way is to solve this really properly - we need to probably separate out the DH share at the beginning somehow.

lisazeyen (approved):

scripts/add_existing_baseyear.py
```diff
@@ -660,6 +710,11 @@ def add_heating_capacities_installed_before_baseyear(
 
     if options["heating"]:
 
+        # one could use baseyear here instead (but dangerous if no data)
+        heating_efficiencies = (
+            pd.read_csv(snakemake.input.heating_efficiencies, index_col=[1, 0])
+        ).loc[int(snakemake.config["energy"]["energy_totals_year"])]
```
fneum, line 716: `snakemake.config` should no longer be used in `scripts/*.py` files due to new scenario management.

```suggestion
        fn = snakemake.input.heating_efficiencies
        year = int(snakemake.params["energy_totals_year"])
        heating_efficiencies = pd.read_csv(fn, index_col=[1, 0]).loc[year]
```

And then add to `add_existing_baseyear` rule definition:

```
params:
    energy_totals_year=config_provider("energy", "energy_totals_year")
```

scripts/build_energy_totals.py
```diff
@@ -654,11 +694,16 @@ def build_energy_totals(
...
-    to_drop = ["passenger cars", "passenger car efficiency"]
     new_index = pd.MultiIndex.from_product(
         [countries, eurostat_years], names=["country", "year"]
     )
 
+    to_drop = idees.columns[
+        idees.columns.str.contains("space efficiency")
+        ^ idees.columns.str.contains("water efficiency")
+    ]
+    to_drop = to_drop.append(pd.Index(["passenger cars", "passenger car efficiency"]))
```
fneum, line 705: simpfliy
```suggestion
    efficiency_keywords = ["space efficiency", "water efficiency"]
    to_drop = idees.columns[~idees.columns.str.contains('|'.join(efficiency_keywords))]
    to_drop = to_drop.append(["passenger cars", "passenger car efficiency"])
```

scripts/build_energy_totals.py
```diff
@@ -576,7 +612,11 @@ def build_idees(countries: List[str]) -> pd.DataFrame:
     # efficiency kgoe/100km -> ktoe/100km so that after conversion TWh/100km
     totals.loc[:, "passenger car efficiency"] /= 1e6
     # convert ktoe to TWh
-    exclude = totals.columns.str.fullmatch("passenger cars")
+    exclude = (
+        totals.columns.str.fullmatch("passenger cars")
+        ^ totals.columns.str.fullmatch(".*space efficiency")
+        ^ totals.columns.str.fullmatch(".*water efficiency")
+    )
```
fneum, line 619:
```suggestion
    patterns = ["passenger cars", ".*space efficiency", ".*water efficiency"]
    exclude = totals.columns.str.fullmatch("|".join(patterns))
```

scripts/build_energy_totals.py
```diff
@@ -1501,6 +1546,61 @@ def build_transformation_output_coke(eurostat, fn):
...
+    """
+
+    years = np.arange(2000, 2022)
+
+    cols = idees.columns[
+        idees.columns.str.contains("space efficiency")
+        ^ idees.columns.str.contains("water efficiency")
+    ]
+
+    logger.info(cols)
```
fneum, line 1580:
```suggestion
```

remove or more descriptive logging info

scripts/prepare_sector_network.py
```diff
@@ -4373,6 +4378,12 @@ def add_enhanced_geothermal(n, egs_potentials, egs_overlap, costs):
     )
     pop_weighted_energy_totals.update(pop_weighted_heat_totals)
 
+    heating_efficiencies = (
+        (pd.read_csv(snakemake.input.heating_efficiencies, index_col=[0, 1]))
+        .swaplevel()
+        .loc[int(snakemake.config["energy"]["energy_totals_year"])]
+    )
```
fneum, line 4385: similar as above; no `snakemake.config` due to scenario management.

```suggestion
    fn = snakemake.input.heating_efficiencies
    year = int(snakemake.params["energy_totals_year"])
    heating_efficiencies = pd.read_csv(fn, index_col=[0, 1]).swaplevel().loc[year]
```

And add to `perpare_sector_network` rule definition:

```
params:
    energy_totals_year=config_provider("energy", "energy_totals_year")
```

scripts/add_existing_baseyear.py
```diff
@@ -418,6 +418,48 @@ def add_power_capacities_installed_before_baseyear(n, grouping_years, costs, bas
...
+
+    if heat_system.sector.value == "residential":
+        key = f"{carrier} residential space efficiency"
+        efficiency = nodes.str[:2].map(heating_efficiencies[key])
+    elif heat_system.sector.value == "services":
+        key = f"{carrier} services space efficiency"
+        efficiency = nodes.str[:2].map(heating_efficiencies[key])
+    else:
+        boiler_costs_name = getattr(heat_system, f"{carrier}_boiler_costs_name")
+        efficiency = costs.at[boiler_costs_name, "efficiency"]
```
fneum, line 458: Throws error in CI:

```
     if heat_system.sector.value == "residential":
       ^^^^^^^^^^^^^^^^^^^^^^^^
AttributeError: 'NoneType' object has no attribute 'value'
```

fneum (changes requested):
Semantically, looks good to me. Some code fixes required before merging:

PR #1259, opened 2024-09-05
Make overdimensioning factor for heating systems specific to central/decentral heating

amos-schledorn:
Currently, the factor `overdimension_individual_heating = 1.1` is applied to both central and decentral heating for adequate investment decisions at heat demand above model input. This disadvantages CAPEX-heavy technologies like heat pumps and is probably not necessary, at least for central heating, due to heat storage and the possibility to invest in inexpensive boilers.

- The config setting `sector/overdimension_individual_heating` is renamed to `sector/overdimension_heat_generators` made heat-system specific allowing to set different values for central and decentral heating:
```
 overdimension_heat_generators: 
    decentral: 1.1  #to cover demand peaks bigger than data
    central: 1.0
```
- It defaults to 10% overdimensioning for decentral and none for central heating

## Checklist

## Testing
In a simple test case (Germany, 1 node)
```
scenario:
  clusters:
  - 1
  planning_horizons:
  - 2030
countries: ['DE']
```

Urban central heat mix increases in favour of heat pumps:

Master:
image

Feature:
image

### DIFF
--- a/config/config.default.yaml
+++ b/config/config.default.yaml
@@ -590,7 +590,7 @@ sector:
   resistive_heaters: true
   oil_boilers: false
   biomass_boiler: true
-  overdimension_heat_generators: 
+  overdimension_heat_generators:
     decentral: 1.1  #to cover demand peaks bigger than data
     central: 1.0
   chp: true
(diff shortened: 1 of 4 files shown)


nworbmot (approved):
Looks good to me! Perhaps @lisazeyen should also double-check. It was stupid before that it was called `overdimension_inidividual_heating` then applied also to district heating (my fault).

fneum (approved):

PR #1275, opened 2024-09-11
Simplify integration of unsustainable solid biomass

cpschau:
This PR attaches the Store for unsustainable solid biomass directly to the (sustainable) solid biomass nodes, eliminating the need for an extra Bus and Link component. This change is enabled by the new operational limit definition of the GlobalConstraint component, addressed in [PyPSA PR #1029](https://github.com/PyPSA/PyPSA/pull/1029). Therefore, some changes introduced in #1254 are now obsolete.

### DIFF
--- a/scripts/prepare_sector_network.py
+++ b/scripts/prepare_sector_network.py
@@ -2477,7 +2477,6 @@ def add_biomass(n, costs):
         )
 
     if biomass_potentials.filter(like="unsustainable").sum().sum() > 0:
-        add_carrier_buses(n, "oil")
         # Create timeseries to force usage of unsustainable potentials
         e_max_pu = pd.DataFrame(1, index=n.snapshots, columns=spatial.gas.biogas)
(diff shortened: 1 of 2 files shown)


scripts/prepare_sector_network.py
```diff
@@ -2477,7 +2477,6 @@ def add_biomass(n, costs):
         )
 
     if biomass_potentials.filter(like="unsustainable").sum().sum() > 0:
-        add_carrier_buses(n, "oil")
```
fneum, line 2480: intended?

fneum (changes requested):
LGTM. Can you also update `envs/environment.yaml` to require at least version v0.30.2 for PyPSA? Then we can merge once the PyPSA conda feedstock is updated.

scripts/prepare_sector_network.py
cpschau: Yes, was redundant to the addition in line 2539.

PR #1279, opened 2024-09-11
remove `data/hydro_capacities.csv` with untraceable source

fneum:
- The source is untraceable.
- The storage capacities seem dubiously high. Better to be conservative.
- The `powerplantmatching` has ok coverage of storage capacities (~40%, many of the larger hydro reservoirs, best coverage in Norway)
- Use 10% quantile of storage capacities as fill-value. Is this a good number? For all of Europe, 40 hours.
- Use 24 hours of storage duration as backstop.

Distribution of storage durations (h)

image


fneum:
Thanks @koen-vg, @lisazeyen and I decided to put this on hold for the moment.

The motivation was indeed to remove data without a source.

We have asked some more hydro-experts we know, whether they can help.

fneum:
Potential alternative source: https://www.mdpi.com/1996-1073/10/11/1841

PR #1049, opened 2024-05-03
draft bot for automated fixed env yaml

FabianHofmann:
this creates a bot that will automatically create a PR to update the fixed env file as soon as there are changes in `env/environment.yaml`.

PR #1281, opened 2024-09-11
fix: env update trigger from #1049

lkstrp:
- This should fix #1049 which was never working
- Also the `envs/environment.fixed.yaml` is empty right now?
- And we could think about renaming that to `pinned`

### DIFF
--- a/.github/workflows/update-fixed-env.yaml
+++ b/.github/workflows/update-fixed-env.yaml
@@ -24,7 +24,7 @@ jobs:
     - name: Update environment.fixed.yaml
       run: |
         conda env export --name pypsa-eur --no-builds > envs/environment.fixed.yaml
-        
+
     - name: Add SPDX header
       run: |
         SPDX_HEADER="# SPDX-FileCopyrightText: : 2021-2024 The PyPSA-Eur Authors\n# SPDX-License-Identifier: CC0-1.0\n"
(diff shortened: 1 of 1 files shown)


.github/workflows/update-fixed-env.yaml
```diff
@@ -1,39 +1,34 @@
...
         environment-file: envs/environment.yaml
-        log-level: debug
-        init-shell: bash
-        cache-environment: true
-        cache-downloads: true
 
     - name: Update environment.fixed.yaml
       run: |
-        mamba env export --file envs/environment.fixed.yaml  --no-builds
+        conda env export --name pypsa-eur > envs/environment.fixed.yaml
```
fneum, line 26: Wouldn't it be better with `--no-builds`?

.github/workflows/update-fixed-env.yaml
```diff
@@ -1,39 +1,34 @@
-name: Fixed Environment YAML Monitor
+name: Environment Checker
```
fneum, line 1:
```suggestion
name: Fixed-Version Environment Checker
```

fneum (changes requested):
Fine to merge, perhaps `--no-builds`?

.github/workflows/update-fixed-env.yaml
lkstrp: Thanks!

PR #1287, opened 2024-09-12
fix load weighted prices

lisazeyen:
Fix bug when calculating load weighted prices in the `make_summary` script

## Checklist


lisazeyen:
@millingermarkus can you check if it is working with this branch?

scripts/make_summary.py
```diff
@@ -540,77 +540,21 @@ def calculate_prices(n, label, prices):
...
-                )
-
-        # Add H2 Store when charging
-        # if carrier == "H2":
-        #    stores = n.stores_t.p[buses+ " Store"].groupby(n.stores.loc[buses+ " Store", "bus"],axis=1).sum(axis=1)
-        #    stores[stores > 0.] = 0.
-        #    load += -stores
+    for carrier in carriers:
+        grouper = n.statistics.groupers.get_bus_and_carrier
+        balance = n.statistics.energy_balance(
```
p-glaum, line 548: you can directely use withdrawal, then you do not need to wilter for values <0

scripts/make_summary.py
```diff
@@ -540,77 +540,21 @@ def calculate_prices(n, label, prices):
...
+            groupby=grouper, aggregate_time=False, nice_names=False, bus_carrier=carrier
+        )
+        load = (balance[balance < 0].groupby(level=1).sum()).T
```
p-glaum, line 551: better to do level="bus"

scripts/make_summary.py
lisazeyen: done!

scripts/make_summary.py
lisazeyen: Done!

p-glaum:
> @p-glaum I am not dropping the prices if the load is zero at a bus. `price = price.reindex(columns=load.columns)` One could also reindex the other way around, but then the mariginal price would not be really load weighted, e.g. with `load.reindex(columns=price.columns, fill_value=1)`

okay, if the prices have a meaning if there is no load at the bus, then we should definitely keep them and weight every timestep equally, i.e. wit `load.reindex(columns=price.columns, fill_value=1)`
However, if they do not have a realistic interpretation, should we drop them?

p-glaum (approved):

lisazeyen:
> > @p-glaum I am not dropping the prices if the load is zero at a bus. `price = price.reindex(columns=load.columns)` One could also reindex the other way around, but then the mariginal price would not be really load weighted, e.g. with `load.reindex(columns=price.columns, fill_value=1)`
> 
> okay, if the prices have a meaning for buses flows going in or out, then we should definitely keep them and weight every timestep equally, i.e. wit `load.reindex(columns=price.columns, fill_value=1)` However, if they do not have a realistic interpretation, should we drop them?

I think if it is load weighted and the load is zero we should drop them.

fneum:
> I think if it is load weighted and the load is zero we should drop them.

I think equal weight would be slightly better, but dropping also ok.

fneum (approved):

PR #1294, opened 2024-09-13
add validator report

lkstrp:
Adds Validator CI. See the report below.
- It is triggered by any new commit in pr.
- Runs on a self-hosted github runner and solves on z1, so basically any configuration can be set. No restrictions. It just takes a while to update the comment if you choose a computationally intensive config.
- Compares the base branch (main) with the latest commit in the pr branch. So runs both, but also caches (in addition to snakemake caching) based on commit hashes. -  Edits one main report comment when new commits are pushed to the repo, to avoid spamming.
- Set all settings in workflow file (mainly which plots to show, pre-commands, environment selection, etc.)

- Some things which are missing:
  - Better comparison (use `n.statistics`)
  - Handle environment updates
  - More file filters to avoid unnecessary runs
  - Concurrency

Also have a look at the [Wishlist](https://github.com/lkstrp/pypsa-validator/issues/10) and see action [repo](https://github.com/lkstrp/pypsa-validator/issues) to create any issues. This will be moved to the PyPSA org at some point.

### DIFF
--- a/.github/workflows/validate.yaml
+++ b/.github/workflows/validate.yaml
@@ -1,0 +1,48 @@
+name: Validator Bot
+
+on:
+  pull_request:
+    branches:
+    - master
+    - main
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
+
+jobs:
+  run-validation:
+    name: Run validation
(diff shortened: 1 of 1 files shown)


lkstrp:
See #1295

PR #1296, opened 2024-09-16
test: fix and update test workflows

lkstrp:
- Updates test ci yaml with some smaller changes
- Fixes the failing test on macos (hot fix though, couldn't find another solution)
- Also triggers #1281 the first time


fneum (approved):
Looks good, but maybe keep test.sh as I like to use it for local tests as well.

Issue #1306, opened 2024-09-17
Excluding certain countries is no longer possible


koen-vg:
## Checklist

- [x] I am using the current [`master`](https://github.com/PyPSA/pypsa-eur/tree/master) branch or the latest [release](https://github.com/PyPSA/pypsa-eur/releases). Please indicate.
- [x] I am running on an up-to-date [`pypsa-eur` environment](https://github.com/PyPSA/pypsa-eur/blob/master/envs/environment.yaml). Update via `conda env update -f envs/environment.yaml`.

## Describe the Bug

Running PyPSA-Eur with a subset of supported countries such that non-included countries are surrounded by included countries is no longer supported. You might think that that's not really a problem (why would you exclude those countries in the first place?) but in some cases, such as when optimising over weather years for which only synthetic load data is available, it become necessary to exclude Kosovo (for which there is no synthetic load data).

The root cause seems to be that when buses are loaded in `base_network.py` (in `_load_buses`), all buses that are located inside the "europe_shape" are included; however, this shape, due to the way it is constructed using the `exterior` function, can include more countries than those specified in the config (for example, Kosovo).

Somehow in the transition to OSM networks, the logic for excluding buses that are not in the specified list of countries seems to have been lost. Maybe in `_set_countries_and_substations`? (Note that buses from the pre-built OSM network already include country annotations, so this function seems less necessary anyway.) I'm not super familiar with the `base_network` rule, so I haven't quite figured it out.

## Error Message
The actual errors arise only in `cluster_network`, when it turns out that `base_s.nc` includes buses in countries that are not given in the configuration.

## Possible fix
I have fixed this on my end with a little bit of logic for excluding buses outside the configured countries as soon as possible, in `_load_buses`. The following changes do the trick:
````diff
@@ -135,7 +135,7 @@ def _find_closest_links(links, new_links, distance_upper_bound=1.5):
     )
 
 
-def _load_buses(buses, europe_shape, config):
+def _load_buses(buses, europe_shape, countries, config):
     buses = (
         pd.read_csv(
             buses,
@@ -161,6 +161,11 @@ def _load_buses(buses, europe_shape, config):
         lambda p: europe_shape_prepped.contains(Point(p)), axis=1
     )
 
+    if "country" in buses.columns:
+        buses_in_countries = buses.country.isin(countries)
+    else:
+        buses_in_countries = pd.Series(True, buses.index)
+
     v_nom_min = min(config["electricity"]["voltages"])
[...]

PR #1308, opened 2024-09-18
Only add buses in specified countries

koen-vg:
This is necessary for the new OSM-based base network, where loaded buses already have a "country" attribute.

Closes https://github.com/PyPSA/pypsa-eur/issues/1306

## Checklist

### DIFF
--- a/doc/release_notes.rst
+++ b/doc/release_notes.rst
@@ -73,6 +73,8 @@ Upcoming Release
 
 * Added 98% imperfect capture rate of Allam cycle gas turbine.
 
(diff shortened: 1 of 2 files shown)


koen-vg:
If I understand the intention of dropping non-OSM support correctly, the if-statement checking if "countries" is a column in `buses` can be simplified once OSM is the only source. I could add a note to that effect, but it seems to me like a lot of code can be simplified at that point, so hopefully that should be obvious at that point?

fneum (approved):

PR #1311, opened 2024-09-19
consider the planning horizon for biomass potentials input

p-glaum:
build_sector.smk: always consider the planning horizon for biomass potentials input in prepare_sector_network

The problem was that for overnight, not the planning horizon was considered for the biomass potential but the year set in the config under `biomass`. This led to the unwanted behavior that one could not do a 2050 optimization with the biomass potentials from ENSPRESO for 2030 and without unsustainable biomass. 

## Checklist

### DIFF
--- a/rules/build_sector.smk
+++ b/rules/build_sector.smk
@@ -1059,14 +1059,7 @@ rule prepare_sector_network:
         dsm_profile=resources("dsm_profile_s_{clusters}.csv"),
         co2_totals_name=resources("co2_totals.csv"),
         co2="data/bundle/eea/UNFCCC_v23.csv",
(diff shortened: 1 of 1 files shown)


fneum:
This is the intended behavior that you can use 2030 biomass potentials for 2050 runs.

fneum:
Alright, I think I got it now. Will you do the adjustment we talked about, using planning horizon as key and ENSPRESO potential year as values in the `config.yaml`? Then, we can merge.

```yaml
biomass:
  year:
    2020: 2020
    2025: 2025
    2030: 2030
    ...
```

fneum (approved):

PR #1312, opened 2024-09-19
Fix industrial demand for ammonia when endogenously modelled

koen-vg:
In calculating industrial energy demand today, demand for ammonia must be resolved into just "ammonia" when endogenously modelled, or into electricity and hydrogen if not. Before this fix, energy demand for ammonia was double-counted at current and near-term planning horizons when `sector['ammonia'] = True`; both in terms of ammonia and in terms of electricity and hydrogen.

While not incredibly large in magnitude, hydrogen demand at 2025 planning horizon was essentially double what it should have been before this fix.

## Checklist

### DIFF
--- a/doc/release_notes.rst
+++ b/doc/release_notes.rst
@@ -75,6 +75,8 @@ Upcoming Release
 
 * Resolved a problem where excluding certain countries from `countries` configuration led to clustering errors.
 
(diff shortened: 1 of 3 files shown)


fneum:
Great! I'll do a small follow-up so that tests for Windows and MacOS run through (the `snakemake` object is lost in the multiprocessing). The dev test is failing because of some breaking changes in `powerplantmatching` that need to be addressed in PyPSA-Eur, so that's ok.

PR #1314, opened 2024-09-20
fix: make `test.sh` executable

lkstrp:


fneum (approved):

FabianHofmann:
in pypsa-earth, we now have a Makefile which is very convenient for all the meta rules. might want to do the same here

PR #1315, opened 2024-09-20
test: use makefile

lkstrp:
Anything else what is useful and we can add here?


FabianHofmann:
what we could optionally add is 

* make install (installs environment, activates it and installs pre-commit)
* make fix_environment

PR #1316, opened 2024-09-20
prepare for new `n.add`

lkstrp:
- Adds some more explicit indexing in`prepare_sector_network.py` since non matching indices are not allowed with new PyPSA version
- See https://github.com/PyPSA/PyPSA/pull/896

### DIFF
--- a/scripts/prepare_sector_network.py
+++ b/scripts/prepare_sector_network.py
@@ -1828,8 +1828,6 @@ def add_EVs(
 
     profile = electric_share * p_set.div(efficiency)
 
-    # breakpoint()
-
     n.madd(
         "Load",
         spatial.nodes,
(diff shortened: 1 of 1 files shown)


scripts/prepare_sector_network.py
```diff
@@ -1828,13 +1828,15 @@ def add_EVs(
 
     profile = electric_share * p_set.div(efficiency)
 
+    # breakpoint()
+
```
fneum, line 1832:
```suggestion
```

fneum (approved):
Looks good! So the new `n.add` would throw a warning if indices or snapshots for time-varying data don't match?

What about switching from `n.madd` to `n.add`? That was also planned, right?

lkstrp:
> Looks good! So the new `n.add` would throw a warning if indices or snapshots for time-varying data don't match?

Error, not a warning. So the next `pypsa` release is breaking. Slightly, but breaking.

> What about switching from `n.madd` to `n.add`? That was also planned, right?

Yes, but this will fail for `pypsa<=30.2`. So once we set the dependency here on `pypsa>=31.0` we can replace all `n.madd` (and `n.madd` will just point to `n.add` with a Deprecation Warning`).

PR #1322, opened 2024-09-23
fix: move pypsa dependency to pip

Uzaaft:
Tried to set this project up, and seems like conda is experiencing some difficulties due to the following:
- pypsa is installed with conda, and requires highspy
- highspy is not available on conda for arm64 on osx
- highspy is being installed in the environment.yml file, but with pip

Seems like this causes conda to error out with highspy not being available to install.

### DIFF
--- a/envs/environment.yaml
+++ b/envs/environment.yaml
@@ -11,7 +11,6 @@ dependencies:
 - pip
 
 - atlite>=0.2.9
-- pypsa>=0.30.2
 - linopy
 - dask
 
@@ -63,3 +62,4 @@ dependencies:
   - snakemake-executor-plugin-slurm
   - snakemake-executor-plugin-cluster-generic
   - highspy
+  - pypsa>=0.30.2


fneum:
Fine as a temporary measure, but I would like to understand where the problem comes from. Maybe something for @lkstrp?

lkstrp:
Same. It is the same issue as the previously failing ci and only happens on arm macos. And I think related to a new [highspy build](https://anaconda.org/conda-forge/highspy/files). Will revert this, once this is solved.

lkstrp:
@Uzaaft 
The issue is fixed now. See #1325. You might need to remove your env and clean the cache of conda/ mamba.

Uzaaft:
The issue seems to come from the fact that highspy in conda doesn't have osx-arm64 support, even if the pip variant has.
Previously, pypsa-eur pulled in pypsa from conda, and its dependency graph looked after highspy in conda, and would fail on osx-arm64. 

But pypsa on pip was fine since pip has a working version. 

At least that's my understanding

lkstrp:
I think that's exactly it. If you install pypsa 30.2 with the build version mentioned in the PR it will work, since the recipe is changed.

PR #1328, opened 2024-09-24
Remove sector_opts and duplicated CHP plants

cpschau:
This PR replaces the references of the formerly used `sector_opts` wildcard by referring to the corresponding config parameters in the `sector` section instead. Additionally, German CHP plants from the PPM dataset are removed if CHP plants from the MaStR dataset are added as custom data, in order to avoid duplication.


scripts/add_existing_baseyear.py
```diff
@@ -492,6 +489,7 @@ def add_chp_plants(n, grouping_years, costs, baseyear, clustermaps):
     ppl = pd.read_csv(snakemake.input.powerplants, index_col=0)
 
     if snakemake.input.get("custom_powerplants"):
+        ppl = ppl.query("~(Set == 'CHP' and Country == 'DE')")
```
lindnemi, line 492: This line implicitly assumes that the custom_powerplants supersede the data for CHPs in DE.

However, if the data in custom_powerplants changes, this might no longer be true. I would like to see a more explicit solution that applies this filter only, when the mastr_chps are read in.

This is important since add_existing is a pypsa-eur script and soon we will merge ariadne2 completly into the master branch.

PR #1366, opened 2024-10-16
Keep existing conventional generation in sector network

martacki:
Closes # (if applicable).

The sector coupled network currently removes existing conventional generators. This PR adds them back (in the traditional sectoral setup with EU Generators + Links)

## Checklist

### DIFF
--- a/scripts/prepare_sector_network.py
+++ b/scripts/prepare_sector_network.py
@@ -518,7 +518,7 @@ def add_carrier_buses(n, carrier, nodes=None):
         capital_cost=capital_cost,
     )
 
-    fossils = ["coal", "gas", "oil", "lignite"]
+    fossils = ["coal", "gas", "oil", "lignite", "uranium"]
     if options["fossil_fuels"] and carrier in fossils:
 
         suffix = ""
@@ -1106,17 +1106,20 @@ def prepare_costs(cost_file, params, nyears):
     return costs
 
 
-def add_generation(n, costs):
+def add_generation(n, costs, existing_capacities=0, existing_efficiencies=None, existing_nodes=None):
     logger.info("Adding electricity generation")
 
-    nodes = pop_layout.index
-
-    conventionals = options["conventional_generation"]
+    conventionals = options.get("conventional_generation", {})
 
     for generator, carrier in conventionals.items():
         carrier_nodes = vars(spatial)[carrier].nodes
 
         add_carrier_buses(n, carrier, carrier_nodes)
+
+        if existing_nodes is None:
+            nodes = pop_layout.index
+        else:
+            nodes = existing_nodes[generator]
 
         n.madd(
             "Link",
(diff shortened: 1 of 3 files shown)


martacki:
Except for typos (my specialty), there seems to be only the same bug as in `master`, which is why Windows OS is failing (nut sure about `why`).

Remaining discussion points could be:
- I designed the same setup for nuclear as for fossil generation (Generator `uranium` connected with a `nuclear` link). I think that makes more sense, currently it simply remained a "nuclear generator", but this was a bit confusing
- Currently, if existing capacities are kept and the carrier is expandable, only the existing locations can be expanded. The same carrier can not be expanded at a different site, if that site has no existing capacity. Could be potentially fixed, but wasn't needed in our case


Otherwise, seems to work.
Nuclear / coal generation in power only model:
image

Nuclear/coal generation at sectoral model:
image
image
image

Sectoral model dispatch (in `links_t.p1`) (`Co2L0.5` for whole EU, `121H`, `59` clusters)
nuclear
image
coal
image

scripts/prepare_sector_network.py
```diff
@@ -516,7 +516,7 @@ def add_carrier_buses(n, carrier, nodes=None):
         capital_cost=capital_cost,
     )
 
-    fossils = ["coal", "gas", "oil", "lignite"]
+    fossils = ["coal", "gas", "oil", "lignite", "uranium"]
     if options["fossil_fuels"] and carrier in fossils:
```
tgilon, line 520:
```suggestion
    fossils = ["coal", "gas", "oil", "lignite"]
    if (options["fossil_fuels"] and carrier in fossils) or (carrier == "uranium"):
```
As `uranium` is not a fossil fuel, this would be less misleading.

tgilon: @daniel-rdt

scripts/prepare_sector_network.py
```diff
@@ -1121,14 +1126,61 @@ def add_generation(n, costs):
...
+                else False
+            ),
+            p_nom=(
+                existing_capacities[generator] / existing_efficiencies[generator]
+                if not existing_capacities == 0
+                else 0
+            ),  # NB: existing capacities are MWel
+            p_max_pu=(
+                0.7 if carrier == "uranium" else 1
+            ),  # be conservative for nuclear (maintenance or unplanned shut downs)
```
tgilon, line 1144: For consistency, it would be good to use this value in `add_existing_baseyear`as well.

https://github.com/PyPSA/pypsa-eur/blob/728eacf5589ed2ede90dfe59fd371c66739aa835/scripts/add_existing_baseyear.py#L319-L337

It would be even better to use the country specific values from [data/nuclear_p_max_pu.csv](https://github.com/PyPSA/pypsa-eur/blob/728eacf5589ed2ede90dfe59fd371c66739aa835/data/nuclear_p_max_pu.csv).

@daniel-rdt

scripts/prepare_sector_network.py
```diff
@@ -4538,6 +4590,17 @@ def add_enhanced_geothermal(n, egs_potentials, egs_overlap, costs):
     )
     pop_weighted_energy_totals.update(pop_weighted_heat_totals)
 
+    if options.get("keep_existing_capacities", False):
```
daniel-rdt, line 4593:
```suggestion
    if options.get("keep_existing_capacities", False) and snakemake.params.foresight not in ["myopic", "perfect"]:
```
It would make sense to limit this functionality to overnight, as for myopic and perfect optimization existing caps are added later on

scripts/prepare_sector_network.py
```diff
@@ -4551,7 +4614,7 @@ def add_enhanced_geothermal(n, egs_potentials, egs_overlap, costs):
 
     spatial = define_spatial(pop_layout.index, options)
 
-    if snakemake.params.foresight in ["myopic", "perfect"]:
+    if snakemake.params.foresight in ["overnight", "myopic", "perfect"]:
```
daniel-rdt, line 4617: Are these not all the cases, i.e. we could leave the if condition out completely now?

scripts/prepare_sector_network.py
```diff
@@ -4538,6 +4590,17 @@ def add_enhanced_geothermal(n, egs_potentials, egs_overlap, costs):
...
+    if options.get("keep_existing_capacities", False):
+        existing_capacities, existing_efficiencies, existing_nodes = (
+            get_capacities_from_elec(
+                n,
+                carriers=options.get("conventional_generation").keys(),
+                component="generators",
+            )
+        )
+    else:
+        existing_capacities, existing_efficiencies, existing_nodes = 0, None, None
```
tgilon, line 4602:
```suggestion
        existing_capacities = existing_efficiencies = existing_nodes = None
```

scripts/prepare_sector_network.py
```diff
@@ -1121,14 +1126,61 @@ def add_generation(n, costs):
             * costs.at[generator, "VOM"],  # NB: VOM is per MWel
             capital_cost=costs.at[generator, "efficiency"]
             * costs.at[generator, "fixed"],  # NB: fixed cost is per MWel
-            p_nom_extendable=True,
+            p_nom_extendable=(
+                True
+                if generator
+                in snakemake.params.electricity.get("extendable_carriers", dict()).get(
```
tgilon, line 1132: "The snakemake object should only be accessed from the if __name__ == "__main__" scope, use an options dict similar to what was introduced for example for add_gas_storage_and_grids (including a list of what's looked for in the doc string)."

see https://github.com/open-energy-transition/pypsa-eur/pull/31#discussion_r1995667697

tgilon (commented):
Some more comments based on https://github.com/open-energy-transition/pypsa-eur/pull/31

tgilon:
@martacki What's the status here? Would you like to continue with this, or shall we close it for now?

PR #1372, opened 2024-10-18
Replace bioenergy stores with generators

cpschau:
Closes #1336 
As the Generator component is extended in https://github.com/PyPSA/PyPSA/pull/1047, it can replace the bioenergy Store components. The extension of the Generator allows for a more efficient representation of finite resources, such as biomass, by eliminating the need for storage consistency constraints. Additionally, the desired model behavior to enforce the use of unsustainable bioenergy types is addressed through the new e_sum_min attribute.

## Checklist


scripts/prepare_sector_network.py
```diff
@@ -3100,14 +3088,15 @@ def add_biomass(n, costs):
...
-            )
-            n.stores_t.e_max_pu.loc[n.snapshots[-1], unsus_stores_idx] = 1
+            # Set e_sum_min to 0 to allow for the faux biomass transport
+            n.generators.loc[
+                n.generators.carrier == "unsustainable solid biomass", "e_sum_min"
+            ] = 0
+            # Set e_sum_max to the potential to limit the faux biomass transport
+            n.generators.loc[
+                n.generators.carrier == "unsustainable solid biomass", "e_sum_max"
+            ] = unsustainable_solid_biomass_potentials_spatial.sum()
```
lindnemi, line 3098: Not sure if this is the right constraint. There is one generator for every bus, and if every bus has access to the sum of the total potential wouldn't we get `n_buses * total_potential`  of import capacity?

lindnemi: ah i see, the GlobalConstraint is just below and remains there. Then the questions if, if this e_sum_max constraint is still needed or only a duplicate?

lindnemi:
So far, you changed only the unsustainable stores to generators. Didn't we discuss with @fneum that now all Biomass stores should be changed to generators (which would save the storage constraints, and avoid the possibility that the unsustainable biomass flows through the stores due to numerical inaccuracies)?

scripts/prepare_sector_network.py
cpschau: I think you are right that it is not only redundant, but actually false. The e_sum_max set at the original generator intialization in https://github.com/PyPSA/pypsa-eur/blob/42c262eeef2380f48c550000b4d2dc949db5137f/scripts/prepare_sector_network.py#L2913 should remain. All the generation needed beyond that local limit must be provided by the other "artifical" generator, that also takes into account the transport costs.

cpschau:
> So far, you changed only the unsustainable stores to generators. Didn't we discuss with @fneum that now all Biomass stores should be changed to generators (which would save the storage constraints, and avoid the possibility that the unsustainable biomass flows through the stores due to numerical inaccuracies)?

You are right! This was just a quick draft to resolve the solvability issues for ariadne. All other stores will be replaced in https://github.com/PyPSA/pypsa-eur/pull/1373.

cpschau:
Closed due to redundancy with https://github.com/PyPSA/pypsa-eur/pull/1373

PR #1373, opened 2024-10-18
Replace bioenergy and MSW stores with generators

cpschau:
Closes # (if applicable).

As the Generator component is extended in https://github.com/PyPSA/PyPSA/pull/1047, it can replace the Store components for bioenergy and municipal solid waste. The extension of the Generator allows for a more efficient representation of finite resources, such as biomass, by eliminating the need for storage consistency constraints. Additionally, the desired model behavior to enforce the use of unsustainable bioenergy types and municipal solid waste is addressed through the new e_sum_min attribute.

bioenergy_balance

## Checklist


fneum (changes requested):
Proposed changes make sense, but some adjustments are necessary:
- The new minimum version of PyPSA should be specified in the `environment.yaml` since this PR uses a new feature.
- A release note should be added, especially since there are some hidden changes (i.e. forced use of MSW, changed carrier name for biomass transport)
- Do you actually still need the operational limit GlobalConstraint for MSW? If you want all MSW to be used, you could also just set `e_sum_min = e_sum_max`, right?

cpschau:
> Proposed changes make sense, but some adjustments are necessary:
> 
>     * The new minimum version of PyPSA should be specified in the `environment.yaml` since this PR uses a new feature.
> 
>     * A release note should be added, especially since there are some hidden changes (i.e. forced use of MSW, changed carrier name for biomass transport)
> 
>     * Do you actually still need the operational limit GlobalConstraint for MSW? If you want all MSW to be used, you could also just set `e_sum_min = e_sum_max`, right?

Thanks for the review @fneum!
1. Isn't the requirement `pypsa>=0.31` already specified in the `environment.yaml` anyway?
2. Done
3. I think you need it for setups with `biomass_spatial:true` and `biomass_transport:false` when the MSW transport is mimicked using the additional Generator component. In that case the usage of MSW  at a bus can be less than the bus' MSW potential (`p_sum_min <= p_sum_max`), as long as the difference is transported to and used at another bus using that same additional Generator as compensation. Therefore, the GlobalConstraint is still required to ensure the usage of the aggregate MSW potential.

fneum (approved):
Thanks for the explanation! If 0.31 is already specified, it's all good!

PR #1376, opened 2024-10-21
fix MtK VOM

p-glaum:
in the master the marginal_cost are currently in $€/MWh_{kerosene}$ but they should be in $€/MWh_{MeOH}$ because bus 0 is a methanol bus

## Checklist


fneum (approved):

fneum:
Good catch!

PR #1384, opened 2024-10-22
Major improvement to OSM-based electricity grid (e.g. using relations, preserving substation locations)

bobbyxng:
- Major improvement to the OSM-based electricity grid
- `build_osm_network` completely rewritten:
  - Use of OSM relations, where available and where a single linestring could be extracted to replace member ways
  - substantial speed improvements
  - Exact locations of original substations are preserved
  - Lines with identical geometries and voltages are aggregated (num_parallel is increased)
  - Lines overpassing "virtual nodes" (not actual substations as per OSM raw data) are merged, if their electric properties allow it (same voltage levels, same number of circuits)
  - Geometries are preserved. Method to extend line/link to the substation has been improved significantly. No "mess" inside of a substation cluster anymore.
  - Method to determine substation point improved, based on "Pole of Inaccessibility" (PoI) method developed by Garcia-Castellanos & Lombardo 2007, doi.org/10.1080/14702540801897809 which ensures that an interior point in non-convex shapes is always found.
  - Clustering resolution has been sharpened from 5000 meter radius to only 500 meters, while still ensuring topological connectivity.
  - Single bus for each voltage level in each substation. Buses of different voltages are arranged in geometric shape (three buses form a triangle, see screenshot)
  - Transformers now have a capacity s_nom based on the maximum of connected capacity connected to each bus
  - Resolves many issues related around cross-border capacity, especially if the substation was located close to a border. @lindnemi @toniseibold @JulianGeis @nworbmot  

image

image

image

## Checklist

- Done here: https://github.com/PyPSA/pypsa-eur/pull/1387


scripts/base_network.py
```diff
@@ -456,7 +456,8 @@ def _set_electrical_parameters_transformers(transformers, config):
 
     ## Add transformer parameters
     transformers["x"] = config.get("x", 0.1)
-    transformers["s_nom"] = config.get("s_nom", 2000)
+    if "s_nom" not in transformers:
+        transformers["s_nom"] = config.get("s_nom", 2000)
```
fneum, line 460: What happens if some transformers have `s_nom`, but not all of them. Will missing values be filled?

scripts/build_gdp_pop_non_nuts3.py
```diff
@@ -123,7 +123,9 @@ def calc_gdp_pop(country, regions, gdp_non_nuts3, pop_non_nuts3):
     if "snakemake" not in globals():
         from _helpers import mock_snakemake
 
-        snakemake = mock_snakemake("build_gdp_pop_non_nuts3")
+        snakemake = mock_snakemake(
+            "build_gdp_pop_non_nuts3", configfiles=["config/config.osm-raw.yaml"]
+        )
```
fneum, line 128: revert?

config/config.default.yaml
```diff
@@ -84,7 +84,7 @@ co2_budget:
 
 # docs in https://pypsa-eur.readthedocs.io/en/latest/configuration.html#electricity
 electricity:
-  voltages: [200., 220., 300., 380., 500., 750.]
+  voltages: [220., 300., 330., 380., 400., 500., 750.]
```
fneum, line 87: Just a short note why 200 kV disappears for clarification would be good.

scripts/build_osm_network.py
```diff
@@ -41,823 +57,1576 @@
...
+            "voltage", None
+        )  # Default to None if not present
+
+        # Extract the geometries for all nodes in the subgraph
+        geometry = [G.nodes[node].get("geometry", None) for node in subgraph.nodes()]
+        geometry = linemerge(geometry)  # Merge the geometries
+
+        # # list of all countries
+        # country = ';'.join([G.nodes[node].get('country', '') for node in subgraph.nodes()])
+        # country = ';'.join(sorted(set(country.split(';'))))
```
fneum, line 427: could remove unused code?

scripts/build_osm_network.py
```diff
@@ -41,823 +57,1576 @@
...
+    all_transformers = gpd.GeoDataFrame(
+        columns=TRANSFORMERS_COLUMNS + ["transformer_id"], crs=geo_crs
+    )
 
-    for p in points:
-        # execute split to all lines and store results
-        temp_list = [split(l, p) for l in list_linestrings]
-        # nest all geometries
-        list_linestrings = [lstring for tval in temp_list for lstring in tval.geoms]
+    for g_name, g_value in buses_all.groupby("station_id"):
```
fneum, line 1125: *If this is slow,* could be converted to an apply function applied to each group.

fneum (commented):
I didn't check every line in detail, but the structure looks good and comparable to the previous version. I have a series of small comments that are easy to address:

config/config.default.yaml
bobbyxng: Yes, I double-checked the voltage levels being used in Europe. Both JAO Static Grid Model and the ENTSO-E map show that that 200 kV is not a standardised value in operation. I then checked the few lines in OSM that were labeled as 200 kV lines, of which some were actually 20 kV lines (with a typo, trailing zero). Now the underlying OSM input data (at least for Europe) does not contain 200 kV lines anymore.

- copied from https://github.com/PyPSA/pypsa-eur/pull/1387#discussion_r1830634924

scripts/build_osm_network.py
bobbyxng: done

scripts/build_osm_network.py
bobbyxng: it is surprisingly fast, few seconds, so no need for now from my side :)

scripts/build_gdp_pop_non_nuts3.py
bobbyxng: done

bobbyxng:
Docstrings added

PR #1387, opened 2024-10-23
Updated to osm-network version 0.5

bobbyxng:
Update osm base network from 0.4 to 0.5:
- Exact locations of original substations and converter stations (interior point/Pole of Inaccessibility) are preserved
- Clustering resolution improved from 5000 to 500 meters
- Lines of same electric parameters are merged, if they cross a virtual bus (that is not a real substation)
- Information from OSM relations are used, wherever applicable. To avoid doubling, members (ways) of the relation are dropped in the set of lines, accordingly
- There are now unique transformers for each voltage level in each station. Transformers now have a nominal capacity, representing the maximum of line capacities connected to either side/bus of the transformer (n-0, nominal capacity)
- Wherever applicable, OSM IDs are preserved and used in the index of the network components.

## Validation:
- Europe, at NUTS2 resolution (318 regions), electricity sector, clustered at 3-hourly resolution
- osm-prebuilt v0.4 (left) and v0.5 (right)
Screenshot from 2024-10-23 11-12-19

## Checklist


lkstrp (approved):

doc/configtables/electricity.csv
```diff
@@ -1,7 +1,7 @@
 ,Unit,Values,Description
-voltages,kV,"Any subset of {200., 220., 300., 380., 500., 750.}",Voltage levels to consider
+voltages,kV,"Any subset of {220., 300., 330., 380., 400., 500., 750.}",Voltage levels to consider
 base_network, --, "Any value in {'entsoegridkit', 'osm-prebuilt', 'osm-raw}", "Specify the underlying base network, i.e. GridKit (based on ENTSO-E web map extract, OpenStreetMap (OSM) prebuilt or raw (built from raw OSM data), takes longer."
-osm-prebuilt-version, --, "float, any value in range 0.1-0.4", "Choose the version of the prebuilt OSM network. Defaults to latest Zenodo release."
+osm-prebuilt-version, --, "float, any value in range 0.1-0.5", "Choose the version of the prebuilt OSM network. Defaults to latest Zenodo release."
```
lindnemi, line 4: It looks like you're changing the zenodo link in this PR to point specifically to 0.5. Since it says here, that it defaults to the "latest release" we could also use a static link that always point to the latest version, right? Not sure if that is possible, but I thought zenodo offered these links that automatically point to the latest version of a dataset.

config/config.default.yaml
```diff
@@ -84,7 +84,7 @@ co2_budget:
 
 # docs in https://pypsa-eur.readthedocs.io/en/latest/configuration.html#electricity
 electricity:
-  voltages: [200., 220., 300., 380., 500., 750.]
+  voltages: [220., 300., 330., 380., 400., 500., 750.]
   base_network: osm-prebuilt
   osm-prebuilt-version: 0.4
```
lindnemi, line 89: why is the osm-prebuilt-version still at 0.4?

bobbyxng: updated now :)

doc/configtables/electricity.csv
bobbyxng: Zenodo does indeed provide a doi link that then redirects to the latest version. 
The reason I implemented it like this is for the user to have more control, i.e. if a previous study was done on an older version, I can keep using X.y.

config/config.default.yaml
```diff
@@ -84,9 +84,9 @@ co2_budget:
 
 # docs in https://pypsa-eur.readthedocs.io/en/latest/configuration.html#electricity
 electricity:
-  voltages: [200., 220., 300., 380., 500., 750.]
+  voltages: [220., 300., 330., 380., 400., 500., 750.]
```
fneum, line 87: Can you provide some background why 200 kV was removed?

fneum (approved):
LGTM, just one clarification required.

config/config.default.yaml
bobbyxng: Yes, I double-checked the voltage levels being used in Europe. Both JAO Static Grid Model and the ENTSO-E map show that that 200 kV is not a standardised value in operation. I then checked the few lines in OSM that were labeled as 200 kV lines, of which some were actually 20 kV lines (with a typo, trailing zero). Now the underlying OSM input data (at least for Europe) does not contain 200 kV lines anymore.

fneum: Thanks for the explanation!

PR #1398, opened 2024-10-28
Bugfix: unsustainable bioliquids renaming

bobbyxng:
Closes # (if applicable).

- Bugfix in `prepare_sector_network.py`
- Renaming of `bioliquids` to `unsustainable bioliquids`, as a mismatch in indexing occurs otherwise

## Checklist

### DIFF
--- a/scripts/prepare_sector_network.py
+++ b/scripts/prepare_sector_network.py
@@ -2541,7 +2541,7 @@ def add_biomass(n, costs):
     if options["regional_oil_demand"]:
         unsustainable_liquid_biofuel_potentials_spatial = biomass_potentials[
             "unsustainable bioliquids"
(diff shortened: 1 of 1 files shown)


cpschau (approved):
Looks good! The same change is also included in #1373

lindnemi:
Great so we fixed it three times
https://github.com/PyPSA/pypsa-eur/pull/1357

lindnemi:
Maybe we think of a way to get simple fixes like this faster into master when @fneum is not available

lkstrp:
> Great so we fixed it three times #1357

Three weeks even

> Maybe Probably we should think of a way to get simple fixes like this faster into master when @fneum is not available

Agreed

PR #1399, opened 2024-10-30
update co2_budget

lindnemi:
I would like to suggest an update of the default yearly co2_budget in PyPSA-Eur. The current values seem to ambitious compared to the EU goals. I don't know where the current values come form, maybe they already include some readjustment for the sectors PyPSA contains, if that's the case i would like to learn more about how they are computed.

For 2020 i suggest to use the average historical emissions for 2019-2020.
For 2025 i suggest to use the projected member states emissions in the optimist "with additional measures" scenario of the EEA
For 2030, 2040, 2050 use the EU targets.
For 2035, 2045 some ad hoc interpolation of the targets.

## Checklist


fneum (changes requested):
I don't have too strong feelings about the default values here, but we should maybe not mix snapshot targets for specific years with budgets. Also, it might be too simplified to expand EU targets to many non-EU countries covered by the model.

While I don't exactly know where the current defaults for the co2_budget come from, I believe they were calculated according to a methodology described by Marta Victoria et al. in the supplementary material of https://www.nature.com/articles/s41467-020-20015-4. These are Europe-attributed total budgets for specific temperature reduction targets (e.g. 1.5°C). 

https://static-content.springer.com/esm/art%3A10.1038%2Fs41467-020-20015-4/MediaObjects/41467_2020_20015_MOESM1_ESM.pdf

On a technical level, the defaults should be documented in the corresponding `.csv` file for the config segment rather than directly in the YAML.

koen-vg:
On one hand I would say that this is a setting which has to be carefully considered by anyone that wants to do "serious" work with pypsa-eur, so the default value doesn't matter all that much. But on the other hand I do agree that if anything, the suggestions in this PR are much more likely to be a reasonable default for most use-cases. To me it seems like there is a good chance that the majority of all studies using pypsa-eur would want to stick to 55%/90%/100% reduction targets for 2030/40/50. So I would merge this. I think it's always better to have more reasonable defaults.

lisazeyen:
I think the targets where coming from the +1.7 budget from Martas study. I would also agree to update to current political targets. We should just add this to the release notes so that people are not surprised by different results

martavp:
Yes, I agree that it makes sense to include the current political targets by default. 
We should just make sure to mention in the release notes that the path does no longer entail using a carbon budget that corresponds to a 1.7C temperature increase (with all the assumptions described in the paper linked above).

koen-vg:
> Also, it might be too simplified to expand EU targets to many non-EU countries covered by the model.

I will just briefly note that in principle it should be possible to compile at least all non-EU emissions reduction targets for 2030 and calculate some kind of weighted average to get an even more precise estimate than 55% reduction. But I know that e.g. Norway has the same target for 2030, UK has a more ambitious target, whereas some Balkan countries have less ambitious targets. All in all I would expect the target for the whole modelling region to be very close to 55%.

For 2040, who knows. Even the EU target is not law yet, and most individual countries don't have targets for 2040. But 90% already seems ambitious, so the current default of ~93% reduction by 2040 seems less realistic to me.

cristobal-GC:
When documenting the new reduction targets, it would be nice to mention that, for analyises not covering all the sectors, the default values taken from policy targets should be modified. For example, for only-electricity analyses, the co2 limit coherent with -55% for 2030 should be more ambitious.

koen-vg:
That's a very good point.

Now that we are talking CO2 limits, I was reminded of something else, which I just wanted to put on the radar. When the EU talks about 55%, 90%, 100% reductions in emissions, they include not only the emissions that pypsa-eur (the sector-coupled version) accounts for, but also non-CO2 emissions from the agriculture sector on one hand and negative emissions from the "land use, land use change and forestry" (LULUCF) sector. Together, these two contribute with net positive CO2-equivalent emissions today, but are expected or hoped to contribute with net negative emissions by 2050.

I added some really simple functionality to pypsa-eur to add those emissions (positive or negative) here: https://github.com/koen-vg/pypsa-eur/commit/8349f10cdfc84709e86aac0f71ab7b57a1b095ec. This works fine, but the main problem is that I only got numbers for the EU so far. If we were to do this properly, it would be great to get data on a country level (also including non-EU countries) and automatically adapt to the selected countries. Unfortunately I don't have time for that right now. But just wanted to put it out there: if anyone else has time to implement this properly, that would be quite useful!

In the end, we are not talking _huge_ numbers; currently about 150Mt/a positive emissions, and by 2050 maybe about -100Mt/a negative emissions, though with a pretty large uncertainty, depending on the future of agriculture etc. Still, it's really not insignificant.

fneum:
You all convinced me! I am merging this with a release note, noting that the budget might be refined in the future to include non-EU targets.

PR #1403, opened 2024-10-31
update shared_resources "base" to not retrigger renewable profiles

p-glaum:
Makes that with `shared_resources=base`, the previous behavior is recovered and every rule starting from `add_electricity` is not shared anymore, but the renewable profiles are shared.  

## Checklist


scripts/_helpers.py
```diff
@@ -107,11 +107,26 @@ def get_run_path(fn, dir, rdir, shared_resources, exclude_from_shared):
...
+            "busmap_base_s_{clusters}.csv",
+            "linemap_base_s_{clusters}.csv",
+            "cluster_network_base_s_{clusters}",
+            "profile_{clusters}_",
+            "build_renewable_profile_{clusters}",
+            "availability_matrix_",
+            "determine_availability_matrix_",
+            "solar_thermal",
+        )
+        if any(prefix in fn for prefix in shared_files) or is_shared:
```
lindnemi, line 128: why is the `or is_shared` necessary? Looks like it could be omitted

p-glaum: In my opinion, this is necessary here.
Because for the first `is_shared`, in the previous behaviour, you had neither a relevant_wildcard nor a shared_rule to make it shared. However, with the new structure, this would always be False for the profiles, meaning that they are never shared in the end.
This is why I added `shared_files`, and if any of those is in the file name, then the resource is always shared.

fneum (commented):
I would like to test this. Do you have a configfile/s to run to validate the intended behaviour?

p-glaum:
I just had two different scenarios "A", "B" like
```
run:
  prefix: 'test_shared_resources'
  name: 
    - A
    - B
  scenarios:
    enable: true
    file: config/scenarios.yaml
  disable_progressbar: true
  shared_resources:
    policy: base
    exclude: []
  shared_cutouts: true
  ```
  
  with the scenario.yaml
  ```
A:
  electricity:
    renewable_carriers: [solar, onwind]

B:
  electricity:
    renewable_carriers: [solar, offwind-dc]
```

scripts/_helpers.py
fneum: `is_shared` is required.

fneum (approved):

PR #1407, opened 2024-11-04
bugfix: only add waste chp to buses which have urban central heat buses

p-glaum:
Closes # (if applicable).

## Checklist


fneum (changes requested):
Not sure, but could make sense to add WtE everywhere but keep bus2 empty when urban central heat is not available.

Good to go once tested it doesn't cause any infeasibilities.

fneum (approved):
Yes, I think this is better!

PR #1410, opened 2024-11-06
Correct bioliquid emissions by BtL efficiency

cpschau:
In `prepare_sector_network` the link between unsustainable bioliquids and oil previously did not correctly account for the amount of CO₂ emissions withdrawn from the atmosphere. The efficiency value was directly copied from the BtL (biomass to liquid) link without adjusting for the primary efficiency value, which represents the conversion between solid biomass and oil.

With this updated efficiency, 0.248 MtCO₂ are withdrawn from the atmosphere per MWh of unsustainable bioliquids introduced into the system. However, there remains a discrepancy when compared to the CO₂ intensity of oil used in the model, which stands at 0.2571. This difference is consistent with the BtL link’s implementation. An alternative approach would be to adjust the efficiency value to match the exact CO₂ intensity of oil, thereby eliminating the gap.

## Checklist

### DIFF
--- a/scripts/prepare_sector_network.py
+++ b/scripts/prepare_sector_network.py
@@ -2722,11 +2722,7 @@ def add_biomass(n, costs):
             bus2="co2 atmosphere",
             carrier="unsustainable bioliquids",
             efficiency=1,
-            efficiency2=(
-                -costs.at["solid biomass", "CO2 intensity"]
-                + costs.at["BtL", "CO2 stored"]
-            )
-            / costs.at["BtL", "efficiency"],
+            efficiency2=-costs.at["oil", "CO2 intensity"],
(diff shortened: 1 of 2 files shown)


scripts/prepare_sector_network.py
```diff
@@ -2722,8 +2722,11 @@ def add_biomass(n, costs):
             bus2="co2 atmosphere",
             carrier="unsustainable bioliquids",
             efficiency=1,
-            efficiency2=-costs.at["solid biomass", "CO2 intensity"]
-            + costs.at["BtL", "CO2 stored"],
+            efficiency2=(
```
lindnemi, line 2725: in would prefer this to be `-costs.at["oil", "CO2 intensity"]`

cpschau: Done

PR #1423, opened 2024-11-14
Bugfix: limit number of vehicles participating in V2G by dsm rate

koen-vg:
Previously, all vehicles were assumed to be V2G-capable if V2G was turned on. But BEV battery capacity was still limited by the `bev_availability` option (0.5 by default). The result was BEV batteries effectively having a storage/discharge ratio of ~2.3 instead of ~4.5 as should be the case with 50kWh batteries and 11kW chargers.

Moreover, V2G is now only enabled if demand-side-management (DSM) is also enabled; V2G had no effect without DSM anyway unless I've misunderstood badly: V2G needs the battery store that's added by the DSM option in order to be useful at all.

Finally, rename the `bev_availability` option to `bev_dsm_availability` to avoid confusion with time-dependent BEV charge/discharge availability profiles, which are also referred to as `bev_availability` elsewhere.

## Checklist

### DIFF
--- a/config/config.default.yaml
+++ b/config/config.default.yaml
@@ -499,7 +499,7 @@ sector:
   EV_lower_degree_factor: 0.98
   EV_upper_degree_factor: 0.63
   bev_dsm: true
-  bev_availability: 0.5
+  bev_dsm_availability: 0.5
   bev_energy: 0.05
   bev_charge_efficiency: 0.9
   bev_charge_rate: 0.011
--- a/doc/configtables/sector.csv
+++ b/doc/configtables/sector.csv
@@ -40,13 +40,13 @@ EV_lower_degree_factor,--,float,Share increase in energy demand in electric vehi
 EV_upper_degree_factor,--,float,Share increase in energy demand in electric vehicles (EV) for each degree difference between the hot environment and the maximum temperature.
 bev_dsm,--,"{true, false}",Add the option for battery electric vehicles (BEV) to participate in demand-side management (DSM)
 ,,,
-bev_availability,--,float,The share for battery electric vehicles (BEV) that are able to do demand side management (DSM)
+bev_dsm_availability,--,float,The share for battery electric vehicles (BEV) that are able to do demand side management (DSM)
 bev_energy,--,float,The average size of battery electric vehicles (BEV) in MWh
 bev_charge_efficiency,--,float,Battery electric vehicles (BEV) charge and discharge efficiency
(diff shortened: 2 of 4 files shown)


koen-vg:
Also while we're on the topic of V2G / DSM, I want to pick on another issue. Right now, vehicles that participate in DSM must be charged to a certain level (default: 75%) at a certain hour (default: 7:00) every day; at any other hour their batteries can have any state of charge at all, all the way down to 0%.

To me, that feels quite unrealistic. Without having looked very carefully at the results, I image that the model will often drain EV batteries fully during the day and charge them during the night to meet the 75% target at 7:00.

It seems much more reasonable to assume that people with EVs aren't going to allow their vehicles to go down to 0% battery on a regular basis. Instead, something like a minimum state-of-charge of 50-75% _throughout the day_ seems reasonable (see e.g. https://www.sciencedirect.com/science/article/pii/S0301421523002343#sec3). In effect, it feels to me like we are over-estimating effective grid-connected battery capacity by a factor of 2-4.

Now, the `bev_availability` (in this PR renamed to `bev_dsm_availability`) parameter could be seen as a proxy to controlling this. One could set it to 0.25 in order to effectively guarantee a 75% minimum state of charge across the _entire_ electric vehicle stock, including those not participating in V2G. But that seems like a bit of a hack to me.

If there is any interest towards this, I would propose adding a config option like `bev_dsm_restriction_all_hours` with a value of 60% by default (open to suggestions for a better name) and use it to amend the profile calculated in `bev_dsm_profile` in `build_transport_demand.py`.

fneum:
Agreed, the BEV DSM is kind of a weak spot in the model. One potentially has to worry about infeasibilities if the lower SOC limit is set too high.

fneum (approved):
Changes look good! I agree to the slightly breaking config change. Thanks @koen-vg.

@lkstrp, feel free to merge.

koen-vg:
Feel free to merge, any DSM stuff can go in a different issue or PR :)

> Agreed, the BEV DSM is kind of a weak spot in the model. One potentially has to worry about infeasibilities if the lower SOC limit is set too high.

I'm actually not sure if infeasibilities would be an issue, at least with the idea I was thinking of. Maybe I didn't communicate it very well. I'm saying we should consider maintaining a minimum SOC of ~60-70% in EV batteries at all times, possibly with an even higher state of charge enforced at certain hours. This would be equivalent to reducing the EV battery capacity to 30-40% of the original, as only the "top" 30-40% of the battery is available for V2G operations. Just reducing the size of EV batteries in the model cannot lead to infeasibilities unless I'm missing something.

PR #1431, opened 2024-11-19
Update to OSM-prebuilt v0.6

bobbyxng:
Closes # (if applicable).

- Added interactive map to the Zenodo databundle: https://zenodo.org/records/14144752
- Can be directly accessed through: https://pypsa-eur.readthedocs.io/en/latest/data-base-network.html Thanks @lkstrp 
- Prebuilt network includes all technical parameters (impedance, nominal current, reactance, susceptance, etc.) for potential use outside PyPSA/PyPSA-Eur
- Scripts `build_osm_network` and `prepare_osm_network_release` reflect this change and include the construction of the interactive map. Base network and subsequent steps are not impacted by this. 
- I have tested the model at hourly resolution for the target year, model converges, results are similar to the gridkit-based network.

## Checklist


FabianHofmann:
@bobbyxng I have problems with the retrieval of map.html. they are not included in the zenodo repo. Is that perhaps something for future zenodo versions?

bobbyxng:
> @bobbyxng I have problems with the retrieval of map.html. they are not included in the zenodo repo. Is that perhaps something for future zenodo versions?

Hi @FabianHofmann ! I just did a test using the latest pypsa-eur master branch, for me `map.html`is downloaded just like all other files. Note the file only exists from version 0.6 onwards. Download takes a bit longer (in my case around 30 seconds), as it's quite large at >40 MB

FabianHofmann:
Gotcha! thanks, I still had the old version!

FabianHofmann:
on the other hand this means that the osm version setting is obsolete. At least setting version <0.6, which the current code suggests to be working, breaks. I would suggest to either only support version tags >= 0.6 or make the code flexible to support also older versions

PR #144, opened 2020-03-17
no defaults channel in environment.yaml

fneum:
Fixes #142 
Let's see...

PR #1444, opened 2024-12-05
Transition from stores to storage units for LTES, introducing energy-to-power ratio

TomKae00:
Closes # (if applicable).

This pull request introduces the following adjustments to the prepare_sector_network script:

- Added PTES.
- Transitioned stores to storage units for ltes, enabling the implementation of an energy-to-power ratio for these components.

## Effect of Changes: Results Analysis

1. Storage Sizing: stores vs. storage units

The transition from stores to storage units results in larger storage sizing:

energy_supply_comparison

2. Charging and Discharging Dynamics

The implementation of the energy-to-power ratio limits the charging and discharging rates relative to storage size. This is illustrated in the following plots:

Urban Central Water Tanks:

charging_discharging_urban_central_water_tanks

Urban Decentral Water Tanks:

charging_discharging_urban_decentral_water_tanks

3. Energy Balances

The changes in charging and discharging dynamics influence the energy balance, as shown for urban central and urban decentral systems in the plots below:

Urban Central:

energy_balance_comparison_urban_central_heat

Urban Decentral:

energy_balance_comparison_urban_decentral_heat

4. Impact on Total System Costs

The shift from stores to storage units impacts the total system costs slightly, resulting in a small decrease in costs for the storage units compared to stores:

<div style="text-align: center;"> total_cost_comparison </div>

## Checklist

- [✓] I tested my contribution locally and it works as intended.
- [✓] Code and workflow changes are sufficiently documented.
- [✓] Changed dependencies are added to `envs/environment.yaml`.
- [✓] Changes in configuration options are added in `config/config.default.yaml`.
- [✓] Changes in configuration options are documented in `doc/configtables/*.csv`.
- [✓] Sources of newly added data are documented in `doc/data_sources.rst`.
- [✓] A release note `doc/release_notes.rst` is added.

### DIFF
--- a/config/config.default.yaml
+++ b/config/config.default.yaml
@@ -1215,6 +1215,8 @@ plotting:
     residential urban decentral water tanks discharger: '#baac9e'
     services rural water tanks discharger: '#bbc2b8'
     services urban decentral water tanks discharger: '#bdd8d3'
+    water pits: '#cc826a'
+    urban central hot water pits: '#d96f4c'
     # heat demand
     Heat load: '#cc1f1f'
     heat: '#cc1f1f'
--- a/scripts/prepare_sector_network.py
+++ b/scripts/prepare_sector_network.py
@@ -2271,14 +2271,6 @@ def add_heat(
                 heat_system.central_or_decentral
             ]
 
-#            n.add(
-#                "Bus",
-#                nodes + f" {heat_system} water tanks",
-#                location=nodes,
-#                carrier=f"{heat_system} water tanks",
-#                unit="MWh_th",
-#            )
-
             n.add(
                 "StorageUnit",
                 nodes + f" {heat_system} water tanks",
@@ -2299,54 +2291,9 @@ def add_heat(
                 e_cyclic=True,
             )
 
-#            n.add(
-#                "Link",
-#                nodes + f" {heat_system} water tanks charger",
-#                bus0=nodes + f" {heat_system} heat",
-#                bus1=nodes + f" {heat_system} water tanks",
-#                efficiency=costs.at["water tank charger", "efficiency"],
-#                carrier=f"{heat_system} water tanks charger",
-#                p_nom_extendable=True,
-#            )
-
-#            n.add(
-#                "Link",
-#                nodes + f" {heat_system} water tanks discharger",
-#                bus0=nodes + f" {heat_system} water tanks",
-#                bus1=nodes + f" {heat_system} heat",
-#                carrier=f"{heat_system} water tanks discharger",
-#                efficiency=costs.at["water tank discharger", "efficiency"],
-#                p_nom_extendable=True,
-#            )
-
-
-#            n.add(
-#                "Store",
-#                nodes + f" {heat_system} water tanks",
-#                bus=nodes + f" {heat_system} water tanks",
-#                e_cyclic=True,
-#                e_nom_extendable=True,
-#                carrier=f"{heat_system} water tanks",
-#                standing_loss=1 - np.exp(-1 / 24 / tes_time_constant_days),
-#                capital_cost=costs.at[
-#                    heat_system.central_or_decentral + " water tank storage", "fixed"
-#                ],
-#                lifetime=costs.at[
-#                    heat_system.central_or_decentral + " water tank storage", "lifetime"
-#                ],
-#            )
-
             if heat_system == HeatSystem.URBAN_CENTRAL:
 
                 n.add("Carrier", f"{heat_system} water pits")
-
-#                n.add(
-#                    "Bus",
-#                    nodes + f" {heat_system} water pits",
-#                    location=nodes,
-#                    carrier=f"{heat_system} water pits",
-#                    unit="MWh_th",
-#                )
 
                 n.add(
                     "StorageUnit",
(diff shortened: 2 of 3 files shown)


config/config.default.yaml
```diff
@@ -1215,6 +1215,8 @@ plotting:
     residential urban decentral water tanks discharger: '#baac9e'
     services rural water tanks discharger: '#bbc2b8'
     services urban decentral water tanks discharger: '#bdd8d3'
+    water pits: '#cc826a'
```
cpschau, line 1218: pit storage should be only added within district heating systems, i.e. as urban central hot water pit

scripts/prepare_sector_network.py
```diff
@@ -2264,56 +2264,57 @@ def add_heat(
...
                 standing_loss=1 - np.exp(-1 / 24 / tes_time_constant_days),
                 capital_cost=costs.at[
                     heat_system.central_or_decentral + " water tank storage", "fixed"
                 ],
                 lifetime=costs.at[
                     heat_system.central_or_decentral + " water tank storage", "lifetime"
                 ],
+                e_nom_extendable=True,
+                e_cyclic=True,
             )
```
cpschau, line 2294: We might need another case distinction for different sizes of water tanks (DEA's "large hot water tank" for urban central heat and "small scale hot water tank" for urban decentral and rural heat), as the DEA provides different energy-to-power-ratio and costs. This would require additional changes in technology-data.

cpschau (changes requested):
Hi @TomKae00,
the implementation looks fine to me. However, the data behind it seems to contain some wrong assumptions that probably sneaked in with the recent changes in the technology-data repository, namely [#144](https://github.com/PyPSA/technology-data/pull/144) and [#152](https://github.com/PyPSA/technology-data/pull/152). In my opinion, we should assume (dis-)charger efficiencies of 1, as they are stated in the DEA catalogue. The indicated round-trip-efficiency-losses are probably chosen to give a proxy for the average standing losses. The standing losses, however, are already represented in our model by the configurable parameter `tes_tau`. Therefore, we would withdraw them twice according to the current implementation. These losses are probably also the reason, why PTES expansion decreases so heavily in your comparison between former "Store" and "StorageUnit" implementation and the total system costs change so much. To fix this issue, I would suggest to 
1. submit a new PR in technology-data that refers the right charger- and discharger efficiencies for tanks and pits,
2. ask @lkstrp to bump a new technology-data release, as soon as it is merged,
3. and finally reference the newly release version in the `config.default.yaml`.

As mentioned in the in-line review, it could also make sense to add the small scale hot water tanks to the technology-data to serve as decentral heat storage option, unless there are arguments against it @amos-schledorn

scripts/prepare_sector_network.py
```diff
@@ -2264,56 +2264,57 @@ def add_heat(
...
-                "Store",
+                "StorageUnit",
                 nodes + f" {heat_system} water tanks",
-                bus=nodes + f" {heat_system} water tanks",
-                e_cyclic=True,
-                e_nom_extendable=True,
+                bus=nodes + f" {heat_system} heat",
                 carrier=f"{heat_system} water tanks",
+                efficiency_store=costs.at["water tank charger", "efficiency"],
+                max_hours=costs.at[
```
cpschau, line 2283: TTES in district heating and decentral water tanks do not have the same EtP ratio.

scripts/prepare_sector_network.py
```diff
@@ -2264,56 +2264,66 @@ def add_heat(
...
+                    heat_system.central_or_decentral + " water tank storage",
+                    "energy to power ratio",
+                ],
+                efficiency_dispatch=costs.at[
+                    heat_system.central_or_decentral + " water tank discharger",
+                    "efficiency",
+                ],
+                p_nom_extendable=True,
                 standing_loss=1 - np.exp(-1 / 24 / tes_time_constant_days),
                 capital_cost=costs.at[
```
cpschau, line 2293: Capital costs are related to power not to energy capacity --> Conversion from €/MWh to €/MW using energy to power ratio.

scripts/prepare_sector_network.py
```diff
@@ -2264,56 +2264,66 @@ def add_heat(
...
+                    ],
+                    max_hours=costs.at[
+                        "central water pit storage", "energy to power ratio"
+                    ],
+                    efficiency_dispatch=costs.at[
+                        "central water pit discharger", "efficiency"
+                    ],
+                    p_nom_extendable=True,
+                    standing_loss=1 - np.exp(-1 / 24 / tes_time_constant_days),
+                    capital_cost=costs.at["central water pit storage", "fixed"],
```
cpschau, line 2322: Same here. Capital costs need to be converted according to unit.

cpschau (commented):
I noticed one more thing that needs to be adjusted.

scripts/prepare_sector_network.py
fneum: Yes, I was thinking that as well. Assuming a single E2P ratio might be very restrictive.

fneum:
The argument for this change is that the E2P ratio cannot be freely optimized, implying that extremely high/low ratios would not be technically feasible, right?

This should make the total system costs slightly higher if all unit/efficiency/cost calculations are done correctly. I was first a bit surprised by the lower cost with StorageUnit solution.

As @cpschau pointed out, giving the model only one E2P option might be very restrictive. One could add multiple (e.g. three E2P options as separate StorageUnits). An alternative -- which I am not sure I would recommend -- would be to keep Store/Link combination and allow ranges of E2P ratios.

amos-schledorn:
@TomKae00 I've just made a new `technology-data` release, so you can bump up the version in `config.default.yaml` to `v0.10.1`. Then, I'll rerun the CI.

TomKae00:
@amos-schledorn
- Checked the charging and discharging efficiencies of the three storage types (PTES, large water tanks, and small water tanks) against the current DEA Technology Catalogue for energy storage. The dis-/charging efficiency for all three types is listed as 1.0.
- Validated the outputs for efficiency_store and efficiency_dispatch during the model run, and the dis-/charging efficiencies are applied correctly for all three storage types.
- Thanks for the updated technology-data release! The config.default.yaml now points to v0.10.1.

@fneum
Yes, that's correct! Using the max_hours attribute in the storage units component implements an energy-to-power ratio, ensuring technically feasible charging and discharging while preventing oversized capacities.

We now added an energy-to-power ratio for the small water tank to distinguish between central and decentralized heat storage options. This change makes the model less restrictive.

Adjusting the storage unit capital costs to depend on power capacity rather than energy capacity (https://github.com/PyPSA/pypsa-eur/pull/1444#discussion_r1892675751) resulted in a slight increase in total system costs. The costs now closely match those of the prior store implementation.

The total costs of the storage units do not exceed those of the stores due to corrections in the handling of charging efficiencies:
- For example, previously (in the store implementation), the dis-/charging efficiency for PTES stores was set at 0.8367, which already included standing losses. However, standing losses were also added separately, leading to double counting.
- In the storage units, this issue has been resolved. Charging and discharging efficiencies are now directly taken from the DEA Technology Catalogue, with standing losses handled separately.

As a result, the costs of the storage units are now nearly the same as those of the stores.

lkstrp:
@TomKae00 @amos-schledorn 
Ok those changes have been reverted because of failing tests. You have to create a new PR again from the same branch now

PR #1468, opened 2024-12-19
feat: check objective in test routines

FabianHofmann:
Option to counter-check the objective value against expected value. With this check, we make sure that no strong deviations from the expected objective are introduced. In case the objective varies stronger than the allowed tolerance, but for a good reason (new tech/tech costs), the expected value has to be updated manually. 

## Checklist


config/config.default.yaml
```diff
@@ -1012,6 +1012,12 @@ solving:
     cbc-default: {} # Used in CI
     glpk-default: {} # Used in CI
 
+  check_objective:
+    enable: false
+    expected_value: None
+    atol: 1000
+    rtol: 0.001
```
fneum, line 1019: I would be more generous for the default (absolute value of 1000 may often be within solver tolerance).
```suggestion
    atol: 1000000
    rtol: 0.01
```

scripts/solve_network.py
```diff
@@ -986,6 +990,19 @@ def extra_functionality(n, snapshots):
         custom_extra_functionality(n, snapshots, snakemake)
 
 
+def check_objective_value(n, solving):
+    check_objective = solving["check_objective"]
+    if check_objective["enable"]:
+        atol = check_objective["atol"]
+        rtol = check_objective["rtol"]
+        expected_value = check_objective["expected_value"]
+        if not np.isclose(n.objective, expected_value, atol=atol, rtol=rtol):
```
fneum, line 999: Should we check `n.objective` or `n.statistics.capex().sum() + n.statistics.opex().sum()`?

fneum (commented):
Good utility!

scripts/solve_network.py
FabianHofmann: I would say objective just to have one loop less.

lkstrp:
Would this have catched the linopy bug? I guess no since the test configs are to small?

Makefile
```diff
@@ -54,23 +54,23 @@ install-pinned-macos: _conda_check
...
-	snakemake solve_elec_networks --configfile config/test/config.electricity.yaml --rerun-triggers=mtime
-	snakemake --configfile config/test/config.overnight.yaml --rerun-triggers=mtime
-	snakemake --configfile config/test/config.myopic.yaml --rerun-triggers=mtime
-	snakemake make_summary_perfect --configfile config/test/config.perfect.yaml --rerun-triggers=mtime
-	snakemake --configfile config/test/config.scenarios.yaml --rerun-triggers=mtime -n
+	snakemake solve_elec_networks --configfile config/test/config.electricity.yaml
+	snakemake --configfile config/test/config.overnight.yaml
+	snakemake --configfile config/test/config.myopic.yaml
+	snakemake make_summary_perfect --configfile config/test/config.perfect.yaml
+	snakemake --configfile config/test/config.scenarios.yaml -n
```
FabianHofmann, line 61: @fneum This is always committed by me unintentionally.  but I like it more this way to make the tests really test everything. would you be fine by keeping it?

FabianHofmann:
> Would this have catched the linopy bug? I guess no since the test configs are to small?

I think so yes! the exclusion of balance contraints which happened was altering the optimization quite significantly.

Makefile
fneum: Yes, just means the tests might run a little longer, but that is fine.

PR #1472, opened 2024-12-30
remove {ll} wildcard and replace with config setting

### LINKED ISSUE #1354: Remove `{ll}` wildcard?
It looks to me like the `{ll}` wildcard remains mainly as a historical artefact -- I'm guessing this goes back to "The role of spatial scale in joint optimisations of generation and transmission for European highly renewable scenarios". However, today I don't really see any particular reason why transmission expansion limits should get their own wildcard? To me it would make perfect sense to make this a configuration option instead, and add a flag to the `{opt}`/`{sector_opts}` wildcards instead.

It's not really consequential, and the main drawback would be a disruptive change for users, having to deal with a new file-name structure.

If it's worth considering breaking backwards-compatibility on the file-name like this, there are some other things to consider:
- https://github.com/PyPSA/pypsa-eur/issues/708
- A little bit more radical, but I would love to merge the `{opts}`/`{sector_opts}` wildcards, eliminating confusion for new users and making it impossible to falsely put options in both of them. Instead, whether a network is electricity-only or sector-coupled should, in my opinion, be a configuration option.

fneum:
Closes #1354.

## Checklist


fneum:
Functionality verified locally.

PR #1479, opened 2025-01-03
Updating all base shapes (country_shapes, europe_shape, nuts3_shapes)

bobbyxng:
Closes # (if applicable).

- Updating all base shapes (country_shapes, europe_shape, nuts3_shapes)
- The workflow has been modified to use higher resolution and more harmonised shapes (NUTS3 and OSM administration level 1):
  - `build_shapes`: Use NUTS3 2021 01M data, `nuts3_shapes.geojson` now includes all NUTS levels (0/country, 1-3) in preparation for another PR allowing for country-specific settings for regional clustering (based on NUTS and ADM1). Population and GDP p.c. data is now included for all PyPSA-Eur regions
  - `retrieve_osm_boundaries`: Added to allow for retrieving country shape directly from OpenStreetMap using Overpass turbo. Used only for non 2021 NUTS3 countries (BA, MD, UA, XK). Known shape projects like geoboundaries and gadm each have their own issues (offset country borders, leaving large gaps or providing the wrong ADM level in the files).
  - `build_osm_boundaries`: Building ADM1 level shapes/boundaries (BA, MD, UA, XK)
  - Data sources updated to JRC ARDECO https://urban.jrc.ec.europa.eu/ardeco/, as the dataset is harmonised with NUTS3 2021, and still includes UK regions, as well as RS, and CH
  - `build_gdp_pop_non_nuts3`: Previously needed for MD and UA, now included into `build_shapes` to create standardised data at the very beginning for all four non NUTS3 countries. For this purpose, the cutouts of the datasets `GDP_per_capita_PPP_1990_2015_v2.nc` and `ppp_2019_1km_Aggregated.tif` have been updated
  - Population and GDP now refer to the year 2019.

image

# Open to dos:
- [x] Update `GDP_per_capita_PPP_1990_2015_v2.nc` and `ppp_2019_1km_Aggregated.tif` in the databundle (--> @fneum)
- [x] Remove sandbox data and rule after updating databundle
- [x] Points in checklist
- [x] Quick comparison of previous model runs vs. new ones. Distributions largely remains the same.

## Checklist

### Country level
Note that their can be small differences due to regional borders "moving" due to the switch from natural earth country borders to Eurostat NUTS3 01M (and OSM)
image

### Nodal load distribution
For **country plots** see: https://github.com/PyPSA/pypsa-eur/pull/1479#issuecomment-2592928436
Explanation for changes:
- Changes can be due to NUTS borders shifting, changes in GDP p.c. and/or population from 2014 to 2019.
- Further, countries like AL, BA, RS and XK where previously only assigned a load on a national level. 
- For UA and MD, load is now determined in `build_shapes` already and mapped at ADM1 level (as opposed to onshore regions level in `build_gdp_pop_non_nuts3` previously (the latter having a "pseudo" higher resolution: While the number of regions was higher, the underlying dataset for pop and gdp was at lower resolution),
image

### Biomass distribution (for 128 buses)
image

image

image

image


rules/build_electricity.smk
```diff
@@ -98,17 +98,47 @@ rule base_network:
...
+rule build_osm_boundaries:
+    params:
+        countries=config_provider("countries"),
+    input:
+        json="data/osm-boundaries/json/{country}_adm1.json",
+        eez=ancient("data/eez/World_EEZ_v12_20231025_LR/eez_v12_lowres.gpkg"),
+    output:
+        boundary="data/osm-boundaries/build/{country}_adm1.geojson",
+    log:
+        logs("build_osm_boundaries_{country}.log"),
```
fneum, line 110:
```suggestion
        "logs/build_osm_boundaries_{country}.log",
```

bobbyxng:
@lkstrp @finozzifa I removed the test_country_cover part from unit_testing (https://github.com/PyPSA/pypsa-eur/pull/1466), as in this PR, `countries()` does not exist anymore. The proposed workflow around building shapes is now based entirely around NUTS3 shapes (grouped to different levels to allow for regional clustering in another PR later).

bobbyxng:
image

image

image

image

image

image

image

image

image

image

image

image

image

image

image

image

image

image

image

image

image

image

image

image

image

image

image

image

image

image

image

image

image

image

image

image

bobbyxng:
@fneum @Irieo : Done with the comparisons, as a remaining todo, the data bundle for `GDP_per_capita_PPP_1990_2015_v2.nc` and `ppp_2019_1km_Aggregated.tif` need to be updated (see sandbox url) and the sandbox code needs to be removed/adapted accordingly.

FabianHofmann:
Impressive work! About the unit tests, do you think it would make sense to add new ones for the new functions? Perhaps our unit test master @finozzifa could help out? 
(For background, I would love to come to the point where we can quickly detect deprecation and ease bug fixing; it would make the overall code base more stable)

bobbyxng:
> Impressive work! About the unit tests, do you think it would make sense to add new ones for the new functions? Perhaps our unit test master @finozzifa could help out? (For background, I would love to come to the point where we can quickly detect deprecation and ease bug fixing; it would make the overall code base more stable)

Thanks @FabianHofmann ! Some of the functions are quite specific to the dataset. In this case, it would only make sense if we further break down the functions. Then again we would need to generate some dummy data for testing purposes. While I am generally supporting any move towards detecting deprecation and bugs quickly, I am not sure if this specific case is a good example of showing the benefits. Open to discussing this further :)

finozzifa:
hey @FabianHofmann and @bobbyxng,

I believe that we can unit test functions even if we do not re-factor them immediately. In other words, unit testing and code re-factoring can happen at different stages. My feeling is that it is safer to develop new code while writing at the same time unit tests for the new or existing functions (as for example we did at https://github.com/PyPSA/technology-data/pull/160).

I am of course very happy to support :)

FabianHofmann:
you are right @finozzifa, perhaps we can say it would a very-nice-to-have for this pr, but should not be a blocker. up to @bobbyxng I would say :)

rules/build_sector.smk
```diff
@@ -468,6 +468,9 @@ rule build_heat_totals:
         "../scripts/build_heat_totals.py"
 
 
+# TODO: update long-term to only need NUTS2021, nuts3_shapes.geojson
+# However: requires manual updating of NUTS code changes 2016 to 2021:
+# https://ec.europa.eu/eurostat/web/nuts/history
```
fneum: perhaps move to issue

fneum (approved):
Data bundle is now updated alongside necessary code adjustments.

PR #1481, opened 2025-01-03
Resource Classes

### LINKED ISSUE #290: Different wind turbine models + option for deploying wind turbines in different wind class areas
We currently only use one wind turbine model (power curve) for all wind turbine locations.

Potential improvements:

* Use a more modern wind turbine model / power curve (check if update is needed)
* Evaluate different locations for their wind class (I to IV) and deploy an appropriate wind turbine (rather than always the same turbine)

### LINKED ISSUE #713: Use wind classes for turbine placement (based on GASP)
GASP = Global Atlas for Siting Parameters

High-resolution (250m) worldwide dataset with among others recommendations for which wind turbine type to place where from DTU/GWA group: https://onlinelibrary.wiley.com/doi/epdf/10.1002/we.2771

image

It could be a good improvement to distinguish wind turbine classes (based on IEC + S + T types), determine their potentials per grid cell based on the GASP dataset and derive wind turbine profiles for the individual classes for each clustered node (instead of single CF time-series and potential per grid cell with generalised power curve assumption)

fneum:
Closes #713.
Closes #290.

**This PR adds the option to split renewable potentials and time series into a configurable number of resource classes per carrier and clustered region. The binning is done based on the average capacity factors.**

```yaml
renewables:
  onwind:
    resource_classes: 4
```

With the above setting, each region would get 4 onshore wind generators, each with different `p_max_pu` and `p_nom_max`. This allows more accurate renewable time series even when the spatial resolution might be highly aggregated (e.g. one region per country).

The default number of resource classes is kept at one per cluster, which is recommended for scenarios with high spatial resolution.

The feature replaces the pre-clustering workaround that used to be in `simplify_network` to have multiple renewable generators of one carrier inside a single cluster.

The **main magic** happens in `build_renewable_profiles`. Mainly, a new dimension "bin" is added to the output `xarray.Dataset` and a new `.geojson` of clustered regions split by resource classes is exported, which later helps with assigning existing wind and solar capacities not only to the buses but also to the correct resource class. The resource classes are numbered from 0 (lowest capacity factor segment) to N (highest).

Here's an example clustered region ("BE0 0") with the average onshore wind capacity factors of the cutout grid cells involved:

step1

If the number of resource classes is 2, each cutout grid cell of the clustered region would be allocated to one of the two resource classes. In this simple case, the corresponding bins would be defined by the borders `cf.min()`, `cf.min() + (cf.max() - cf.min()) / 2` and `cf.max()` - i.e. simple linear spacing. *Note that the bins are specific to the clustered region! Different regions have different splits!*

step2

Based on this, we can now also split the shapefile of the clustered region (e.g. to allocate existing RES capacities to the right resource class). *Note that these splits do not have to be contiguous! You could have resource class 0 in the North and South and resource class 1 in the middle!*

step3

Since the resource class indicator matrix is multiplied by the availability matrix, the potentials / land eligibility are also geographically split. Of course, the overlaps of involved grid cells into other clustered regions are also taken care of! The potentials (`p_nom_max`) now have coordinates (bus, bin):

```
<xarray.DataArray (bus: 5, bin: 2)> Size: 80B
array([[ 344.37801566, 1978.24694203],
       [ 300.90176106, 2388.85578706],
       [3000.08993939,  365.11551634],
       [1939.36472633, 3834.88855738],
       [ 318.39363366,  323.66654609]])
Coordinates:
  * bus      (bus) <U5 100B 'BE0 0' 'BE0 1' 'BE0 2' 'BE0 3' 'BE0 4'
  * bin      (bin) int64 16B 0 1
```

The output of `build_renewable_profiles` now looks like this (which is then handled in the subsequent rules and scripts):

image

Another feature added as an aside is the use of **Global Energy Monitor (GEM)** renewable capacity data, which now replaces the less complete OPSD VRE data in `add_electricity`. As a note for a future PR, this should also be used for `add_existing_baseyear` very soon, but I intentionally left it out here to reduce complexity.

Furthermore, a small functionality reduction is included in this PR. The **technology mapping** from `powerplantmatching` to `pypsa` now has to be 1:1 rather than 1:n. I.e. existing offshore wind has to be `-dc` or `-ac`. Since it aligns with how it is done in `add_existing_baseyear`, in my view, this is a negligible drawback.

## TODO

- [x] Some testing and benchmarking for full-scale runs is required.

## Checklist

- [x] I tested my contribution locally and it works as intended. (for CI tests)


fneum:
FYI @oruhnau @aleks-g (it still requires a little bit of testing etc.)

fneum:
Some tests for onshore wind all seem plausible:

With two bins, many regions use one bin for cells close to the shore and the other bin for inland regions.

image

As expected, compared to a single-bin run, average capacity factors for one bin are higher and lower for the other bin:

image

Potentials summed up across bins match exactly between single- and multi-bin runs:

image

fneum:
> Thanks a lot, @fneum, for implementing this!
> 
> I tested it and can confirm that the build_renewable_profiles rule is working as expected.
> 
> However, I get an error in the add_electricity rule:
> 
> ```
> File "\pypsa-eur\.snakemake\scripts\tmpjcxt_pns.add_electricity.py", line 1104, in <module>
>     attach_GEM_renewables(n, tech_map)
>   File "\pypsa-eur\.snakemake\scripts\tmpjcxt_pns.add_electricity.py", line 777, in attach_GEM_renewables
>     tech_string = ", ".join(tech_map.values())
>                   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
> TypeError: sequence item 0: expected str instance, list found
> ```
> 
> Any idea how this could be solved?

I think this is an interaction with an outdated configuration. With this PR, the technology mapping must be a string, not a list.

```yaml
electricity:
  estimate_renewable_capacities:
    technology_mapping:
      Offshore: offwind-ac
      Onshore: onwind
      PV: solar
```

PR #1490, opened 2025-01-10
Hot fix for compatibility with split between solid biomass and biogas in new ppm version

daniel-rdt:
Introduces a hot fix for compatibility with split between solid biomass and biogas in new ppm version.
It is still required to properly address the implementation of solid biomass and biogas in pypsa-eur.

@fneum Not sure if a release note is needed for this.

## Checklist

### DIFF
--- a/.github/workflows/push-images.yaml
+++ b/.github/workflows/push-images.yaml
@@ -8,7 +8,7 @@ on:
   workflow_dispatch:
 
 env:
-  IMAGE_NAME: eur-dev-env
+  IMAGE_NAME: pypsa/eur-dev-env
   BASE_ENV: envs/linux-pinned.yaml
 
 jobs:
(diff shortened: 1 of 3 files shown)


lkstrp (commented):
Not sure if I like the level of hotfixness 😄 
You need it to get your fix in powerplantsmatching?

This certainly needs a big `#TODO` here and a note in `environment.yaml` where powerplantmatching is pinned.

We can also just release ppm `0.5.20` with your fix, which might be much cleaner.

daniel-rdt:
> Not sure if I like the level of hotfixness 😄 You need it to get your fix in powerplantsmatching?
> 
> This certainly needs a big `#TODO` here and a note in `environment.yaml` where powerplantmatching is pinned.
> 
> We can also just release ppm `0.5.20` with your fix, which might be much cleaner.

I definitely understand your reservations here 😄 

It indeed was intended so that I can use my bug fix from ppm. I was discussing this with @fneum the other day and it seemed like feasible intermediate solution to allow compatibility also for possible other bug fixes. But your approach would be definitely cleaner and more consistent between packages. It would also move addressing the implementation of solid biomass and biogas up the list of priorities for PyPSA-Eur, I guess.

So, I like releasing a ppm 5.20 which would be sufficient for me to solve my issue.

lkstrp:
Releasing `0.5.20` is rather annoying because of some merge conflicts when cherry picking.

So if Fabian agrees, let's go with the hotfix. It also has the advantage that the added MASTR dataset can be used.

fneum:
@lkstrp hotfix is ok!

fneum (approved):

lkstrp:
Should we remove the ppm pin as well?

fneum:
ah, yes!

PR #1514, opened 2025-01-26
add scheme to deprecate/rename config entries

FabianHofmann:
## Add scheme for deprecating config entries and warn for config validation

This PR introduces specific warning classes for deprecated and invalid configuration entries. A scheme is proposed to deprecate and optionally rename config entries through the `config/deprecations.yaml`.

Like this snakemake raises:

- a `DeprecationConfigWarning` for deprecated config entries 
- a `InvalidConfigWarning` for unsupported/invalid config entries which are not included in config.default.yaml

### Testing

Added assertions to verify warning classes in:
- `test_config_deprecations()`
- `test_config_invalid_entries()`

## Checklist

### DIFF
--- a/Snakefile
+++ b/Snakefile
@@ -10,7 +10,14 @@ from snakemake.utils import min_version
 
 min_version("8.11")
 
-from scripts._helpers import path_provider, copy_default_files, get_scenarios, get_rdir
+from scripts._helpers import (
+    path_provider,
+    copy_default_files,
+    get_scenarios,
+    get_rdir,
+    check_deprecated_config,
+    check_invalid_config,
+)
 
 
 copy_default_files(workflow)
@@ -19,6 +26,9 @@ copy_default_files(workflow)
 configfile: "config/config.default.yaml"
 configfile: "config/config.yaml"
 
+
+check_deprecated_config(config, "config/deprecations.yaml")
+check_invalid_config(config, "config/config.default.yaml")
 
 run = config["run"]
 scenarios = get_scenarios(run)
--- a/config/deprecations.yaml
+++ b/config/deprecations.yaml
@@ -1,0 +1,8 @@
+# SPDX-FileCopyrightText: Contributors to PyPSA-Eur <https://github.com/pypsa/pypsa-eur>
+#
+# SPDX-License-Identifier: MIT
+
+# Format: list of deprecation entries with:
+# - old_entry: Dot-separated path to deprecated entry (e.g. "electricity:co2_limit")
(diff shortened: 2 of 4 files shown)


fneum:
This is cool!

lkstrp (changes requested):
While I strongly support any config validation, I am not sure if we wanna implement custom functionality for that.

If this is only about Deprecations, it's fine. But when checking for invalid configs, specially if we add more advanced checks, this will be much easier and simpler with some data validation library.

There are json schema validation libraries out there, but if I point to [similar](https://github.com/PyPSA/PyPSA/pull/1128) validations in PyPSA and the data validation layers discussed on Friday, which would be really useful in the mid term within pypsa-eur, maybe we can go for Pydantic for config checks as well.

I can draft something how this could look for json data. But I think it is better than introducing another custom solution which will be much harder to extend

fneum:
I think it’s for deprecation. For validation of cong filers there is already built in functionality in snakemake: https://snakemake.readthedocs.io/en/stable/snakefiles/configuration.html#validation

lkstrp:
Great, then json schema is already there. But this is even better, we don’t need any extra imports and have everything:
- `InvalidConfigWarning`: Value validation fails
- `DeprecationConfigWarning`: Key is not in schema

Json schema allows for basic value checks which should be enough for config validation. 
And also has a deprecation property, which raises those warnings. This needs to be checked with snakemake, but looks like its build on top and should handle that all.

FabianHofmann:
Gotcha, so perhaps the naming is a bit misleading, the implementation only checks whether a **key** provided by the user is deprecated or not supported by the default.yaml. That said I am totally open to generalize this and use a more solid json schema from snakemake which might even check values. I am not sure though, how well these can be incorporated with our scnenario configs, but there is definitely a way. I will take a look. Also the question arises whether it is worth to define the scheme now, before we are doing some potential restructuring changes that might also touch the config ( in the preparation of  which I made this PR).

fneum:
@FabianHofmann, yes I see this as two independent steps. Key deprecation can go ahead.

FabianHofmann:
okay, given the considerations of @lkstrp I would even consider that we leave this PR and make the whole thing via the jsonschema. there my questions would only be 

* can the jsonschema indicate deprecations and custom messages for those?
* can we set the condition that all configurations used (also via the scenario management) must be in the jsonschema and a warning is returned if not

@lkstrp since know more about this, any clue if this is supported? from your answer above it seems so

PR #1516, opened 2025-01-27
Fix geothermal heat potentials

amos-schledorn:
The current geothermal heat potentials from Manz et al. 2024 (`https://www.sciencedirect.com/science/article/pii/S0960148124001769`) are likely inflated because we use those before Manz et al mapped them to what they assume to be future district heating areas (different from PyPSA-Eur but urban areas). 
This was discovered in correspondence with the authors.

- This update uses an Excel output from Manz et al rather than a geopackage. The new data has already been mapped to future district-heating areas.
- The new data is provided on LAU-region level which heavily simplifies the spatial matching. That allows us to drop the `OnshoreRegion` class and do a simple matching. LAU region geo data is now retrieved for this rule.
- In preparation for integrating further heat sources, the previous generic mapping of any ISI heat potentials (i.e. those computed by Manz et al.) has been replaced by a dedicated geothermal potential calculation. In this process, some unneeded config settings have been dropped.

## Open issue
- LAU regions might sometimes be mapped incorrectly to onshore regions. I suggest tracking this in a separate issue

## Checklist

### Config
```
scenario:
  clusters:
    - 40
  planning_horizons:
    - 2030

countries: ['DE']

foresight: overnight

clustering:
  temporal:
    resolution_sector: 365h


sector:
  heat_pump_sources:
      urban central:
      - air
      - geothermal
```

### Results
As expected, geothermal heat utilisation decreases:

##### Master
image

##### Feature
image


fneum (changes requested):
Looks good as far as I can see. Once @cpschau gives the go, can be merged.

The one thing I request: For `mock_snakemake` we now move all downloads to dedicated `retrieve*` rules (https://github.com/PyPSA/pypsa-eur/pull/1274). This should be done for the storage() inputs here as well.

fneum (approved):

PR #1528, opened 2025-02-04
remove `highspy` pin

lkstrp:


fneum:
Should 1.8 be blocked? This was the version that caused problems right?

lkstrp:
`1.9` made some problems at some point as well. Not sure if this was because of sub dependencies or something else.

Also, our blocked dependencies are not complete at all. And this is a conda env, not a package dependency list, so it will not be installed in other environments and blocking old versions will not make much difference anyway. 

I think it's fine to not block old versions and reference the pinned env

FabianHofmann (approved):

FabianHofmann:
ah, you tackled the jupyter issue here already, could you add entsoe-py as well as done in #1531?

PR #1532, opened 2025-02-06
Add an option to overwrite cost attributes from the configuration file.

tgilon:

This PR suggests to add an options to overwrite investment, lifetime, FOM, VOM, efficiency and fuel attributes from the configuration file. This mimics the existing capital and marginal cost behaviour.

## Checklist


config/config.default.yaml
```diff
@@ -864,6 +864,10 @@ costs:
     fuel cell: 0.
     battery: 0.
     battery inverter: 0.
+#  fuel:
+#    gas: 38.84
+#    OCGT: 38.84
+#    CCGT: 38.84
```
fneum, line 870: remove

config/config.default.yaml
```diff
@@ -851,7 +851,7 @@ costs:
     lifetime: 25
     "CO2 intensity": 0
     "discount rate": 0.07
-  # Marginal and capital costs can be overwritten
+  # Marginal and capital costs, investment, lifetime, FOM, VOM, efficiency and fuel can be overwritten
```
fneum, line 854: wondering if those overwrites could be packaged into

```yaml
costs:
  overwrites:
    marginal_cost: ...
```
even though this would be a breaking change for `marginal_cost` and `capital_cost`.

fneum (approved):
LGTM

config/config.default.yaml
fneum: Let's hear another opinion @FabianHofmann @lkstrp

FabianHofmann: I would also prefer a dedicated section, too. and this breaking change would be "softened" by the json schema since the user will get a clear message that the old place is not supported anymore

lkstrp: Agree, that should go in `overwrites`. But I would leave it non breaking for now (just move new attributes to `overwrites` and break once we have the deprecation messages.

Also, shouldn't `fill_values` (`defaults` is a better name here) and `overwrites` then allow for the exact same keys?

fneum: > Also, shouldn't fill_values (defaults is a better name here) and overwrites then allow for the exact same keys?

Yes, with the exception of `marginal_cost` and `capital_cost`.

fneum: resolved in #1565

fneum:
I will close this in favour of  #1565

PR #1537, opened 2025-02-09
Consistent function scope

FabianHofmann:
One step from #1529 

With this change the functions from add_*, prepare_* and solve_network.py become stand-alone and portable. The previously reliance on the global `snakemake` object was removed by using explicit argument. 

A CI test was added to check the usage of undefined variables by pylint.  

With this there is no obstacle to distribute the functions definitions in `prepare_sector_network.py` to 
- add_industry.py
- add_transport.py
- add_heating.py
- add_...

## Checklist

### DIFF
--- a/scripts/solve_network.py
+++ b/scripts/solve_network.py
@@ -31,6 +31,7 @@ import logging
 import os
 import re
 import sys
+from typing import Any
 
 import numpy as np
 import pandas as pd
@@ -55,9 +56,19 @@ class ObjectiveValueError(Exception):
     pass
 
 
-def add_land_use_constraint_perfect(n):
+def add_land_use_constraint_perfect(n: pypsa.Network) -> None:
     """
     Add global constraints for tech capacity limit.
+
+    Parameters
+    ----------
+    n : pypsa.Network
+        The PyPSA network instance
+
+    Returns
+    -------
+    pypsa.Network
+        Network with added land use constraints
     """
     logger.info("Add land-use constraint for perfect foresight")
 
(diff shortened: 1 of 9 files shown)


fneum (approved):
Looks great! Thanks also for the added docstrings!

FabianHofmann:
It was added to the environment. maybe the pinned ones were not updated yet? 

This is quite a special case for which ruff does not provide an equivalent. We need all functions in prepare and add scripts to be standalone and not being dependent on global variables in order to restructure the workflow with steps drafted (but discussed further internally with @fneum) in #1529 

but if you have a solution with ruff, more then happy to apply that

lkstrp:
`F821 Undefined name` is already enabled, but ruff does not catch `snakemake` because it is defined as global in the main guard of the script.

This case is not special at all, defining all functions as pure is very much appreciated! We could not only make them standalone, but also move them into a library. This would be most consistent and resolve using pylint only for a single check, which is a mess.

FabianHofmann:
> `F821 Undefined name` is already enabled, but ruff does not catch `snakemake` because it is defined as global in the main guard of the script.

Exactly, the same counts for `spatial`, `options` and so on

> This case is not special at all, defining all functions as pure is very much appreciated! We could not only make them standalone, but also move them into a library. This would be most consistent and resolve using pylint only for a single check, which is a mess.

We made the case special because we always used global variables which makes things super complicated to clear up. I totally agree and don't like the special pylint and totally want to remove it after we distribute the functions into package modules. But until then I would like to prevent people from continuing relying on global vars

scripts/solve_network.py
```diff
@@ -1089,12 +1191,12 @@ def solve_network(n, config, params, solving, **kwargs):
     with memory_logger(
         filename=getattr(snakemake.log, "memory", None), interval=logging_frequency
     ) as mem:
-        n = solve_network(
+        solve_network(
             n,
             config=snakemake.config,
             params=snakemake.params,
             solving=snakemake.params.solving,
-            log_fn=snakemake.log.solver,
```
coroa, line 1097: @FabianHofmann Why did you remove `log_fn`?

PR #1554, opened 2025-02-25
feat(ci): Remove defaults channel from conda setup and use strict channel prio

coroa:
With the default settings the generated pinned environments may still contain references to the defaults channel in certain cases (currently to zstandard), and the test workflow does not fail because it also falls back to the defaults channel.

Was able to reproduce locally the creation of a locked environment with conda
which could not be installed by mamba (unless the defaults channel was explicitly
configured). Ensuring the defaults channel does not appear in `~/.condarc` is
enough to create locked environments which work with my current mamba setup.

Strict channel priority should improve the situation further.


coroa:
Success :clown_face:  https://github.com/PyPSA/pypsa-eur/actions/runs/13530573076/job/37811548433#step:10:21 failed with the expected `mkl=3.0` complaint.

fneum (approved):
Not opposed :)

PR #1560, opened 2025-02-27
fix(solve): Fix planning_horizons in solve_network

coroa:
Follow-up to #1537 .

Closes #1551 .

In perfect foresight, the planning_horizons param remains the list
of all planning horizons (to trigger re-runs).

The current planning horizons is always taken from the planning_horizons
wildcard and passed into extra_functionality as a partial kwarg.

## Checklist

### DIFF
--- a/rules/solve_electricity.smk
+++ b/rules/solve_electricity.smk
@@ -7,7 +7,6 @@ rule solve_network:
     params:
         solving=config_provider("solving"),
         foresight=config_provider("foresight"),
-        planning_horizons=config_provider("scenario", "planning_horizons"),
         co2_sequestration_potential=config_provider(
             "sector", "co2_sequestration_potential", default=200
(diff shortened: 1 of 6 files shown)


FabianHofmann (approved):

yechenyan:
In this PR, if CCL is added to the perfect forecast, errors will occur as shown in the following image：

image

This is because the network is missing `planning_horizons`.
I think it doesn't affect the merging of this PR; this is just a reminder that there are still issues when adding CCL to the perfect forecast.

coroa:
@yechenyan you are right, i added now a very concrete error message, that CCL has not been implemented for perfect foresight, yet

coroa:
the `n.planning_horizons` was actually an intermediate idea that leaked into the commit. sorry for that.

PR #1570, opened 2025-03-11
enh(add_existing_baseyear): Pull out valid_grouping_years algorithm from heat

coroa:

Minimal refactoring to re-use the grouping_years validation also for power. And fix a wrong indent.

## Checklist

### DIFF
--- a/config/config.default.yaml
+++ b/config/config.default.yaml
@@ -429,8 +429,8 @@ solar_thermal:
 
 # docs in https://pypsa-eur.readthedocs.io/en/latest/configuration.html#existing-capacities
 existing_capacities:
-  grouping_years_power: [1920, 1950, 1955, 1960, 1965, 1970, 1975, 1980, 1985, 1990, 1995, 2000, 2005, 2010, 2015, 2020, 2025]
-  grouping_years_heat: [1980, 1985, 1990, 1995, 2000, 2005, 2010, 2015, 2019] # heat grouping years >= baseyear will be ignored
+  grouping_years_power: [1920, 1950, 1955, 1960, 1965, 1970, 1975, 1980, 1985, 1990, 1995, 2000, 2005, 2010, 2015, 2020, 2025] # power grouping years > baseyear will be ignored
+  grouping_years_heat: [1980, 1985, 1990, 1995, 2000, 2005, 2010, 2015, 2019] # heat grouping years > baseyear will be ignored
   threshold_capacity: 10
   default_heating_lifetime: 20
   conventional_carriers:
--- a/scripts/add_existing_baseyear.py
+++ b/scripts/add_existing_baseyear.py
@@ -7,6 +7,7 @@ horizon.
 """
 
 import logging
+from collections.abc import Sequence
 from types import SimpleNamespace
 
 import country_converter as coco
@@ -464,6 +465,25 @@ def get_efficiency(
     return efficiency
 
 
+def valid_grouping_years(
+    grouping_years: Sequence[int] | pd.Index,
+    baseyear: int,
+    max_lifetime: float | None = None,
+) -> pd.Index:
+    grouping_years = pd.Index(sorted(grouping_years)).astype(int)
+
+    valid = grouping_years <= baseyear
+    if max_lifetime is not None:
+        valid &= grouping_years + max_lifetime >= baseyear
+    if not valid.all():
+        logger.warning(
+            "Only grouping years between [baseyear - max. lifetime, baseyear] are used."
+            f"Dropping {grouping_years[~valid]}."
+        )
+
+    return grouping_years[valid]
+
+
 def add_heating_capacities_installed_before_baseyear(
     n: pypsa.Network,
     costs: pd.DataFrame,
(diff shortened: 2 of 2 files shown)


config/config.default.yaml
```diff
@@ -429,8 +429,8 @@ solar_thermal:
 
 # docs in https://pypsa-eur.readthedocs.io/en/latest/configuration.html#existing-capacities
 existing_capacities:
-  grouping_years_power: [1920, 1950, 1955, 1960, 1965, 1970, 1975, 1980, 1985, 1990, 1995, 2000, 2005, 2010, 2015, 2020, 2025]
-  grouping_years_heat: [1980, 1985, 1990, 1995, 2000, 2005, 2010, 2015, 2019] # heat grouping years >= baseyear will be ignored
+  grouping_years_power: [1920, 1950, 1955, 1960, 1965, 1970, 1975, 1980, 1985, 1990, 1995, 2000, 2005, 2010, 2015, 2020, 2025] # power grouping years > baseyear will be ignored
+  grouping_years_heat: [1980, 1985, 1990, 1995, 2000, 2005, 2010, 2015, 2019] # heat grouping years > baseyear will be ignored
```
FabianHofmann, line 433: this is interesting, so the filtering was (partially) already there for heat technology?

scripts/add_existing_baseyear.py
```diff
@@ -464,6 +465,25 @@ def get_efficiency(
...
 
 
+def valid_grouping_years(
+    grouping_years: Sequence[int] | pd.Index,
+    baseyear: int,
+    max_lifetime: float | None = None,
+) -> pd.Index:
+    grouping_years = pd.Index(sorted(grouping_years)).astype(int)
+
+    valid = grouping_years <= baseyear
```
FabianHofmann, line 475: I am always confused about the type of `baseyear`, the old code explicitly used `int(baseyear)`, guess not needed as snakemake converts the dtype automatically of the wildcard to int?

scripts/add_existing_baseyear.py
```diff
@@ -532,32 +552,14 @@ def add_heating_capacities_installed_before_baseyear(
...
-            assert valid_grouping_years.is_monotonic_increasing
-
-            # get number of years of each interval
-            _years = valid_grouping_years.diff()
-            # Fill NA from .diff() with value for the first interval
-            _years[0] = valid_grouping_years[0] - baseyear + default_lifetime
-            # Installation is assumed to be linear for the past
-            ratios = _years / _years.sum()
+        # get number of years of each interval
+        _years = pd.Index([grouping_years[0] - baseyear + default_lifetime]).append(
```
FabianHofmann, line 556: I am a bit confused about this expression here, this can be a negative number or? If grouping_year[0] == 1990 and baseyear == 2020 and lifetime == 20 for example.

scripts/add_existing_baseyear.py
coroa: The same is actually true for `grouping_years` so that we could also get rid of the `.astype(int)`

scripts/add_existing_baseyear.py
coroa: I must admit, i did not think about this. It is a literal port of Amos':
```
# Fill NA from .diff() with value for the first interval
_years[0] = valid_grouping_years[0] - baseyear + default_lifetime
```

coroa: It sounds like there was some back and forth, i'll not try to understand this fully now, maybe its best to have a short discussion.

amos-schledorn: Sorry for the late reply @coroa! I think I only made some small refactoring changes here. I'll be OOO for a week but happy to discuss this further then, if helpful.

amos-schledorn (commented):
I haven't really worked with this, so I'm a little out of my comfort zone. I'd just suggest a function docstring for `valid_grouping_years`. Otherwise happy if @FabianHofmann is happy.

coroa:
@lindnemi if it makes sense, we can also have a short call next week to try to see what to do with it, and what would be the ideal aspiration.

lindnemi:
For a proper refactor I would suggest to unify `grouping_years_heat` and `grouping_years_power` from the config, and to unify how existing capacities are treated (currently overwrite `p_nom_min`/exclude the baseyear). I think ideally we would distinguish historical components from  components that are active during optimization, possibly by adding `historical` to the name of the components.

coroa:
Hi @lindnemi , funnily i am now also coming back with only quite a delay to this PR.

Thanks for your special case description, ie point 2, i was not aware of this tricky difference.

What do you think are the benefits of splitting the extendable capacity from the fixed capacity (with a historical suffix) in comparison to the p_nom/p_nom_min combination?

coroa:
And thanks for point 3. That is indeed a mistake.

lindnemi:
> Hi @lindnemi , funnily i am now also coming back with only quite a delay to this PR.
> 
> Thanks for your special case description, ie point 2, i was not aware of this tricky difference.
> 
> What do you think are the benefits of splitting the extendable capacity from the fixed capacity (with a historical suffix) in comparison to the p_nom/p_nom_min combination?

Only benefit is that the distinction between historical assets and expanded assets would be clearer, but on second thought it's clear enough already, so let's just leave it as is. The harmonization between electricity and heating sector would have more benefits.

fneum:
Where do we land here @lindnemi @coroa?

PR #1574, opened 2025-03-12
Add an option to create a filegraph in the dag rule

tgilon:
Closes # (if applicable).

## Checklist


fneum (changes requested):
Nice utility! But avoid definition of `dag_type` outside rule definition.

It seems the current approach mixes two approaches (wildcards and parameters). With the current setup you could actually use

```
snakemake --{dag_type} all | sed -n "/digraph/,\$p" > {output.dot}
```

to run `snakemake dag_filegraph.pdf` or `snakemake dag_rulegraph.pdf` and forget about the `params` or config setting.


I would also be fine to just create both rulegraph and filegraph on run `snakemake dag`.

tgilon:
Thanks @fneum for the review. Using only wildcards implies that we cannot have a default behaviour. That's why I suggest using `config` and a variable. I think the `dag` rule needs to be an easy to use rule as the `dag` is key to understanding the workflow (especially in the beginning).

```shell
# Not possible
snakemake dag

# Possible
snakemake resources/dag_rulegraph.pdf
snakemake resources/dag_filegraph.pdf
```

Also, as the `filegraph` may not be useful for everyone, I thought it would be a good idea to make it optional.

Would there be an implementation with wildcards and a default value?

fneum (approved):

PR #1575, opened 2025-03-13
Windows snakefile

Eric-Nitschke:
### Summary
Modify `download_wdpa` and `download_wdpa_marine` in `rules/retrieve.smk` to properly work on Windows.

### Reason
Previously, both rules would not work on Windows, since the shell commands `cp` and `unzip` don't exist on Windows.
Since the corresponding commands (`copy`/`xcopy`) did not work for me when testing, and the `tar` command (roughly `unzip`) is not available on older Windows versions, I switched to Powershell.

I am not fully proficient regarding shell or powershell, so please comment any improvements you would suggest!

## Checklist


fneum:
A slightly cleaner alternative would be using Python for this rule, e.g. `shutil`: 

https://docs.python.org/3/library/shutil.html

Eric-Nitschke:
@fneum great idea! Do you suggest using for Windows only or regardless of the operating system?

fneum:
regardless of OS -- we'd try to avoid special solutions for different OSs

Eric-Nitschke:
The commits worked for me (Windows).
A few notes:
- I had to assing a key to the input, since shutils.copy requires a path (or similar) as input and not InputFiles. Thus I had to select the specific (and only) input either by index or key. I thought assigning a key might be a bit more robust regarding any potential future changes.
- I had to import copy from shutil as shcopy, since copy (not shutil.copy) is imported in common.smk, which leads to issues. We could considere changing it there to `from copy import deepcopy`, since it is the only function used from copy.

fneum (approved):
Works great!

PR #1578, opened 2025-03-13
Bidding zones representation + custom busmap

FabianHofmann:
The following changes add an additional layer for bidding zones on the administrative shapes introduced in #1502. 

On top, a cluster mode `custom_busshapes` was added to support passing user-defined busshapes (instead of custom busmap) which is used to perform the clustering. 

Note that this is two distinct features. 

### TODO
- [ ] Crete does not have a substation according to OSM prebuilt. however there should be one, as well as the interconnection between the main land and the island (https://www.admie.gr/en/erga/erga-diasyndeseis/diasyndesi-tis-kritis-me-tin-peloponniso). This missing piece is crashing the party a bit, as create has its own bidding zone. 

@bobbyxng @tgilon if you already want to take a look feel free. 

## Checklist

### DIFF
--- a/config/test/config.clusters.yaml
+++ b/config/test/config.clusters.yaml
@@ -1,0 +1,10 @@
+# SPDX-FileCopyrightText: Contributors to PyPSA-Eur <https://github.com/pypsa/pypsa-eur>
+#
+# SPDX-License-Identifier: CC0-1.0
+
+
+
+clustering:
+  mode: administrative
+  administrative:
+    level: bz
--- a/scripts/base_network.py
+++ b/scripts/base_network.py
@@ -1438,6 +1438,7 @@ def build_admin_shapes(
         1: "level1",
         2: "level2",
         3: "level3",
+        "bz": "bidding_zone",
     }
 
     adm1_countries = ["BA", "MD", "UA", "XK"]
@@ -1450,6 +1451,7 @@ def build_admin_shapes(
 
     if clustering == "administrative":
         logger.info(f"Building bus regions at administrative level {level}")
+
         nuts3_regions["column"] = level_map[level]
 
         # Only keep the values whose keys are in countries
@@ -1485,14 +1487,8 @@ def build_admin_shapes(
         nuts3_regions["admin"] = nuts3_regions["country"]
 
     # Group by busmap
-    admin_shapes = nuts3_regions[["admin", "geometry"]]
-    admin_shapes = admin_shapes.groupby("admin")["geometry"].apply(
-        lambda x: x.union_all()
-    )
-    admin_shapes = gpd.GeoDataFrame(
-        admin_shapes, geometry="geometry", crs=nuts3_regions.crs
-    )
-    admin_shapes["country"] = admin_shapes.index.str[:2]
+    admin_shapes = nuts3_regions[["admin", "geometry", "country"]]
+    admin_shapes = admin_shapes.dissolve("admin")
 
     # Identify regions that do not contain buses and merge them with smallest
     # neighbouring area of the same parent region (NUTS3 -> NUTS2 -> NUTS1 -> country)
--- a/scripts/build_bidding_zones.py
+++ b/scripts/build_bidding_zones.py
@@ -1,0 +1,274 @@
+# SPDX-FileCopyrightText: Contributors to PyPSA-Eur <https://github.com/pypsa/pypsa-eur>
+#
+# SPDX-License-Identifier: MIT
+from functools import reduce
+from itertools import takewhile
+from operator import attrgetter
+
+import geopandas as gpd
+import numpy as np
+import pandas as pd
+from shapely.geometry import MultiPolygon
+
+
+def _simplify_polys(polys, minarea=0.1, tolerance=None, filterremote=True):
+    if isinstance(polys, MultiPolygon):
+        polys = sorted(polys.geoms, key=attrgetter("area"), reverse=True)
+        mainpoly = polys[0]
+        mainlength = np.sqrt(mainpoly.area / (2.0 * np.pi))
+        if mainpoly.area > minarea:
+            polys = MultiPolygon(
+                [
+                    p
+                    for p in takewhile(lambda p: p.area > minarea, polys)
+                    if not filterremote or (mainpoly.distance(p) < mainlength)
+                ]
+            )
+        else:
+            polys = mainpoly
+    if tolerance is not None:
+        polys = polys.simplify(tolerance=tolerance)
+    return polys
+
+
+def get_countries(naturalearth: str, country_list: set[str]) -> gpd.GeoDataFrame:
+    """
+    Load geometries for missing countries from Natural Earth dataset.
+
+    Args:
+        naturalearth: Path to Natural Earth dataset
+        missing_countries: Set of country codes that are missing in the bidding zones
+
+    Returns:
+        GeoDataFrame: Contains geometries for all missing countries
+    """
+    df = gpd.read_file(naturalearth)
+
+    # Names are a hassle in naturalearth, try several fields
+    fieldnames = (
+        df[x].where(lambda s: s != "-99") for x in ("ISO_A2", "WB_A2", "ADM0_A3")
+    )
+    df["name"] = reduce(lambda x, y: x.fillna(y), fieldnames, next(fieldnames)).str[:2]
+    df.replace({"name": {"KV": "XK"}}, inplace=True)
+
+    df = df.loc[
+        df.name.isin(country_list) & ((df["scalerank"] == 0) | (df["scalerank"] == 5))
+    ]
+    return gpd.GeoDataFrame(
+        df.set_index("name")[["geometry"]].geometry.map(_simplify_polys).set_crs(df.crs)
+    )
+
+
+def parse_zone_names(zone_names: pd.Series) -> tuple[set[str], pd.Series]:
+    """
+    Parse bidding zone names to extract country codes and handle combined zones.
+
+    Args:
+        zone_names: Series of zone names (e.g., 'DE', 'DE_LU', etc.)
+
+    Returns:
+        tuple containing:
+            - set of country codes covered by the zones
+            - Series of combined country strings (e.g., 'DE,LU')
+    """
+    # Extract primary country codes (first two characters)
+    primary_countries = set(zone_names.str[:2])
+
+    # Parse combined zones (e.g., 'DE_LU' -> 'LU')
+    secondary_countries = (
+        zone_names.str.split("_", expand=True)[1]
+        .dropna()
+        .loc[lambda x: x.str.len() == 2]
+    )
+
+    # Create combined country strings (e.g., 'DE,LU')
+    country_strings = zone_names.str[:2] + (
+        "," + secondary_countries.reindex(zone_names.index)
+    ).fillna("")
+
+    # Remove secondary countries from primary set
+    covered_countries = primary_countries | set(secondary_countries)
+
+    return covered_countries, country_strings
+
+
+def split_combined_zones(gdf, country_shapes):
+    """Split multi-country zones into individual country entries."""
+    crs = gdf.crs
+    new_rows = []
+
+    for _, row in gdf.iterrows():
+        countries = row.country.split(",")
+        if len(countries) == 1:
+            new_rows.append(row)
+            continue
+
+        # Split geometry for combined countries
+        for country in countries:
+            country_geom = country_shapes.loc[country].geometry
+            intersection = row.geometry.intersection(country_geom)
+
+            if not intersection.is_empty:
+                new_row = row.copy()
+                new_row.country = country
+                new_row.geometry = intersection
+                new_rows.append(new_row)
+
+    return gpd.GeoDataFrame(new_rows, crs=crs)
+
+
+def extract_shape_by_bbox(
+    gdf: gpd.GeoDataFrame,
+    country: str,
+    min_lon: float,
+    max_lon: float,
+    min_lat: float,
+    max_lat: float,
+    region_id: str,
+):
+    """
+    Extracts a shape from a country's GeoDataFrame based on latitude and longitude bounds.
+
+    Parameters
+    ----------
+        - gdf (GeoDataFrame): GeoDataFrame containing country geometries.
+        - country (str): The country code or name to filter.
+        - min_lon, max_lon (float): Longitude bounds for extraction.
+        - min_lat, max_lat (float): Latitude bounds for extraction.
+        - region_id (str): String to assign an ID to the extracted region.
+
+    Returns
+    -------
+        - gdf_new: Updated GeoDataFrame with the extracted shape separated.
+    """
+    country_gdf = gdf.explode().query(f"country == '{country}'").reset_index(drop=True)
+
+    extracted_region = country_gdf.cx[min_lon:max_lon, min_lat:max_lat].assign(
+        zone_name=region_id
+    )
+
+    remaining_country = (
+        country_gdf.drop(extracted_region.index).dissolve(by="country").reset_index()
+    )
+
+    return pd.concat(
+        [
+            gdf.query(f"country != '{country}'"),
+            remaining_country,
+            extracted_region.dissolve(by="country").reset_index(),
+        ]
+    ).reset_index(drop=True)
+
+
+def format_names(s: str):
+    s = (
+        s.replace("DK-DK1", "DKW1")
+        .replace("DK-DK2", "DKE1")
+        .replace("GB", "UK")
+        .replace("IT_NORD", "ITN1")
+        .replace("IT_SUD", "ITS1")
+        .replace("LU", "LUG1")
+        .replace("NO-NO3", "NOM1")
+        .replace("NO-NO4", "NON1")
+        .replace("SE-SE", "SE0")
+        .replace("_", "")
+        .ljust(4, "0")
+    )[:4]
+    return s
+
+
+if __name__ == "__main__":
+    if "snakemake" not in globals():
+        from _helpers import mock_snakemake
+
+        snakemake = mock_snakemake("build_bidding_zones")
+
+    # Load core bidding zones and country shapes
+    countries = snakemake.params.countries
+
+    rename_columns = {"zoneName": "zone_name", "countryKey": "country"}
+    bidding_zones_elecmaps = gpd.read_file(
+        snakemake.input.bidding_zones_electricitymaps
+    )
+    bidding_zones_elecmaps = bidding_zones_elecmaps.rename(columns=rename_columns)
+    bidding_zones = bidding_zones_elecmaps[
+        bidding_zones_elecmaps.country.isin(countries)
+    ].drop(columns="countryName")
+
+    if not set(countries).issubset(bidding_zones.country):
+        raise ValueError("Missing countries in electricitymaps bidding zones")
+
+    # Manual corrections: replace Italy by entsoepy version
+    bidding_zones = bidding_zones[bidding_zones.country != "IT"]
+
+    bidding_zones_entsoe = gpd.read_file(snakemake.input.bidding_zones_entsoepy)
+    bidding_zones_entsoe = bidding_zones_entsoe.rename(
+        columns={"zoneName": "zone_name"}
+    )
+    core_countries, country_strings = parse_zone_names(
+        bidding_zones_entsoe["zone_name"]
+    )
+    bidding_zones_entsoe["country"] = country_strings
+    italian_zones = bidding_zones_entsoe[bidding_zones_entsoe.country == "IT"]
+    bidding_zones = pd.concat([bidding_zones, italian_zones], ignore_index=True)
+    bidding_zones = bidding_zones.sort_values("zone_name").reset_index(drop=True)
+
+    # manual corrections: remove islands
+    islands = [
+        "DK-BHM",
+        "ES-CE",
+        "ES-CN-HI",
+        "ES-CN-IG",
+        "ES-CN-LP",
+        "ES-CN-LZ",
+        "ES-IB-FO",
+        "ES-IB-IZ",
+        "ES-IB-ME",
+        "ES-IB-MA",
+        "ES-CN-FV",
+        "ES-CN-GC",
+        "ES-CN-TE",
+        "ES-ML",
+        "GB-ORK",
+        "GB-ZET",
+        "PT-MA",
+        "PT-AC",
+    ]
+    bidding_zones = bidding_zones[~bidding_zones.zone_name.isin(islands)]
+
+    # manually merge southern norwegian zones
+    nos0_idx = bidding_zones.query("zone_name in ['NO-NO1', 'NO-NO2', 'NO-NO5']").index
+    bidding_zones = pd.concat(
+        [
+            bidding_zones.drop(nos0_idx),
+            bidding_zones.loc[nos0_idx]
+            .dissolve(by="country")
+            .reset_index()
+            .assign(zone_name="NOS0"),
+        ]
+    )
+
+    # Extract Crete
+    bidding_zones = extract_shape_by_bbox(
+        bidding_zones,
+        country="GR",
+        min_lon=24.0,
+        max_lon=26.5,
+        min_lat=35.0,
+        max_lat=35.7,
+        region_id="GR03",
+    )
+
+    # manually add zone group for DE_LU
+    bidding_zones.loc[
+        bidding_zones.zone_name.isin(["DE", "LU"]), "cross_country_zone"
+    ] = "DE_LU"
+
+    # if turkey is not in the list of countries, drop northern cyprus
+    if "TR" not in countries:
+        bidding_zones = bidding_zones[~bidding_zones.zone_name.eq("XX")]
+
+    # rename zones
+    bidding_zones["zone_name"] = bidding_zones["zone_name"].apply(format_names)
+
+    bidding_zones.to_file(snakemake.output.file)
--- a/scripts/build_shapes.py
+++ b/scripts/build_shapes.py
@@ -252,6 +252,7 @@ def create_regions(
     offshore_shapes,
     nuts3_gdp,
     nuts3_pop,
+    bidding_zones_path,
     other_gdp,
     other_pop,
 ):
@@ -294,7 +295,6 @@ def create_regions(
     regions = regions.rename(
         columns={"NUTS_ID": "id", "CNTR_CODE": "country", "NAME_LATN": "name"}
     )
-
     # Normalise text
     regions["id"] = regions["id"].apply(normalise_text)
 
@@ -340,6 +340,10 @@ def create_regions(
     regions["geometry"] = regions["geometry"].difference(
         offshore_shapes.geometry.union_all()
     )
+
+    # Add bidding zone information
+    bidding_zones = gpd.read_file(bidding_zones_path)
+    regions = map_bidding_zones_to_regions(regions, bidding_zones)
 
     # GDP and POP for NUTS3 regions
     # GDP
@@ -488,6 +492,53 @@ def calc_gdp_pop(country, regions, gdp_non_nuts3, pop_non_nuts3):
     return gdp_pop
 
 
+def map_bidding_zones_to_regions(
+    regions: gpd.GeoDataFrame, bidding_zones: gpd.GeoDataFrame
+) -> gpd.GeoDataFrame:
+    """
+    Map bidding zones to regions on a country-by-country basis.
+
+    Parameters
+    ----------
+    regions : geopandas.GeoDataFrame
+        The regions GeoDataFrame with a 'country' column
+    bidding_zones : geopandas.GeoDataFrame
+        The bidding zones GeoDataFrame with a 'zone_name' and "country" column.
+
+    Returns
+    -------
+    regions : geopandas.GeoDataFrame
+        The regions GeoDataFrame with an added 'bidding_zone' columnk
+    """
+
+    # Initialize the bidding_zone column with NaN values
+    regions["bidding_zone"] = pd.NA
+
+    # Get unique countries in regions
+    countries = regions["country"].unique()
+
+    # Process each country separately
+    for country in countries:
+        # Get regions for this country
+
+        country_sel = regions["country"] == country
+        country_regions = regions[country_sel].copy()
+        # Get bidding zones that might correspond to this country
+        country_bidding_zones = bidding_zones[bidding_zones.country == country]
+
+        # Perform spatial join for this country
+        if country_regions.empty or country_bidding_zones.empty:
+            continue
+        joined = gpd.sjoin(
+            country_regions, country_bidding_zones, how="left", predicate="within"
+        )
+
+        # Update the bidding_zone column for this country
+        regions.loc[country_sel, "bidding_zone"] = joined["zone_name"]
+
+    return regions
+
+
 if __name__ == "__main__":
     if "snakemake" not in globals():
         from _helpers import mock_snakemake
@@ -511,6 +562,7 @@ if __name__ == "__main__":
         offshore_shapes,
         snakemake.input.nuts3_gdp,
         snakemake.input.nuts3_pop,
+        snakemake.input.bidding_zones,
         snakemake.input.other_gdp,
         snakemake.input.other_pop,
     )
--- a/scripts/build_shapes.py
+++ b/scripts/build_shapes.py
@@ -341,10 +341,6 @@ def create_regions(
         offshore_shapes.geometry.union_all()
     )
 
-    # Add bidding zone information
-    bidding_zones = gpd.read_file(bidding_zones_path)
-    regions = map_bidding_zones_to_regions(regions, bidding_zones)
-
     # GDP and POP for NUTS3 regions
     # GDP
     logger.info(f"Importing JRC ARDECO GDP data for year {GDP_YEAR}.")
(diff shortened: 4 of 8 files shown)


tgilon (commented):
Thank you @FabianHofmann!  I added some comments. I haven't had the chance to review `cluster_network.py` yet.

scripts/build_bidding_zones.py
```diff
@@ -0,0 +1,195 @@
...
+
+    # Create combined country strings (e.g., 'DE,LU')
+    country_strings = zone_names.str[:2] + (
+        "," + secondary_countries.reindex(zone_names.index)
+    ).fillna("")
+
+    # Remove secondary countries from primary set
+    covered_countries = primary_countries | set(secondary_countries)
+
+    return covered_countries, country_strings
```
tgilon, line 38: This output could be removed as it is not used.

```suggestion
    return country_strings
```

scripts/build_bidding_zones.py
```diff
@@ -0,0 +1,195 @@
...
+    Args:
+        zone_names: Series of zone names (e.g., 'DE', 'DE_LU', etc.)
+
+    Returns:
+        tuple containing:
+            - set of country codes covered by the zones
+            - Series of combined country strings (e.g., 'DE,LU')
+    """
+    # Extract primary country codes (first two characters)
+    primary_countries = set(zone_names.str[:2])
```
tgilon, line 21: Not used if `covered_countries` is dropped
```suggestion
```

config/test/config.clusters.yaml
```diff
@@ -0,0 +1,10 @@
+# SPDX-FileCopyrightText: Contributors to PyPSA-Eur <https://github.com/pypsa/pypsa-eur>
+#
+# SPDX-License-Identifier: CC0-1.0
+
+
```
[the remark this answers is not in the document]
FabianHofmann: The problem comes in build_electricity demand. there is no synthetic demand data for these two. but don't know why actually...

FabianHofmann:
great @tgilon that you could track it down!

FabianHofmann:
@tgilon would you mind adding a release note and checking that configtables documentation (in doc/configtables/) is in line?

tgilon:
> @tgilon would you mind adding a release note and checking that configtables documentation (in doc/configtables/) is in line?

Done, but not pushed yet!

@FabianHofmann I haven't tested the `custom_busshapes` feature yet. What do you think about keeping it separate from this PR?

FabianHofmann:
mmh, I would say let's keep it, it is quite convenient. @fneum @bobbyxng does one of you also want to review?

scripts/build_bidding_zones.py
```diff
@@ -0,0 +1,190 @@
...
+        zone_name=region_id
+    )
+
+    remaining_country = (
+        country_gdf.drop(extracted_region.index).dissolve(by="country").reset_index()
+    )
+
+    return pd.concat(
+        [
+            gdf.query(f"country != '{country}'"),
```
fneum, line 72:
```suggestion
            gdf.query("country != @country"),
```

scripts/build_bidding_zones.py
```diff
@@ -0,0 +1,190 @@
...
+    bidding_zones_entsoe = gpd.read_file(snakemake.input.bidding_zones_entsoepy)
+    bidding_zones_entsoe = bidding_zones_entsoe.rename(
+        columns={"zoneName": "zone_name"}
+    )
+    country_strings = parse_zone_names(bidding_zones_entsoe["zone_name"])
+    bidding_zones_entsoe["country"] = country_strings
+    italian_zones = bidding_zones_entsoe[bidding_zones_entsoe.country == "IT"]
+    bidding_zones = pd.concat([bidding_zones, italian_zones], ignore_index=True)
+
+    # manual corrections: remove islands
```
fneum, line 131: is it really correct to always remove these islands? or should this be configurable?

scripts/build_bidding_zones.py
```diff
@@ -0,0 +1,190 @@
...
+        "ES-CN-TE",
+        "ES-ML",
+        "GB-ORK",
+        "GB-ZET",
+        "PT-MA",
+        "PT-AC",
+    ]
+    bidding_zones = bidding_zones[~bidding_zones.zone_name.isin(islands)]
+
+    # manually merge southern norwegian zones
```
fneum, line 154: why is this necessary? wouldn't it be better to keep all three bidding zones?

scripts/cluster_network.py
```diff
@@ -504,7 +638,25 @@ def update_bus_coordinates(
...
             )
+        elif mode == "custom_busshapes":
+            n.determine_network_topology()
+            custom_shapes = gpd.read_file(snakemake.input.custom_busshapes)
+            use_shapes_index = True
+            if {"country", "id"} & set(custom_shapes.columns):
+                # keep pypsa-eur naming convention and add an subnetwork id "0" after the country
+                index = custom_shapes.country + "0 " + custom_shapes.id.astype(str)
+                custom_shapes.index = index
+                use_shapes_index = True
```
fneum, line 649: I think it's fine to deviate from the "naming convention". The NUTS clustering also does not adhere to it and all code should not depend on this anyways.

scripts/cluster_network.py
```diff
@@ -106,6 +106,148 @@ def weighting_for_country(df: pd.DataFrame, weights: pd.Series) -> pd.Series:
...
+    column = cluster_names or "name"
+    busmap = gpd.sjoin(coords, shapes, how="left")[column].rename("busmap")
+
+    if busmap.isnull().any():
+        unassigned = coords[busmap.isnull()]
+        # Take a projection which properly handles distances for European areas.
+        with warnings.catch_warnings():
+            warnings.simplefilter("ignore", category=DeprecationWarning)
+            unassigned_converted = unassigned.to_crs(3035)
+        shapes_converted = shapes.to_crs(3035).set_index(column)
```
fneum, line 191: why not address the deprecation?

scripts/cluster_network.py
```diff
@@ -106,6 +106,148 @@ def weighting_for_country(df: pd.DataFrame, weights: pd.Series) -> pd.Series:
...
+                    "Link",
+                    f"copper_{zone}_{i}",
+                    carrier="copper",
+                    bus0=first_bus,
+                    bus1=bus,
+                    p_nom=float("inf"),
+                    p_min_pu=-1,
+                    underwater_fraction=0.0,
+                    under_construction=0.0,
+                )
```
fneum, line 248: Doesn't that create many unnecessary links? Given that they will never be binding, they could also generate more degeneracies. Why not map all buses to bidding zone buses for all components, a bit like in https://pypsa.readthedocs.io/en/latest/examples/scigrid-redispatch.html#Build-market-model-m-with-two-bidding-zones?

```py
for c in m.iterate_components(m.one_port_components):
    c.static.bus = c.static.bus.map(zones)

for c in m.iterate_components(m.branch_components):
    c.static.bus0 = c.static.bus0.map(zones)
    c.static.bus1 = c.static.bus1.map(zones)
    internal = c.static.bus0 == c.static.bus1
    m.remove(c.name, c.static.loc[internal].index)
```

(and remove all original buses)

This would also be, in principle, identical to the regular clustering approach.

scripts/cluster_network.py
```diff
@@ -545,6 +698,13 @@ def update_bus_coordinates(
 
     nc = clustering.n
 
+    if mode == "administrative" and "bz" in params.administrative.values():
+        bidding_zones = gpd.read_file(snakemake.input.bidding_zones)
+        zone_name = bidding_zones.set_index("zone_name").cross_country_zone
+        zone_name = zone_name.where(zone_name.notnull(), zone_name.index)
+        nc.buses["zone"] = zone_name
+        copperplate_buses(nc, zone_name)
```
fneum, line 706: As far as I understood, this is for the case that you want one country with bidding zone representation and others in NUTSX, right? I didn't understand why this is not handled directly when creating the `busmap`. Wouldn't the regular handling of the `busmap` variable do everything you need?

scripts/build_shapes.py
```diff
@@ -488,6 +493,120 @@ def calc_gdp_pop(country, regions, gdp_non_nuts3, pop_non_nuts3):
     return gdp_pop
 
 
+def bidding_zone_map(
+    regions: gpd.GeoDataFrame, bidding_zones: gpd.GeoDataFrame
+) -> pd.Series:
+    """
+    Map bidding zones to regions on a country-by-country basis, assigning each region
+    to the bidding zone with which it has the largest overlap. If a region doesn't
+    overlap with any bidding zone, it's assigned to the nearest one as a fallback.
```
fneum, line 502: This function can be simplified to a simple `sjoin` on the substation coordinates. Similar to the NUTS clustering, we don't want to use the (aggregated) Voronoi shapes for anything. The regions associated with the clustered buses should match the original bidding zone shapes.

fneum (changes requested):
Looks mostly good,

Some documentation adjustments for clustering settings and release notes are yet to be included.

Main points:
- Infinite links in `copperplate_buses`
- Bidding zone mapping based on points rather than Voronoi overlap
- Optionality of `build_bidding_zones`

I'll give it a spin now.

scripts/base_network.py
```diff
@@ -1295,14 +1295,18 @@ def get_nearest_neighbour(
...
+            gdf.to_crs(epsg=3035)
+            .sjoin_nearest(
+                nearest_neighbours.to_crs(epsg=3035),
+                how="left",
+            )["index_right"]
+            .values[0]
+        )
+    except KeyError:
+        logger.warning(f"Skipping for {country}")
+        nearest_neighbour = ""
```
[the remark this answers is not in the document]
tgilon: They will be removed later [here](https://github.com/PyPSA/pypsa-eur/pull/1578/files#diff-283fbe3659c200bd33a2ea947f085dcf42f56f6d40fd016228c4ac1a59943717R1556) (e.g.: Crete).

scripts/base_network.py
```diff
@@ -1552,7 +1552,8 @@ def build_admin_shapes(
             lambda row: [get_nearest_neighbour(row, admin_shapes)],
             axis=1,
         )
-
+        # remove detached regions that don't contain substations
+        admin_shapes = admin_shapes[~admin_shapes.isempty]
```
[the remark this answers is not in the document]
tgilon: See comment above

scripts/build_bidding_zones.py
tgilon: The [Scenario Methodology Report](https://2024.entsos-tyndp-scenarios.eu/download/) of TYNDP 2024 merges them together. 

image

fneum: Hmm, let's make it optional then, with the default of not aggregating any actual bidding zones (e.g. a setting `aggregate_to_entsoe_scenario...`).

scripts/cluster_network.py
FabianHofmann: this unfortunately does not work well with the rest of the workflow. there are loads of functions that rely on a clear country assignment of components (power plants, energy totals, etc.). If we allowed for buses that don't represent a single country but multiple, most of these functions would fails due to missing reference. keeping the countries distinct and copperplating them (we want to make this optional) is the smoothest options as far as I can see.

FabianHofmann (commented):
@fneum thanks for the review as always. @tgilon will take over most parts, I will generalize the copperplate functionality which is atm very restricted and tight to the bidding zone representation.

scripts/cluster_network.py
fneum: Ah, so DE_LU is the only cross-country bidding zone. There would be one non-binding infinite link between the two, but the rest of the buses in Germany are clustered to one electricity bus for Germany (and same for Luxemburg)? That's fine! I just wouldn't do it for all buses in Germany/Luxemburg.

fneum:
@bobbyxng, could you look into the substation for Crete? The network on Crete is on 150 kV but with a 500kV DC link connection, so there should be one bus, right?

@tgilon, it would be great if it were configurable which bidding zones can be aggregated/omitted (e.g., the option to keep Balearic Islands, Bornholm, and South Norwegian bidding zones separate).

Small stuff:

1. It would be great if ITN1 were snapped to the boundaries of the surrounding bidding zones (i.e. no gaps/overlaps; discrepancy due to different data origin)

https://medium.com/@brendan_ward/how-to-leverage-geopandas-for-faster-snapping-of-points-to-lines-6113c94e59aa

https://geopandas.org/en/latest/docs/reference/api/geopandas.GeoSeries.snap.html

2. Maybe fill the two holes in SE03 (due to lakes)

3. IE00 + UKNI (Northern Ireland) would be another example of a cross_country_zone, similar to DE-LU. They operate as an "all-island" market zone.

https://www.epexspot.com/en/news/single-day-ahead-coupling-sdac-and-gb-parties-are-preparing-anticipated-impacts-brexit?utm_source=chatgpt.com

scripts/cluster_network.py
FabianHofmann: yes, exactly! it only does it for the electrical country buses (there is only one bus for DE and LU)

fneum:
I discussed with @bobbyxng to add a manual bus for Crete.

scripts/build_bidding_zones.py
tgilon: I'll add an option to remove islands (default `false`).

scripts/build_bidding_zones.py
tgilon: I'll add an option to aggregate to TYNDP topology (default `false`).

scripts/cluster_network.py
tgilon: It will be removed and we'll assume that there will be a "name" column.

scripts/cluster_network.py
tgilon: I'll make this option available for all the clustering modes. It's not handled in the `busmap` because we need at least one node per country for the following steps of the workflow. This option enables it.

tgilon: And to answer the question, it's not specifically for mixed topologies. The main use will be to copperplate nodes to model market zones.

scripts/build_shapes.py
tgilon: The idea discussed with @bobbyxng and @FabianHofmann  was to "just" add the bidding zones as another aggregation level. This is what's added in `build_shapes`, so that it can be processed by `base_network` like any other aggregation level. This leads to this edge case of border alignments, that this function solves.

tgilon:
Thank you very much @fneum and @FabianHofmann for the reviews! I've addressed all the comments and the PR is ready for another round of review. Here is a summary of the changes:

- I've added an explicit **option to remove islands**. This excludes the Balearic Islands, Bornholm, the Canary Islands, the Orkney Islands, the Shetland Islands, the Azores Islands and Madeira from the shape file. With OSM data and `bz` mode, you get only one node in Mallorca.
  image

- I've added an explicit **option to aggregate the TYNDP levels**. This aggregates the southern Norwegian bidding zones and extracts Crete as a separate zone from the Greek shape.

- I've **snapped the borders of ITN1** to the neighbouring countries. To be honest, it's a best effort and there are still minor discrepancies. 
  Before
  Screenshot from 2025-04-14 11-47-12
  After
  Screenshot from 2025-04-14 11-46-23

- I've removed the holes in SE03.
  image

- I've refactored `busmap_from_shapes` to strictly use columns as names for the shapes (using `name` column by default).

- I've generalised the `copperplate_buses` function. It now supports an explicit option to list buses to copperplate.

FabianHofmann:
@fneum this is quite mature now. anything speaks against merging it?

fneum:
I will try to re-review next week. Got some catching up to do after holiday.

fneum:
## Test Configuration 2

```yaml
run:
  name: "bidding-zone-2"

scenario:
  clusters:
  - adm
  planning_horizons:
  - 2050

clustering:
  mode: administrative
  administrative:
    level: bz
  copperplate_regions:
  - [DE00, LUG1]
  - [UKNI, IE00]
  build_bidding_zones:
    remove_islands: false
    aggregate_to_tyndp: false
  simplify_network:
    remove_stubs: true
    remove_stubs_across_borders: true
```

Mostly has the same shapefile problems.

image

fneum:
## Test Configuration 3

```yaml
run:
  name: "bidding-zone-3"

snapshots:
  start: "2013-01-01"
  end: "2013-01-08"
  inclusive: 'left'

scenario:
  clusters:
  - adm
  planning_horizons:
  - 2050

clustering:
  mode: administrative
  administrative:
    level: bz
  copperplate_regions:
  - [DE00, LUG1]
  - [UKNI, IE00]
  build_bidding_zones:
    remove_islands: false
    aggregate_to_tyndp: false
  simplify_network:
    remove_stubs: true
    remove_stubs_across_borders: false
```

This one fails -- but I think it is necessary that this setting works as it preserves the country boundaries (`remove_stubs_across_borders: false`).

<details><summary>Details</summary>
<p>

[Mon Apr 28 17:41:04 2025]
localrule cluster_network:
    input: resources/bidding-zone-3/networks/base_s.nc, resources/bidding-zone-3/admin_shapes.geojson, resources/bidding-zone-3/bidding_zones.geojson, resources/bidding-zone-3/regions_onshore_base_s.geojson, resources/bidding-zone-3/regions_offshore_base_s.geojson, resources/bidding-zone-3/electricity_demand_base_s.nc
    output: resources/bidding-zone-3/networks/base_s_adm.nc, resources/bidding-zone-3/regions_onshore_base_s_adm.geojson, resources/bidding-zone-3/regions_offshore_base_s_adm.geojson, resources/bidding-zone-3/busmap_base_s_adm.csv, resources/bidding-zone-3/linemap_base_s_adm.csv
    log: logs/bidding-zone-3/cluster_network_base_s_adm.log
    jobid: 1
    benchmark: benchmarks/bidding-zone-3/cluster_network_base_s_adm
    reason: Missing output files: resources/bidding-zone-3/networks/base_s_adm.nc; Input files updated by another job: resources/bidding-zone-3/networks/base_s.nc, resources/bidding-zone-3/electricity_demand_base_s.nc, resources/bidding-zone-3/admin_shapes.geojson, resources/bidding-zone-3/regions_onshore_base_s.geojson, resources/bidding-zone-3/bidding_zones.geojson, resources/bidding-zone-3/regions_offshore_base_s.geojson
    wildcards: clusters=adm
    resources: tmpdir=/tmp, mem_mb=10000, mem_mib=9537

Waiting for more resources.
INFO:pypsa.io:Imported network base_s.nc has buses, carriers, lines, links, shapes
INFO:__main__:Clustering at administrative level bz.
INFO:__main__:Note that the following countries can only be clustered at a maximum administration level of 1: ['XK', 'BA'].
INFO:__main__:Mapping buses to administrative regions.
100%|█████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████| 34/34 [00:00<00:00, 76.47it/s]
INFO:__main__:Updating x, y coordinates of buses based on administrative regions.
ERROR:root:Uncaught exception
Traceback (most recent call last):
  File "/home/fneum/pr/pypsa-eur/.snakemake/scripts/tmplei9i6tn.cluster_network.py", line 687, in <module>
    clustering = clustering_for_n_clusters(
                 ^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/fneum/pr/pypsa-eur/.snakemake/scripts/tmplei9i6tn.cluster_network.py", line 421, in clustering_for_n_clusters
    clustering = get_clustering_from_busmap(
                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/fneum/miniconda3/envs/pypsa-eur-250330/lib/python3.12/site-packages/pypsa/clustering/spatial.py", line 489, in get_clustering_from_busmap
    buses = aggregatebuses(n, busmap, custom_strategies=bus_strategies)
            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/fneum/miniconda3/envs/pypsa-eur-250330/lib/python3.12/site-packages/pypsa/clustering/spatial.py", line 323, in aggregatebuses
    aggregated = n.buses.groupby(busmap).agg(strategies)
                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/fneum/miniconda3/envs/pypsa-eur-250330/lib/python3.12/site-packages/pandas/core/groupby/generic.py", line 1432, in aggregate
    result = op.agg()
             ^^^^^^^^
  File "/home/fneum/miniconda3/envs/pypsa-eur-250330/lib/python3.12/site-packages/pandas/core/apply.py", line 190, in agg
    return self.agg_dict_like()
           ^^^^^^^^^^^^^^^^^^^^
  File "/home/fneum/miniconda3/envs/pypsa-eur-250330/lib/python3.12/site-packages/pandas/core/apply.py", line 423, in agg_dict_like
    return self.agg_or_apply_dict_like(op_name="agg")
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/fneum/miniconda3/envs/pypsa-eur-250330/lib/python3.12/site-packages/pandas/core/apply.py", line 1608, in agg_or_apply_dict_like
    result_index, result_data = self.compute_dict_like(
                                ^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/fneum/miniconda3/envs/pypsa-eur-250330/lib/python3.12/site-packages/pandas/core/apply.py", line 497, in compute_dict_like
    getattr(obj._gotitem(key, ndim=1), op_name)(how, **kwargs)
  File "/home/fneum/miniconda3/envs/pypsa-eur-250330/lib/python3.12/site-packages/pandas/core/groupby/generic.py", line 294, in aggregate
    return self._python_agg_general(func, *args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/fneum/miniconda3/envs/pypsa-eur-250330/lib/python3.12/site-packages/pandas/core/groupby/generic.py", line 327, in _python_agg_general
    result = self._grouper.agg_series(obj, f)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/fneum/miniconda3/envs/pypsa-eur-250330/lib/python3.12/site-packages/pandas/core/groupby/ops.py", line 864, in agg_series
    result = self._aggregate_series_pure_python(obj, func)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/fneum/miniconda3/envs/pypsa-eur-250330/lib/python3.12/site-packages/pandas/core/groupby/ops.py", line 885, in _aggregate_series_pure_python
    res = func(group)
          ^^^^^^^^^^^
  File "/home/fneum/miniconda3/envs/pypsa-eur-250330/lib/python3.12/site-packages/pandas/core/groupby/generic.py", line 324, in <lambda>
    f = lambda x: func(x, *args, **kwargs)
                  ^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/fneum/miniconda3/envs/pypsa-eur-250330/lib/python3.12/site-packages/pypsa/clustering/spatial.py", line 133, in consense
    assert (x == v).all() or x.isnull().all(), (
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
AssertionError: In Bus cluster carrier, the values of attribute carrier do not agree:
Bus
SE21-400                 AC
SE22-220                 AC
SE24-400                 AC
SE28-400                 AC
SE32-220                 AC
SE35-420                 AC
SE36-400                 AC
SE38-400                 AC
SE43-400                 AC
relation/10153505-400    AC
relation/11097406-420    AC
relation/11334274-400    AC
relation/11964241-400    AC
relation/6081310-400     AC
relation/7108546-220     AC
way/1022235514-400       AC
way/1065701571-400       AC
way/106607138-220        AC
way/1265541077-400       AC
way/138622087-220        AC
way/140248154            DC
way/144381726-220        AC
way/160474491-400        AC
way/168056174-220        AC
way/168351799-400        AC
way/168908158-220        AC
way/168935344-220        AC
way/200052139-400        AC
way/207159887-400        AC
way/208965529-400        AC
way/209175136-400        AC
way/23470015-400         AC
way/24415564-220         AC
way/24534144-400         AC
way/255570234-220        AC
way/257920140-400        AC
way/260986936-400        AC
way/261447466-400        AC
way/263697062-400        AC
way/263701163-400        AC
way/265416362-400        AC
way/26755070-220         AC
way/281023714-400        AC
way/281390335-400        AC
way/28276448-400         AC
way/284182193-400        AC
way/294798805-220        AC
way/294897999-400        AC
way/38105008-400         AC
way/44636668-400         AC
way/51252509-400         AC
way/5174038-400          AC
way/57675923-400         AC
way/80702333-400         AC
way/934908818-400        AC
Name: carrier, dtype: object
RuleException:
CalledProcessError in file "/home/fneum/pr/pypsa-eur/rules/build_electricity.smk", line 690:
Command 'set -euo pipefail;  /home/fneum/miniconda3/envs/pypsa-eur-250330/bin/python3.12 /home/fneum/pr/pypsa-eur/.snakemake/scripts/tmplei9i6tn.cluster_network.py' returned non-zero exit status 1.

</p>
</details>

fneum (changes requested):
Thanks for all the revisions! I agree with almost everything, but I have a few things to fine-tune:

I've gone through 3 test cases, which should work in my view. The first two have some issues with the shapes (especially in Italy), which I think are worth revisiting before merging. The third case fails (mixed AC/DC buses). It would be good if @tgilon could trace this error and provide a fix.

The other aspect I was previously concerned about was the calculation of overlaps between bidding zones and NUTS3 regions -- which created some fuzziness around the borders of the resulting bidding zones / network aggregation. I understand now that this is a good compromise solution if you want to mix NUTS resolutions for some countries and bidding zones for others. I am only unsure how frequently this use case would be encountered.

A cleaner solution I have been thinking about is implementing the bidding zone clustering without involvement of the "base" administrative regions by introducing another clustering approach:

```yaml
clustering:
  mode: bidding_zones
```

This would be more similar to the custom busshapes approach, simply capturing bus coordinates into bidding zone shapefiles through an `.sjoin` operation. This would avoid the fuzziness.

What do you think of this alternative? I would just like to understand if/why this might be a worse alternative.

FabianHofmann:
> Thanks for all the revisions! I agree with almost everything, but I have a few things to fine-tune:
> 
> I've gone through 3 test cases, which should work in my view. The first two have some issues with the shapes (especially in Italy), which I think are worth revisiting before merging. The third case fails (mixed AC/DC buses). It would be good if @tgilon could trace this error and provide a fix.
> 
> The other aspect I was previously concerned about was the calculation of overlaps between bidding zones and NUTS3 regions -- which created some fuzziness around the borders of the resulting bidding zones / network aggregation. I understand now that this is a good compromise solution if you want to mix NUTS resolutions for some countries and bidding zones for others. I am only unsure how frequently this use case would be encountered.
> 
> A cleaner solution I have been thinking about is implementing the bidding zone clustering without involvement of the "base" administrative regions by introducing another clustering approach:
> 
> ```yaml
> clustering:
>   mode: bidding_zones
> ```
> 
> This would be more similar to the custom busshapes approach, simply capturing bus coordinates into bidding zone shapefiles through an `.sjoin` operation. This would avoid the fuzziness.
> 
> What do you think of this alternative? I would just like to understand if/why this might be a worse alternative.

Just in short, @tgilon might complement this:

1. I think the italian is quite hard, I think @tgilon mentioned this is as good as you can get without extensive tedious manual work
2. AC/DC buses I don't know, reappears from time to time it feels, perhaps we need a more robust general solution for this?
3. what you are raising about the alternative clustering method. this is now supported via the custom busshapes functionality. however, we decided to not use this in the end (this was the initial version) as the voronoi cells which build the basis of the clustering routine do not match the final bidding zone shapes. By using the admin shapes as a basis for the clustering, we resolve this by leveraging @bobbyxng's restructuring routine where voronoi cells are brought inline with the admin shapes.

tgilon:
> I've gone through 3 test cases, which should work in my view. The first two have some issues with the shapes (especially in Italy), which I think are worth revisiting before merging. The third case fails (mixed AC/DC buses). It would be good if @tgilon could trace this error and provide a fix.

Regarding the Italian case, I managed to reproduce a similar issue using a NUTS2 config.

```yaml
run:
  name: "bidding-zone-4"

scenario:
  clusters:
  - adm
  planning_horizons:
  - 2050

clustering:
  mode: administrative
  administrative:
    level: 2
  copperplate_regions:
  - [DE00, LUG1]
  - [UKNI, IE00]
  build_bidding_zones:
    remove_islands: true
    aggregate_to_tyndp: true
  simplify_network:
    remove_stubs: true
    remove_stubs_across_borders: true
```

In this case, the same strange shape appears in Italy. This can be explained by examining in the base network that we are using.

Reference using your test config 1
image

Clustering using NUTS2
image

Looking at the base network, we can see that there is a line running from Firenze to the south of Volterra. There is also a line running from Prato to the North of Tarquinia. These lines cross. Removing the stubs, we link the Voltera zone with the Arezzo zone (including Firenze). 
image.

The same thing happens in southern Italy.
image

One could think of solving this with `remove_stubs: false` but this would take us back to the AC/DC mix (https://github.com/PyPSA/pypsa-eur/pull/1578/#issuecomment-2761177721).

tgilon:
I would recommend setting `remove_stubs_across_borders` to `false` by default (and in the test config). This solves our issues. 

Here is the associated shapefile (regions_onshore_base_s_adm.geojson):
image

However, a different approach is required for https://github.com/PyPSA/pypsa-eur/pull/1578/#issuecomment-2761177721.

I suggest changing the OSM data. Would it be possible to have a AC bus @bobbyxng on the island ?
image

This is the [raw grid](https://openinframap.org/#7.56/57.441/18.211).
image

Another option would be to force the carrier for this bus using something like this:
```python
n.buses.loc["way/140248154", "carrier"] = "AC"
```
I'm not a big fan of this one.

bobbyxng:
> I would recommend setting `remove_stubs_across_borders` to `false` by default (and in the test config). This solves our issues.
> 
> Here is the associated shapefile (`regions_onshore_base_s_adm.geojson`): image
> 
> However, a different approach is required for #1578.
> 
> I suggest changing the OSM data. Would it be possible to have a AC bus @bobbyxng on the island ? image
> 
> This is the [raw grid](https://openinframap.org/#7.56/57.441/18.211). image
> 
> Another option would be to force the carrier for this bus using something like this:
> 
> ```python
> n.buses.loc["way/140248154", "carrier"] = "AC"
> ```
> 
> I'm not a big fan of this one. However, we could hardcode it while we wait for an OSM fix.

Hi Thomas! I will look into this. Sorry for the late reply, have been on conference/workshop trips the past two weeks.
Generally it is possible do do this. I checked on openinframap: The reason why the AC bus is missing is because a section between the converter to the 220 kV substation is missing. I can try to fix it on OSM directly, then the scripts should be handle it and we can update the OSM dataset.

tgilon:
> Sorry for the late reply, have been on conference/workshop trips the past two weeks. 

No problem, it always takes me a while to catch up with new reviews.

> I can try to fix it on OSM directly, then the scripts should be handle it and we can update the OSM dataset.

How much time do you need? Should I suggest a quick fix in the meantime (and merge the PR)? Or should we wait for your fix?

FabianHofmann:
I order to give this a boost (this has already taking quite some resources), I would suggest a quick fix for now with a dependency on the OSM version and try to get the fix into the next version of the OSM.

bobbyxng:
Regarding my previous comment: I checked and got this mixed up with Mallorca. Gotland on the AC side only has 110 kV. I don't know what the best approach would be here, keeping a 110 kV AC bus would be the easiest and most correct, just need to make sure that it does not cause any issues downstream. I am looking into it now and will propose a solution. Maybe give me until Friday latest so I can make sure everything is fine before pushing a new dataset to Zenodo :) Is there anything else that you want me to focus on in particular?

tgilon:
I applied the fix.

This is the final `base_network` using `config/test/config.clusters.yaml`.
image

And this is the associated `regions_onshore_base_s_adm.geojson`.
image

Enabling both `remove_islands` and `aggregate_to_tyndp` gives this.
image

image

bobbyxng (commented):
finished reviewing and testing the PR. 
Only open points form my side:
- "GB" vs. "UK" substring issue in `build_biomass_potentials`

tgilon:
>  "GB" vs. "UK" substring issue in `build_biomass_potentials`

Thanks @bobbyxng! I'm currently testing my fix. This has resulted in some other collateral damage.

tgilon:
@bobbyxng Thank you for the review, suggestion and proactivity! I've fixed the naming convention. This now means that we are no longer following the bidding zone naming convention of the TYNDP 2024. The only exceptions are `GB00` and `GBNI` (rather than `UK00` and `UKNI`).

To ensure everything runs smoothly, I applied same type of fix for the Shetland Islands bus as we did for Öland. Could you please fix this in the next OSM prebuilt?

The workflow runs entirely on my side using `config/test/config.clusters.yaml`. I had to include the following PR to make it work: https://github.com/PyPSA/pypsa-eur/pull/1695.

fneum (approved):
I'm fine with this version!

I would not have mixed the bidding zone clustering with the administrative shapes, and would have overwritten the aggregated Voronoi regions with the original bidding zone shapefiles after a plain `.sjoin` operation on the substations.

But this is also functional, and it's generally very useful to have a bidding zone level representation as an option.

- `remove_stubs_across_borders` should indeed be false.
- I have amended the test config with the renaming of UKNI to GBNI.

Ready to merge?

PR #1580, opened 2025-03-13
clustering with NUTS3 in Spain breaks

cristobal-GC:

I've been testing the new NUTS clustering feature (congratulations, it's a great update!). It works pretty well, but I get an error when running it only for Spain and for clustering at NUTS3 level.

The error occurs during the 'base_network' rule. 

In short, the problem seems to be that the "clean_dict" function is not robust if the provided dic has no values with len=1.

In detail: At some point, "build_admin_shapes" calls to "merge_regions_recursive" at:

https://github.com/PyPSA/pypsa-eur/blob/a3938febf80d6b93dea32afdb1586984b1093e3d/scripts/base_network.py#L1551 

In this function there is a call to the function "clean_dict" on line:

https://github.com/PyPSA/pypsa-eur/blob/a3938febf80d6b93dea32afdb1586984b1093e3d/scripts/base_network.py#L1387

where the input dictionary at this point is just: {'ES532': ['ES531', 'ES533']}

(just for info: ES531 and ES533 are two small islands without electrical buses, and here the code aims to add them to the larger island of Mallorca, ES532, which is connected with peninsular Spain with a DC-link). 

The problem is that "clean_dict" seems to break at line:

https://github.com/PyPSA/pypsa-eur/blob/a3938febf80d6b93dea32afdb1586984b1093e3d/scripts/base_network.py#L1250

because the variable "single_occurrences" is empty, due to the fact that there are no values in the input dictionary with length=1). This probably does not happen when running a case for Europe, but for the case of a single country like Spain, it leads to the mentioned break.

So, this PR is just to add a check step on "single_occurrences", to proceed only if it is not empty.

## Checklist

- [ x] I tested my contribution locally and it works as intended.

### DIFF
--- a/scripts/base_network.py
+++ b/scripts/base_network.py
@@ -1232,26 +1232,34 @@ def clean_dict(
 
     # Sub dictionary where values are only of length 1
     single_occurrences = {k: v for k, v in diction.items() if len(v) == 1}
-    # Create list of key, value pairs and store in list
-    single_values = list(chain.from_iterable(single_occurrences.values()))
-    single_keys = list(single_occurrences.keys())
+    
+    
+    # Enters only if single_occurrences is not empty
+    if single_occurrences:
+        # Create list of key, value pairs and store in list
+        single_values = list(chain.from_iterable(single_occurrences.values()))
+        single_keys = list(single_occurrences.keys())
 
-    # Create dict with single_keys() as keys and sets of single_keys and single_values as values using
-    single_df = pd.DataFrame(
-        zip(
-            single_keys,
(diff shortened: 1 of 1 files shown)


bobbyxng (approved):
Thank you @cristobal-GC for catching and fixing the bug, your fix looks good to me!
I only made some very minor edits:
- Fixed another small bug that is somewhat related to this: The "contains" column will only list neighbours once instead of multiple times
- Removed the redundant "else: drop_keys = ..." as drop_keys is never called outside the if condition. The first if condition on `single_occurrences` is enough :)

PR #1593, opened 2025-03-19
Bugfix: CO2 sequestration potential

bobbyxng:
- `build_sequestration_potentials.py`: CO2 sequestration input file for some reason contains many exact duplicates (same IDs, same geometries, same columns, etc.) --> Drop duplicates
- `prepare_sector_network.py`: max_size in `add_co2_tracking` function is multiplied by 1e3 to filter out unwanted sequestration sizes, however, the dataset's unit is consistently given in Mt. 1e3 factor needs to be removed. Finally, a single multiplication by 1e6 is needed only to convert to Mt to t (CO2 carrier unit in the model)
- Proposal to specify max_size in Mtpa (easier to interpret for a single modelling year than over a lifetime of e.g. 25 years)

## Checklist


fneum (changes requested):
The dropping of duplicates is fine (it should be added to #1606).

Everything else I would keep as is, as the min and max size should not depend on the annualisation factor.

fneum:
added dropping of duplicates here #1606

PR #1596, opened 2025-03-19
Use linopy model kwargs

### LINKED ISSUE #1594: Allow setting the directory where linopy saves files

It would be nice to set the directory where linopy temporarily persists files. By default, this is the /tmp directory which is well suited for almost all use cases. However, my deployment scenario requires me to change that directory to a NAS because of restricted storage space on the host file system.

pworschischek-aggmag:
Closes #1594.

Register linopy kwargs from config.yaml ulitimately passed to `pypsa.optimize.create_model()`. The proposed change will pass ```{"solver_dir": "/tmp"}``` by default instead of using the mutable default argument ```{}``` in the signature of the `create_model` function.
If the default setting ```solver_dir: "/tmp"``` causes problems on other platforms than Linux, the default should be removed from the config.default.yaml. I only tested on Linux. 

I also remove the out commented "/tmpdir" from the config.default.yaml because I found it was misleading. 

## Checklist


config/config.default.yaml
```diff
@@ -1025,6 +1024,10 @@ solving:
     atol: 1_000_000
     rtol: 0.01
 
+  model_kwargs:
+    solver_dir: "/tmp"
+  keep_files: false
```
fneum, line 1029: Let's move these into `solving.options` (where most other arguments to `n.optimize` are).

fneum (changes requested):
Good feature, some suggestions.

config/config.default.yaml
pworschischek-aggmag: perfect

fneum (approved):

PR #1602, opened 2025-03-20
Fix cutout directory on Windows

### LINKED ISSUE #1598: Snakemake can't find retrieve_cutouts
## Checklist

- [x] I am using the current [`master`](https://github.com/PyPSA/pypsa-eur/tree/master) branch or the latest [release](https://github.com/PyPSA/pypsa-eur/releases). Please indicate.

## Describe the Bug

This bug might be OS specific, and only occurs on Windows.

When trying to run PyPSA-Eur for the first time, snakemake could not identify `retrieve_cutouts` as a rule which provides `cutouts\europe-2013-sarah3-era5.nc`. The issue arises for other cutouts as well, such as the tutorial one `cutouts\be-03-2013-era5.nc`.

Further investigations identified the following issue:
- Through the changes introduced in PR #1515, CDIR is now detrermined by `CDIR = join(cutout_dir, ("" if run["shared_cutouts"] else RDIR))`, instead of directly being set to `"cutouts/"`.
- `join()` is `os.path.join()`.
- at least on Windows `join()` adds a `"\"` after joining, resulting in `"cutouts\"`. I assume that on Linux it adds `"/"`, resulting in the old behavior.
- I suspect that this leads to the behavior, where Snakemake thinks, that `"cutouts\europe-2013-sarah3-era5.nc"` can't be generated with `retrieve_cutouts`, which outputs `CDIR + "{cutout}.nc"`. This might be, because `"\e"` and `"\b"` are escape characters.

Setting `CDIR = "cutouts/"`  fixes the issue, but removes the improvements added by PR #1515. Maybe we could replace `join()` with something, that consistantly adds `"/"`.

## Error Message

```
Building DAG of jobs...
MissingInputException in rule determine_availability_matrix in file [...]\pypsa-eur\rules/build_electricity.smk, line 254:
Missing input files for rule determine_availability_matrix:
    output: resources/Test/[...]/availability_matrix_5_solar.nc
    wildcards: clusters=5, technology=solar
    affected files:
        cutouts\europe-2013-sarah3-era5.nc
```

Eric-Nitschke:
Bugfixes:
- change CDIR in Snakefile to always contain forward slashes instead of depending on the OS.

Closes #1598 

- change the definition of CDIR in Snakefile to not include `os.path.join` anymore. More details on this can be found in #1598.

## Checklist

### DIFF
--- a/rules/build_electricity.smk
+++ b/rules/build_electricity.smk
@@ -180,7 +180,9 @@ if config["enable"].get("build_cutout", False):
 rule build_ship_raster:
     input:
         ship_density="data/shipdensity_global.zip",
-        cutout=lambda w: CDIR.joinpath(config_provider("atlite", "default_cutout")(w) + ".nc").as_posix(),
+        cutout=lambda w: CDIR.joinpath(
+            config_provider("atlite", "default_cutout")(w) + ".nc"
+        ).as_posix(),
     output:
         resources("shipdensity_raster.tif"),
     log:
@@ -219,7 +221,9 @@ rule determine_availability_matrix_MD_UA:
             if w.technology in ("onwind", "solar", "solar-hsat")
             else resources("regions_offshore_base_s_{clusters}.geojson")
         ),
-        cutout=lambda w: CDIR.joinpath(config_provider("renewable", w.technology, "cutout")(w) + ".nc").as_posix(),
+        cutout=lambda w: CDIR.joinpath(
+            config_provider("renewable", w.technology, "cutout")(w) + ".nc"
+        ).as_posix(),
     output:
         availability_matrix=resources(
             "availability_matrix_MD-UA_{clusters}_{technology}.nc"
@@ -287,7 +291,9 @@ rule determine_availability_matrix:
             if w.technology in ("onwind", "solar", "solar-hsat")
             else resources("regions_offshore_base_s_{clusters}.geojson")
         ),
-        cutout=lambda w: CDIR.joinpath(config_provider("renewable", w.technology, "cutout")(w) + ".nc").as_posix(),
+        cutout=lambda w: CDIR.joinpath(
+            config_provider("renewable", w.technology, "cutout")(w) + ".nc"
+        ).as_posix(),
     output:
         resources("availability_matrix_{clusters}_{technology}.nc"),
     log:
@@ -312,7 +318,9 @@ rule build_renewable_profiles:
         availability_matrix=resources("availability_matrix_{clusters}_{technology}.nc"),
         offshore_shapes=resources("offshore_shapes.geojson"),
         regions=resources("regions_onshore_base_s_{clusters}.geojson"),
-        cutout=lambda w: CDIR.joinpath(config_provider("renewable", w.technology, "cutout")(w) + ".nc").as_posix(),
+        cutout=lambda w: CDIR.joinpath(
+            config_provider("renewable", w.technology, "cutout")(w) + ".nc"
+        ).as_posix(),
     output:
         profile=resources("profile_{clusters}_{technology}.nc"),
     log:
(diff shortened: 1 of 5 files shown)


lkstrp (commented):
Did you also try `pathlib` here? It is generally better for cross-platform compatibility. And we wanna leave those string paths behind.

> I was under the impression that Snakemake works with "Linux-Style" paths regardless of OS

Not sure how this interferes with Snakemake. But I don't like to rely on it

fneum:
I agree with @lkstrp — try `pathlib`. Also, the `CDIR` changes proposed behave differently regarding "/" suffixes.  Make sure this is the same as before.

Eric-Nitschke:
> Changing it to
> 
> ```
> CDIR = PurePosixPath(cutout_dir).joinpath("" if run["shared_cutouts"] else RDIR)
> ```
> 
> and changing every other occurance to
> 
> ```
> str(CDIR.joinpath("{cutout}.nc"))
> ```
> 
> (or similar), seemed to work in a dry-run.

It might be slighty cleaner to go with
```
CDIR = Path(cutout_dir, "" if run["shared_cutouts"] else RDIR)
```
and
```
CDIR.joinpath("{cutout}.nc").as_posix()
```
I'll try a full run with it today.

rules/build_electricity.smk
```diff
@@ -163,11 +163,11 @@ if config["enable"].get("build_cutout", False):
             regions_onshore=resources("regions_onshore.geojson"),
             regions_offshore=resources("regions_offshore.geojson"),
         output:
-            protected(CDIR + "{cutout}.nc"),
+            protected(CDIR.joinpath("{cutout}.nc").as_posix()),
```
fneum, line 166:
```suggestion
            protected(CDIR / "{cutout}.nc"),
```

Instead of `joinpath`, you should be able to use the` / ` operator (simplifies code)

`.as_posix()` should also not be necessary as

```

In [8]: from pathlib import Path

In [9]: Path("test") / "mypath"
Out[9]: PosixPath('test/mypath')

In [10]: Path("test").joinpath("mypath")
Out[10]: PosixPath('test/mypath')

In [11]: Path("") / Path("")
Out[11]: PosixPath('.')

In [12]: Path("")
Out[12]: PosixPath('.')

In [13]: "" / Path("test")
Out[13]: PosixPath('test')
```
This minimises code changes in `.smk` files to changing "+" to "/".

fneum (changes requested):
Here's a suggestion to simplify. Should hopefully also work on Windows.

coroa:
Sorry for dropping in here late, but i was also fighting this path nonsense with snakemake a while ago. Snakemake is in principle fine with `/` as a path separator or the corresponding backslash `\`, if they are used uniformly.

ie a string may not mix between them.

The way CDIR as a string is used and compounded with other paths you are introducing in so many places "/" into the string after the CDIR that this is a loosing battle and you should just use a literal "/" as a separator.

That works reliable because snakemake runs on all paths os.path.normpath and that converts `/` to `\` on windows.

At least that was my take-away.

Note that you can play around with an ipython on windows using wine:
(https://github.com/PietJankbal/powershell-wrapper-for-wine, and uv)

```
wget https://github.com/PietJankbal/powershell-wrapper-for-wine/raw/master/install_pwshwrapper.exe
wine install_pwshwrapper.exe
wine powershell
```
and then in the new powershell window:
```
powershell -ExecutionPolicy ByPass -c "irm https://astral.sh/uv/install.ps1 | iex"
```
and after reopening it:
```
uvx ipython
```

Have fun.

fneum (approved):
This version is acceptable to me! Thanks for your patience dealing with Windows paths! Iaf we find a better alternative later, we can still update. We just need Windows to run as this is 90% of first-time users.

PR #1612, opened 2025-03-25
Couple TES temperatures to district heating supply temperatures

cpschau:
This PR implicitly couples the top and bottom layer of PTES with the forward and return flow temperature of the local district heating network. Assuming a direct integration of the storage system, the bottom temperature equals the return flow temperature. The top temperature equals the forward flow temperature in the domain up to 90°C. At forward flow temperatures above 90°C the top temperature of the storage is kept at this operational limit of 90°C.

The variations of the network temperature lead to a dynamic storage capacity, that is described by the `e_max_pu` parameter/attribute of the Store component. According to Equation (1) in [Sorknaes (2018)](https://doi.org/10.1016/j.energy.2018.03.152), the dynamic storage capacity linearly depends on the deltaT between top and bottom temperature. The DEA catalogue indicates a deltaT of 55 K (top temperature of 90 and bottom temperature of 35 °C) as reference for the model. These reference values can be varied in the config settings `sector:district_heating:ptes:max_top_temperature` and `sector:district_heating:ptes:min_bottom_temperature` . As the spread between forward and return flow is smaller for a majority of systems, especially in summer, the effective storage capacity is smaller than the optimal value for the decision variable `e_nom_opt`.

2050, all countries (39 nodes), 8H resolution
image

The dynamic variation of the pit storage capacity can be turned off by setting the config parameter `sector:district_heating:ptes:dynamic_capacity` to `false`. The comparison of the SOC between a storage with a static and a dynamic storage capacity from an aggregated European perspective shows, that in  the latter case the maximum SOC does never come close to the nominal value of the energetic capacity. It remains below 50%.

2050, all countries (39 nodes), 8H resolution
image

The changes in the storage operation are also characterized by a higher amount of cycles. Overall, the reduction in average PTES capacity makes it less attractive, leading to a reduction in PTES investments and an increase in total system costs.

|                |   System Costs [bn€/yr] |   PTES Capacity [TWh] |   PTES Charger Capacity [GW] |   PTES Cycles [1/yr] |
|:---------------|------------------------:|----------------------:|-----------------------------:|---------------------:|
| Static         |              668.998    |              136.687  |                     911.246  |              4.0134  |
| Dynamic        |              674.144    |               97.2345 |                     648.23   |              5.93258 |
| Difference [%] |                0.769166 |              -28.8633 |                     -28.8633 |             47.8193  |

## Checklist


amos-schledorn:
Great feature and very useful plots! Striking by how much the storage capacity drops in summer.
Not sure I understand the time series plots. What does `e_max_py (weighted)` on the secondary y-axis mean exactly?

Code-wise, I would suggest to
- move this to a separate rule that either returns flat ones or the dynamic `e_max_pu`
- make 90/35C config settings or `technology-data` entries
- rename the first `e_max_pu` in the if-clause to `delta_t` or similar (or just remove it and use `e_max_pu = (t_top - t_bottom) / (t_top_nom  - t_bottom_nom)`
- add the Sorknaes paper and other relevant sources as references

Could you also add release notes and update the docs?

cpschau:
> Great feature and very useful plots! Striking by how much the storage capacity drops in summer. Not sure I understand the time series plots. What does `e_max_py (weighted)` on the secondary y-axis mean exactly?
> 
> Code-wise, I would suggest to
> 
>     * move this to a separate rule that either returns flat ones or the dynamic `e_max_pu`
> 
>     * make 90/35C config settings or `technology-data` entries
> 
>     * rename the first `e_max_pu` in the if-clause to `delta_t` or similar (or just remove it and use `e_max_pu = (t_top - t_bottom) / (t_top_nom  - t_bottom_nom)`
> 
>     * add the Sorknaes paper and other relevant sources as references
> 
> 
> Could you also add release notes and update the docs?

I addressed all of your points @amos-schledorn. In the plot you asked about, the `e_max_pu` timeseries of the pit components are weighted by their optimal capacity values to aggregate them to a value between 0 and 1.

scripts/prepare_sector_network.py
```diff
@@ -2760,6 +2760,8 @@ def add_heat(
     cop_profiles_file: str,
     direct_heat_source_utilisation_profile_file: str,
     hourly_heat_demand_total_file: str,
+    forward_flow_temperature_file: str,
+    return_flow_temperature_file: str,
```
amos-schledorn, line 2764: We can remove the two new input arguments now, right?

scripts/prepare_sector_network.py
```diff
@@ -3172,18 +3174,31 @@ def add_heat(
...
                 )
-
                 n.links.loc[
                     nodes + f" {heat_system} water pits charger",
                     "energy to power ratio",
                 ] = energy_to_power_ratio_water_pit
 
+                if options["district_heating"]["ptes"]["dynamic_capacity"]:
+                    # Load pre-calculated e_max_pu profiles
+                    e_max_pu_file = snakemake.input.ptes_e_max_pu_profiles
```
amos-schledorn, line 3184: I think, we don't want `snakemake` inside `prepare_sector_network` functions anymore. Let's pass this as an argument to `add_heat`.

scripts/prepare_sector_network.py
```diff
@@ -5950,9 +5965,10 @@ def add_import_options(
 
         snakemake = mock_snakemake(
             "prepare_sector_network",
+            run="tes_dynamic_capacity",
             opts="",
-            clusters="10",
```
amos-schledorn, line 5954: were the `mock_snakemake` changes intended?

scripts/prepare_sector_network.py
```diff
@@ -6075,6 +6091,8 @@ def add_import_options(
             cop_profiles_file=snakemake.input.cop_profiles,
             direct_heat_source_utilisation_profile_file=snakemake.input.direct_heat_source_utilisation_profiles,
             hourly_heat_demand_total_file=snakemake.input.hourly_heat_demand_total,
+            forward_flow_temperature_file=snakemake.input.central_heating_forward_temperature_profiles,
+            return_flow_temperature_file=snakemake.input.central_heating_return_temperature_profiles,
```
amos-schledorn, line 6095: The function arguments can be removed here, too, I think.

scripts/add_brownfield.py
```diff
@@ -292,17 +292,50 @@ def update_heat_pump_efficiency(n: pypsa.Network, n_p: pypsa.Network, year: int)
...
+        n_p.stores.index.str.contains("water pits")
+    ]
+    # construct names of same-technology dynamic pit storage in the current iteration
+    corresponding_idx_this_iteration = dynamic_ptes_idx_previous_iteration.str[
+        :-4
+    ] + str(year)
+    # update pit storage capacity in previous iteration in-place to capacity in this iteration
+    n_p.stores_t.e_max_pu[dynamic_ptes_idx_previous_iteration] = n.stores_t.e_max_pu[
+        corresponding_idx_this_iteration
+    ].values
```
amos-schledorn, line 324: I might be splitting hairs here, but if the nominal PTES temperatures change in the future, this should be re-calculated, correct? (since the min. bottom and max. top temperatures would not be constant across `PTES-<this-iteration>` and `PTES-<previous-iteration>`)

amos-schledorn (changes requested):
Looks really good! I just have a few minor comments. Crazy by how much `p_max_pu` is reduced - we should look into validating DH supply temperatures.

scripts/prepare_sector_network.py
cpschau: Yes we can!

scripts/add_brownfield.py
cpschau: Yes, you are right. Maybe this can be tracked in a separate issue together with the heat pumps, where the same issue persists for the efficiency dynamics?

config/config.default.yaml
```diff
@@ -471,6 +471,10 @@ sector:
       upper_threshold_ambient_temperature: 10
       rolling_window_ambient_temperature: 72
       relative_annual_temperature_reduction: 0.01
+    ptes:
+      dynamic_capacity: true
+      max_top_temperature: 90 # °C
+      min_bottom_temperature: # °C
```
fneum, line 477: What should the default value be?

Also: no comments in `config.default.yaml`. All in `doc/configtables*` (and later in json-schema #1547).

doc/configtables/sector.csv
```diff
@@ -17,6 +17,10 @@ district_heating,--,,`prepare_sector_network.py <https://github.com/PyPSA/pypsa-
 -- -- upper_threshold_ambient_temperature,C,float, Assume `min_forward_temperature` if ambient temperature is above this threshold
 -- -- rolling_window_ambient_temperature, h, int, Rolling window size for averaging ambient temperature when approximating supply temperature
 -- -- relative_annual_temperature_reduction,, float, Relative annual reduction of district heating forward and return temperature - defaults to 0.01 (1%)
+-- ptes,,,
+-- -- dynamic_capacity,--,"{true, false}",Add option for dynamic temperature-dependent capacity of pit storage in district heating
```
fneum, line 21:
```suggestion
-- -- dynamic_capacity,--,"{true, false}",Add option for dynamic temperature-dependent energy capacity of pit storage in district heating

```

doc/release_notes.rst
```diff
@@ -11,6 +11,10 @@ Release Notes
 Upcoming Release
 ================
 
+* Added option to use dynamic capacity for pit storage using the `e_max_pu` attribute of the store component, which is calculated in the new rule `build_tes_capacity_profiles` and added to the network in `prepare_sector_network`.
```
fneum, line 14:
```suggestion
* Added option to use dynamic capacity for pit storage using the ``e_max_pu`` attribute of the store component, which is calculated in the new rule :mod:`build_tes_capacity_profiles` and added to the network in :mod:`prepare_sector_network`.
```
RST syntax requires double-"`" for code. Rules you can refer to with `:mod:`<script_name>`.

scripts/build_tes_capacity/run.py
fneum, line 1: Could you add this to `doc/sector.rst` similar to the other rules referenced there?

scripts/build_tes_capacity/run.py
```diff
@@ -0,0 +1,88 @@
...
+    t_forward = xr.open_dataarray(
+        snakemake.input.central_heating_forward_temperature_profiles
+    )
+    t_return = xr.open_dataarray(
+        snakemake.input.central_heating_return_temperature_profiles
+    )
+
+    # Define operational limits for PTES
+    max_top_temperature = snakemake.params.get("max_top_temperature", 90)  # °C
+    min_bottom_temperature = snakemake.params.get("min_bottom_temperature", 35)  # °C
```
fneum, line 67:
```suggestion
    max_top_temperature = snakemake.params["max_top_temperature"]
    min_bottom_temperature = snakemake.params["min_bottom_temperature"]
```

Defaults are already set in `config.default.yaml`.

scripts/add_brownfield.py
```diff
@@ -292,6 +292,38 @@ def update_heat_pump_efficiency(n: pypsa.Network, n_p: pypsa.Network, year: int)
     )
 
 
+def update_dynamic_ptes_capacity(n, n_p, year):
```
fneum, line 295:
```suggestion
def update_dynamic_ptes_capacity(n: pypsa.Network, n_p: pypsa.Network, year: int) -> None:
```
We want more type hints for new code :)

scripts/add_brownfield.py
fneum: Separate is better.

fneum (commented):
Looks great! Just a few style comments.

amos-schledorn (approved):
Thanks for the changes! Can be merged once Fabian's comments are addressed, imo.

fneum (approved):

PR #1615, opened 2025-03-27
docs: update templates and help references

lkstrp:


fneum:
Maybe still point to the mailing list for reference to the "knowledge base" of previously answered questions.

PR #1617, opened 2025-03-28
Missing geothermal data

cpschau:

This PR implements two ways to deal with the missing geothermal potential data for Great Britain (GB) in the dataset of Manz et al. that was integrated into the workflow in #1516. The error was occurring when including geothermal as heat source for direct utilisation and heat pumps in district heating using the config parameter ``sector: heat_pump_sources: urban_central:``. The user can choose to ignore the missing data by assuming 0 potentials in GB by setting the parameter ``sector: district_heating: limited_heat_sources: geothermal: ignore_missing_regions`` to `true`. Otherwise, an error is thrown in the rule `build_geothermal_heat_potential`.

## Checklist

### DIFF
--- a/config/config.default.yaml
+++ b/config/config.default.yaml
@@ -482,6 +482,7 @@ sector:
     limited_heat_sources:
       geothermal:
         constant_temperature_celsius: 65
+        ignore_missing_regions: false
     direct_utilisation_heat_sources:
     - geothermal
   heat_pump_sources:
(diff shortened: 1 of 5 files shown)


amos-schledorn (approved):
Thanks for the changes! I think we should handle the country list more elegantly but I'd suggest doing this globally (see #1620 ).

config/config.default.yaml
```diff
@@ -482,6 +482,7 @@ sector:
     limited_heat_sources:
       geothermal:
         constant_temperature_celsius: 65
+        ignore_missing_regions: false
```
fneum, line 485:
```suggestion
        ignore_missing_regions: true
```
The default should be more lenient, I would say. Otherwise, the standard config will fail, wouldn't it?

fneum (approved):
Fine for me, but I think a plain warning without any configuration option, just handling of the reindex without raising ValueErrors would have been more than sufficient. 

```py
                logger.warning(
                    f"The onshore regions outside EU 27 ({non_covered_regions.to_list()}) have no heat source power. Filling with zeros."
                )
```

PR #1619, opened 2025-03-30
move snakemake pin to `>=9`

fneum:


fneum:
`mock_snakemake` works without trouble for various rules.

lkstrp:
Ref #1628

`snakemake-storage-plugin-http<0.3` is not compatible with snakemake 9, therefore pins are now
```
snakemake-storage-plugin-http>=0.3
snakemake-minimal>=9
```

PR #1626, opened 2025-04-02
Bugfix: Make plot_balance_maps more robust, when inactive links exist

bobbyxng:
- Adding another check whether carriers in the solved network also exist in `n.statistics.energy_balance()`
- Needed for those cases where inactive links (e.g. inactive CO2 pipelines) exist, but do not appear in energy balance. In this case, the previous workflow would break

```
carriers_in_eb = carriers[carriers.isin(eb.index.get_level_values("carrier"))]
eb.loc[components] = eb.loc[components].drop(index=carriers_in_eb, level="carrier")
```

## Checklist


fneum:
`ruff check --fix`

p-glaum (approved):

PR #1643, opened 2025-04-15
use absolute imports

lkstrp:
Closes # (if applicable).

- Change all imports of internal modules to absolute. Right now it is mostly a mix of sometimes absolute and sometimes not
- This might make problems with people's `mock_snakemake` setup. You will need to add the working dir to your Python path to resolve it. I'll add a release note for that.

## Checklist

### DIFF
--- a/rules/common.smk
+++ b/rules/common.smk
@@ -10,7 +10,7 @@ import os, sys, glob
 path = workflow.source_path("../scripts/_helpers.py")
 sys.path.insert(0, os.path.dirname(path))
 
-from _helpers import validate_checksum, update_config_from_wildcards
+from scripts._helpers import validate_checksum, update_config_from_wildcards
 from snakemake.utils import update_config
 
 
--- a/scripts/__init__.py
+++ b/scripts/__init__.py
@@ -2,4 +2,4 @@
 #
 # SPDX-License-Identifier: MIT
 
-# coding: utf-8
+
--- a/scripts/add_brownfield.py
+++ b/scripts/add_brownfield.py
@@ -11,14 +11,15 @@ import numpy as np
 import pandas as pd
 import pypsa
 import xarray as xr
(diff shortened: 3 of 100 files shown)


FabianHofmann:
this looks good @lkstrp and quite useful (just faced a case where I need this).

FabianHofmann (approved):

fneum:
Fine for me as long as `mock_snakemake` can still be used without much complication.

coroa:
Puh, i just found this now and this is opening us up to very tricky bugs.

TLDR: You now need to ALWAYS use import's with `from scripts.module import thing`, NEVER mix with `from module import thing`.

Python has an internal map that is seen exposed as a dictionary at `sys.modules`, which maps the import path to a loadable or loaded module. If you now have in your sys.path several ways to access the same modules, as we have:
```python
[
 ...
│   '.../test',
...
│   '.../test/scripts',
│   '.snakemake/pip-deployments'
]
```

`.../test` here is the repo root or it would be the `workflow` directory under it, if it existed. Let's imagine we have in `scripts` a `module.py` with only `class thing: pass` and a very simple `test.py` like
```python
from scripts.module import thing as abs_thing
from module import thing as rel_thing

print(abs_thing is rel_thing)
```

Then the first absolute import will be found through the `.../test` path and loads the module into `sys.module["scripts.module"]` and make `A` available as `abs_A`.

The second (actually also absolute) import will be found through the `.../test/scripts`  python path and load the same code into a new module `sys.module["module"]` (which maybe reuses the same `__pycache__` bytecode), but it definitely adds a new module object so that.

`abs_thing is rel_thing` will turn out to be `False`, as will `isinstance(abs_thing(), rel_thing)` and variations. ie the two code-identical classes are understood as separate entities, which can lead to very difficult to debug problems.

PR #1646, opened 2025-04-16
Add a TYNDP base network

tgilon:
**This PR builds on https://github.com/PyPSA/pypsa-eur/pull/1578 and must be merged after.**

## Checklist

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -60,6 +60,7 @@ test:
 	snakemake make_summary_perfect --configfile config/test/config.perfect.yaml
 	snakemake resources/test/networks/base_s_adm.nc --configfile config/test/config.clusters.yaml
 	snakemake --configfile config/test/config.scenarios.yaml -n
+	snakemake resources/tyndp/maps/power-network-s-all.pdf --configfile config/config.tyndp.yaml
 	echo "All tests completed successfully."
 
 unit-test:
@@ -72,6 +73,7 @@ clean-tests:
 	snakemake --configfile config/test/config.myopic.yaml --delete-all-output
 	snakemake make_summary_perfect --configfile config/test/config.perfect.yaml --delete-all-output
 	snakemake resources/test/networks/base_s_adm.nc --configfile config/test/config.clusters.yaml --delete-all-output
+	snakemake resources/tyndp/maps/power-network-s-all.pdf --configfile config/config.tyndp.yaml --delete-all-output
 	snakemake --configfile config/test/config.scenarios.yaml -n --delete-all-output
 
 # Removes all created files except for large cutout files (similar to fresh clone)
@@ -86,3 +88,7 @@ reset:
 		rm ./config/config.yaml || true; \
 		echo "Reset completed." \
 	) || echo "Reset cancelled."
+
+tyndp:
+	snakemake resources/tyndp/maps/power-network.pdf resources/tyndp/maps/power-network-s-all.pdf --configfile config/config.tyndp.yaml --rerun-incomplete $(args)
+	snakemake rulegraph filegraph --configfile config/config.tyndp.yaml
--- a/Makefile
+++ b/Makefile
@@ -60,7 +60,7 @@ test:
 	snakemake make_summary_perfect --configfile config/test/config.perfect.yaml
 	snakemake resources/test/networks/base_s_adm.nc --configfile config/test/config.clusters.yaml
 	snakemake --configfile config/test/config.scenarios.yaml -n
-	snakemake resources/tyndp/maps/power-network-s-all.pdf --configfile config/config.tyndp.yaml
+	snakemake resources/test/maps/power-network-s-all.pdf --configfile config/test/config.tyndp.yaml
 	echo "All tests completed successfully."
 
 unit-test:
@@ -73,7 +73,7 @@ clean-tests:
 	snakemake --configfile config/test/config.myopic.yaml --delete-all-output
 	snakemake make_summary_perfect --configfile config/test/config.perfect.yaml --delete-all-output
 	snakemake resources/test/networks/base_s_adm.nc --configfile config/test/config.clusters.yaml --delete-all-output
-	snakemake resources/tyndp/maps/power-network-s-all.pdf --configfile config/config.tyndp.yaml --delete-all-output
+	snakemake resources/test/maps/power-network-s-all.pdf --configfile config/test/config.tyndp.yaml --delete-all-output
 	snakemake --configfile config/test/config.scenarios.yaml -n --delete-all-output
 
 # Removes all created files except for large cutout files (similar to fresh clone)
@@ -88,7 +88,3 @@ reset:
 		rm ./config/config.yaml || true; \
 		echo "Reset completed." \
 	) || echo "Reset cancelled."
-
-tyndp:
-	snakemake resources/tyndp/maps/power-network.pdf resources/tyndp/maps/power-network-s-all.pdf --configfile config/config.tyndp.yaml --rerun-incomplete $(args)
-	snakemake rulegraph filegraph --configfile config/config.tyndp.yaml
--- a/config/config.default.yaml
+++ b/config/config.default.yaml
@@ -61,6 +61,7 @@ snapshots:
 enable:
   retrieve: auto
   retrieve_databundle: true
+  retrieve_tyndp_bundle: false
   retrieve_cost_data: true
   build_cutout: false
   retrieve_cutout: true
--- a/config/test/config.tyndp.yaml
+++ b/config/test/config.tyndp.yaml
@@ -3,7 +3,11 @@
 # SPDX-License-Identifier: CC0-1.0
 
 run:
-  prefix: "tyndp"
+  name: "test-elec-tyndp"
+  disable_progressbar: true
+  shared_resources:
+    policy: "test"
+  shared_cutouts: true
 
 scenario:
   clusters:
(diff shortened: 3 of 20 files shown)


tgilon:
@fneum To create a more relevant CI integration test for this feature, we would need to change the default test setting. By default, Belgium is used to reduce the data. This would not work here as Belgium has only one bidding zone. A good workaround would be to use a 1-week cutout, as suggested by @FabianHofmann. We can provide the data but must add them to the [Zenodo](https://zenodo.org/records/15130997).

tgilon:
@coroa FYI

Makefile
bobbyxng, line 1: why not call the rule name (instead of the output file)?

config/config.default.yaml
bobbyxng, line 1: does it make sense to split grid data from the remaining TYNDP input data? I see that the bundle is quite large, maybe some users only want to use the grid without using the full TYNDP scenarios

config/test/config.tyndp.yaml
```diff
@@ -0,0 +1,78 @@
...
+
+snapshots:
+  start: "2013-03-01"
+  end: "2013-03-08"
+
+enable:
+  retrieve_tyndp_bundle: true
+
+electricity:
+  base_network: tyndp-raw
```
bobbyxng, line 28: why call it "tyndp-raw" (the differentiation in osm was only necessary as there is osm-prebuilt and osm-raw)?

fneum:
> @fneum To create a more relevant CI integration test for this feature, we would need to change the default test setting. By default, Belgium is used to reduce the data. This would not work here as Belgium has only one bidding zone. A good workaround would be to use a 1-week cutout, as suggested by @FabianHofmann. We can provide the data but must add them to the [Zenodo](https://zenodo.org/records/15130997).

Can be found here now: https://zenodo.org/records/15349674/files/europe-2013-03-sarah3-era5.nc?download=1

config/config.default.yaml
fneum: Agreed.

config/test/config.tyndp.yaml
```diff
@@ -0,0 +1,78 @@
...
+    policy: false
+  shared_cutouts: true
+
+scenario:
+  clusters:
+  - all
+  planning_horizons:
+  - 2030
+
+countries: ['AL', 'AT', 'BA', 'BE', 'BG', 'CH', 'CZ', 'CY', 'DE', 'DK', 'EE', 'ES', 'FI', 'FR', 'GB', 'GR', 'HR', 'HU', 'IE', 'IT', 'LT', 'LU', 'LV', 'ME', 'MK', 'MT', 'NL', 'NO', 'PL', 'PT', 'RO', 'RS', 'SE', 'SI', 'SK']
```
fneum, line 18: what about Kosovo in these scenarios?

scripts/retrieve_tyndp_bundle.py
```diff
@@ -0,0 +1,60 @@
...
+import logging
+import os
+import zipfile
+
+from _helpers import configure_logging, progress_retrieve, set_scenario_config
+
+logger = logging.getLogger(__name__)
+
+# Define the base URL
+url = "https://zenodo.org/records/14230568/files/TYNDP_2024_data_bundle.zip"
```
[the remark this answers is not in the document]
tgilon: Done

config/config.default.yaml
tgilon: Done

Makefile
tgilon: I changed it to the rule name

scripts/cluster_network.py
bobbyxng, line 1: What about allowing for clustering? There were previous use cases in the past at a TSO to cluster bidding zones, if they were not in the scope of research, e.g. treating SE or IT as a single bidding zone.

tgilon: I’m open to that. I just want to make sure that the `base_network` is protected and remains consistent with its definition from the TYNDP 2024, in line with the user’s configuration.

config/test/config.tyndp.yaml
tgilon: I've switched to `tyndp`.

scripts/cluster_network.py
tgilon: I've removed the limitation. This enables the network to be clustered using administrative clustering.

For example, with level 0:

`networks/base_s_adm.nc`
image

`regions_onshore_base_s_adm.geojson`
image

rules/build_electricity.smk
```diff
@@ -799,7 +802,7 @@ rule prepare_network:
         "../scripts/prepare_network.py"
 
 
-if config["electricity"]["base_network"] == "osm-raw":
+if lambda w: config_provider("electricity", "base_network")(w) == "osm-raw":
```
fneum: I don't think that this works, because what is taken for `w` here? No wildcards exist outside rules.

config/test/config.tyndp.yaml
```diff
@@ -0,0 +1,78 @@
+# SPDX-FileCopyrightText: Contributors to PyPSA-Eur <https://github.com/pypsa/pypsa-eur>
```
fneum, line 1: Make sure it's reduced to only diff to `config.default.yaml`

config/config.default.yaml
```diff
@@ -61,6 +61,7 @@ snapshots:
 enable:
   retrieve: auto
   retrieve_databundle: true
+  retrieve_tyndp_bundle: false
```
fneum: Remove. Whether download is necessary should be automatically determined depending on data input requirements.

rules/retrieve.smk
```diff
@@ -169,6 +169,20 @@ if config["enable"]["retrieve"] and config["enable"].get("retrieve_cutout", True
             validate_checksum(output[0], input[0])
 
 
+if config["enable"]["retrieve"] and config["enable"].get("retrieve_tyndp_bundle", True):
```
fneum:
```suggestion
if config["enable"]["retrieve"] and config["enable"]["retrieve_tyndp_bundle"]:
```

All config settings are defined in `config.default.yaml`. We don't need to implement a fallback.

rules/retrieve.smk
```diff
@@ -169,6 +169,20 @@ if config["enable"]["retrieve"] and config["enable"].get("retrieve_cutout", True
...
+
+    rule retrieve_tyndp_bundle:
+        output:
+            dir=directory("data/tyndp_2024_bundle"),
```
fneum: Duplicated information. Use:

`Path(snakemake.input.reference_grid).parent.parent`

scripts/prepare_sector_network.py
```diff
@@ -6442,7 +6446,9 @@ def add_import_options(
         add_electricity_grid_connection(n, costs)
 
     for k, v in options["transmission_efficiency"].items():
-        if k in options["transmission_efficiency"]["enable"]:
+        if (k in options["transmission_efficiency"]["enable"]) and (
+            k == "DC" and snakemake.params != "tyndp"
```
fneum: `snakemake.params != "tyndp"` is missing a concrete reference to a parameter.

I don't think this is a good idea as it blocks the use of transmission efficiencies with the TYNDP network. Users who want to use the TYNDP network without losses should set the transmission efficiency to one in the configuration file.

scripts/prepare_sector_network.py
```diff
@@ -2037,6 +2037,10 @@ def add_storage_and_grids(
         h2_pipes = create_network_topology(
             n, "H2 pipeline ", carriers=["DC", "gas pipeline"]
         )
+        h2_buses_loc = n.buses.query("carrier == 'H2'").location
+        h2_pipes = h2_pipes.loc[
+            (h2_pipes.bus0.isin(h2_buses_loc)) & (h2_pipes.bus1.isin(h2_buses_loc))
+        ]
```
fneum:
```suggestion
  h2_buses_loc = n.buses.query("carrier == 'H2'").location
  h2_pipes = h2_pipes.query("bus0 in @h2_buses_loc and bus1 in @h2_buses_loc")
```

scripts/retrieve_tyndp_bundle.py
```diff
@@ -0,0 +1,86 @@
...
+The data bundle contains input data for the 2024 TYNDP scenario building process.
+
+This rule downloads the TYNDP data bundle from `zenodo
+<https://doi.org/10.5281/zenodo.14230568>` and extracts it in the ``data``
+subdirectory, such that all files of the bundle are stored in the
+``data/tyndp_2024_bundle`` subdirectory.
+
+**Outputs**
+
+- ``data/tyndp_2024_bundle``: input data for TYNDP 2024 scenario building
```
fneum: Needs to be updated to match revised URLs.

Please also add this new data source to `date-sources.rst`.

rules/build_electricity.smk
tgilon: You're right, I'll revert my changes.

config/test/config.tyndp.yaml
tgilon: Done

scripts/prepare_sector_network.py
tgilon: Good catch!

I'll remove it from the code and disable `DC` in the test configuration file as illustration.

config/config.default.yaml
tgilon: Done

rules/retrieve.smk
tgilon: This will be removed anyway due to the suppression of the `retrieve_tyndp_bundle` configuration.

rules/retrieve.smk
tgilon: Done

scripts/retrieve_tyndp_bundle.py
tgilon: Done

tgilon:
> Thanks for the revisions @tgilon!
> 
> Some minor aspects to iron out.

Thank you @fneum for the review! I've addressed all of your comments.

fneum (approved):

PR #1647, opened 2025-04-16
fix: specify cores for `snakemake>=9.2`

lkstrp:
With Snakemake 9.2 `-–cores` needs to be specified again. 
`ERROR    snakemake.logging:exceptions.py:181 Error: cores have to be specified for local execution (use --cores N with N being a number >= 1 or 'all')`

Maybe at some point we also wanna provide a couple of profiles.

Ref #1645

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -54,10 +54,10 @@ install-pinned-macos: _conda_check
 # Run default tests
 test:
 	set -e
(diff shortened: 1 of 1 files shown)


lkstrp:
Closed in favour of #1650

PR #1649, opened 2025-04-17
remove `config.yaml`

lkstrp:
Closes # (if applicable).

- Remove `config/config.yaml` from the snakemake config chain 
  - Currently the chain is `config/config.default.yaml`, `config/plotting.default.yaml`, `config/config.yaml` and then your passed config. For PyPSA-DE another conifg is added 
  - The function `copy_default_files` is very confusing and not good snakemake practice I would assume. The created `config.yaml` can lead to an artifact being used without the user realising it 
  - If we need a template which is not the default config, we can add a `config.template.yaml` but not run it by default 

Removing this is quite breaking. If we agree, I'll adjust the docs, release notes etc.

## Checklist

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -30,6 +30,7 @@ doc/_build
 /scripts/create_scenarios.py
 /config/create_scenarios.py
 
+config/config.yaml
 config/scenarios.yaml
 
 dconf
--- a/Makefile
+++ b/Makefile
@@ -98,7 +98,7 @@ clean-tests:
 
 # Removes all created files except for large cutout files (similar to fresh clone)
 reset:
-	@echo "Do you really wanna continue? This will remove logs, resources, benchmarks, results, and .snakemake directories (y/n): " && \
+	@echo "Do you really wanna continue? This will remove logs, resources, benchmarks, results, and .snakemake directories (config/config.yaml will not be deleted) (y/n): " && \
 	read ans && [ $${ans} = y ] && ( \
 		rm -r ./logs || true; \
 		rm -r ./resources || true; \
--- a/Snakefile
+++ b/Snakefile
@@ -20,6 +20,7 @@ from scripts._helpers import (
 
 configfile: "config/config.default.yaml"
 configfile: "config/plotting.default.yaml"
+configfile: "config/config.yaml"
 
 
 run = config["run"]
--- a/doc/configuration.rst
+++ b/doc/configuration.rst
@@ -9,7 +9,45 @@
 Configuration
 ##########################################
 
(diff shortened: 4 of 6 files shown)


FabianHofmann:
I agree with removing the full copy to defaults to config.yaml. this lead to a lot of issues. however removing the `config.yaml` completely, I am not sure about. it is quite breaking (which should be fine as this is our policy in pypsa-eur). just to get the rationale: The idea is that the user always sets a `--configfile` argument in the snakemake call as soon as they want to divert from the default?

lkstrp:
So you would remove the complete default copy, but keep `configfile: "config/config.yaml"` in the snakefile? So if people want to avoid passing `--configfile` on every run, they can still manually create and use that file?

Sounds good to me. You would still have the fourfold chain if you pass your own config file, and it is not super obvious that a `config.yaml` that does not even exist in the repo is being read in, but probably worth it for the convenience. Also makes this change less breaking

daniel-rdt:
@lkstrp Is there any update on this?

lkstrp:
No, we just need to choose one of the two options:
1. Remove `config.yaml` completely; you would always need to add `--configfile` as a flag
2. Remove `config.yaml` from the repository, but keep it in the chain so that it can still be used without adding a flag

I guess people would prefer the second

daniel-rdt:
> No, we just need to choose one of the two options:
> 
> 1. Remove `config.yaml` completely; you would always need to add `--configfile` as a flag
> 2. Remove `config.yaml` from the repository, but keep it in the chain so that it can still be used without adding a flag
> 
> I guess people would prefer the second

I personally find that having `config.yaml` in the chain can be quite annoying and also confusing, especially, for first time users (because it often happened by accident that the `config.yaml` is in there as an outdated copy of the default config) which leads to problems.

I think we should remove the copy rule but we could keep the `config.yaml` in the chain but have it as an empty custom config file. This way by default it does not overwrite anything but only does so if the user explicitly adds anything inside this custom config? We could even rename the file per default to something like: `config.custom.yaml` or some similar naming.

fneum:
> So you would remove the complete default copy, but keep configfile: "config/config.yaml" in the snakefile? So if people want to avoid passing --configfile on every run, they can still manually create and use that file?

This sounds like the best compromise. The important aspects to me are:
- No requirement to pass `--configfile`
- Untracked `config.yaml` file. (Otherwise, I bet most PRs will have contributors config)
- No renaming so it's not breaking.
- Agree to remove default copy.

lkstrp:
This is good to go now

PR #1650, opened 2025-04-21
Revert "docs: remove `-call` argument (#1386)"

tgilon:
This reverts commit 225dae0e1ce781344d3e1b09745491241bbadaa5, PR https://github.com/PyPSA/pypsa-eur/pull/1386.

As of `snakemake>=9.2.0` ([changelog](https://snakemake.readthedocs.io/en/stable/project_info/history.html#id4)), the default behaviour of `-cores` for `snakemake` is restored. This PR suggests reverting the changes and updates the environment.

From `snakemake` changelog:

> restore accidentally changed behavior of –cores. If not specified, Snakemake now complains again, asking for –cores to be specified (of –jobs in case of remote exec). As before, if a default is desired, it can be easily set via a profile ([#3531](https://github.com/snakemake/snakemake/issues/3531)) ([11b2b30](https://github.com/snakemake/snakemake/commit/11b2b307002c2cb16c2b0faf334c068b68e83a24))

## Checklist

### DIFF
--- a/envs/environment.yaml
+++ b/envs/environment.yaml
@@ -21,7 +21,7 @@ dependencies:
 - xlrd
 - openpyxl
 - seaborn
-- snakemake-minimal>=9.2
+- snakemake-minimal>=9
 - memory_profiler
 - yaml
 - pytables
--- a/envs/environment.yaml
+++ b/envs/environment.yaml
@@ -21,7 +21,7 @@ dependencies:
 - xlrd
 - openpyxl
 - seaborn
-- snakemake-minimal>=9
+- snakemake-minimal>=9.2
 - memory_profiler
 - yaml
 - pytables
--- a/Makefile
+++ b/Makefile
@@ -54,11 +54,11 @@ install-pinned-macos: _conda_check
 # Run default tests
 test:
 	set -e
-	snakemake solve_elec_networks --configfile config/test/config.electricity.yaml
-	snakemake --configfile config/test/config.overnight.yaml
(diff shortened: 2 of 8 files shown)


lkstrp (commented):
But `--cores` is not the same as `--call`?

The issue with `--cores` you are mentioning has been addressed in #1647, but is not merged yet

tgilon:
Did I missed smth? As far as I know `-call` is the contraction of `--cores all`.

But correct, I missed your PR. It might be good to update more than just the `Makefile`.

lkstrp:
Ahh, no, I missed something. `-call` wasn't in the docs when I searched for it, but I didn't make the connection with the contraction of `--cores all` either. Which makes much sense. Then this PR is much better, thank you!

envs/environment.yaml
```diff
@@ -21,7 +21,7 @@ dependencies:
 - xlrd
 - openpyxl
 - seaborn
-- snakemake-minimal>=9
+- snakemake-minimal>=9.2
```
lkstrp, line 24: But we don't need that limitation `snakemake>=9.2`, is it?

tgilon: I wanted to be careful here, as the behaviour is restored in 9.2.0 ([changelog](https://snakemake.readthedocs.io/en/stable/project_info/history.html#id4)).

lkstrp: It was, but we are adding a flag that was optional before, so `>=9` is fine

PR #1663, opened 2025-04-30
chore: add env file to support `uv`

lkstrp:
Ref #1660 

@stefan-oet something like this we could add next to #1660 

```
# requirements.in file to be used with package managers like pip or uv and the
# equivialent to 'envs/environment.yaml' for conda. 
# The locked version of this file is in `envs/locks/requirements.txt` and can is 
# regularly updated.

# To create a environment with uv, run:
# uv venv # (create new environment)
# uv pip sync envs/locks/requirements.txt # (install dependencies from locked file)

# WARNING
# The PyPSA-Eur workflow depends on non-Python packages like glpk, which cannot be 
# installed via uv or pip. You need to install them separately if you want to use
# this requirements file. To get a environment with all needed dependencies installed, 
# use conda/mamba (see 'envs/environment.yaml')
```

## Checklist


coroa:
Could those reps move into a project.toml?

So that I can just do uv run snakemake?

stefan-oet:
`pyproject.toml` would definitely be nice to have even if we don't have a package yet.

In addition to future modularization, we could also already define different dependency groups (e.g. core vs development) which is nicer than having for example a separate `-dev.txt` file.

lkstrp:
I'm convinced, let's go 🚀

It's also nice to be able to add other configurations. I'll adjust the PR

lkstrp:
Sorry, I haven't got to it yet. But go ahead — we can still merge your version via another pull request/ in here if the data catalogue one stays open for a while

fneum:
No objections to merging this, but could we add this option to the documentation then (with a list of packages that need to installed by other means)?

Wondering where the GLPK dependency comes from. It should no longer be required.

fneum:
I tried locally and it also worked smoothly for me.

We keep conda support.

In the docs, perhaps add installation instructions with `uv` and mark as "experimental".

Someone does the test on HPC - we don't have to wait for this to merge.

lkstrp:
I did an HPC test and it didn't work. So we need to mention that in the docs

fneum:
@lkstrp, should we just get this on the way? We can mark it as experimental.

lkstrp:
Yes! I just discussed it with @Irieo and was reving this. I only wanna wait for #1675 to include the new dependencies and have a stable CI

lkstrp:
Closed in favour of #1886

PR #1664, opened 2025-04-30
feat: add PTES booster heat pumps

TomKae00:

This PR adds supplemental heating for pit thermal energy storages (PTES) via a booster heat pump. 

- Supplemental Heating can be enables in the configs to cap the PTES temperature at 90 °C, preventing any district-heating flow above this limit from discharging directly from the PTES.
- When enabled in the config, a booster heat pump raises forward temperatures above the PTES maximum.
- PTES storage capacity and supplemental heating scripts have been merged into ptes_operations for more consistent code structure and documentation.
- The model now estimates when supplemental heating is needed, and it treats the PTES as a heat source for heat pumps when enabled (binary flag).
- The PTES clipped-top temperature is now calculated in a dedicated script used for capacity approximation and for computing the booster heat-pump COP.

## Bugfix 

- When running with ``foresight: perfect``, the code encountered an error in ``prepare_perfect_foresight.py`` due to the dynamic ``efficiency2`` of the ``ptes_heat_pump``. I’ve implemented a workaround to address the issue and added a TODO to mark this in the script for future resolution.

## Configs

To activate supplemental heating for PTES with a booster_heat_pump, set the following in your configuration:

```yaml
sector:
  district_heating: 
     ptes:
        supplemental_heating:
           enable: true
           booster_heat_pump: true
```

## Testing

```yaml
scenario:
  clusters:
    - 8

planning_horizons:
  - 2030

foresight: overnight
countries: ['DE', 'PL']

clustering:
  temporal:
    resolution_sector: 12h
```

# Results

## With booster HP and supplemental heating for PL 01

When the district-heating temperature exceeds 90 °C, direct PTES discharge halts and the booster heat pump takes over. The PTES then behaves more like a seasonal storage. Note that, due to high investment costs, the booster pump sizing remains much smaller than the PTES charge/discharge units.

overview_6plots_PL0_1

## w/o booster heat pump and no supplemental heating for PL 01 

With no supplemental heating, the PTES operates as a highly flexible storage asset.

ptes_profile_PL0_1

## With booster HP and supplemental heating for DE 01

overview_6plots_DE0_1

## w/o booster heat pump and no supplemental heating for DE 01 

ptes_profile_DE0_1

## Systems costs 

Activating supplemental heating and the booster heat pump increases overall system costs.

additional_heat_cost_compare_DE0_1

## Energy Balance with booster HP and supplemental heating

The PTES charge/discharge cycles decline under temperature restrictions, reducing system flexibility.

final_urban_central_heat

## Energy Balance w/o booster heat pump and no supplemental heating 

master_urban_central_heat

## Checklist


scripts/build_tes_supplemental_heating/run.py
```diff
@@ -0,0 +1,102 @@
+# SPDX-FileCopyrightText: Contributors to PyPSA-Eur <https://github.com/pypsa/pypsa-eur>
```
amos-schledorn, line 1: It seems a bit overkill to introduce two additinal classes for this. I'd suggest to:
- Make the PTES temperature and direct utilisation profile an output of the `build_tes_capacity` rule
- If that's not possible, re-use some of the code from that rule, possibly extending it by an explicit output of the PTES temperatures.

In any case, we should make sure we don't implement the same logic twice, since that might cause inconsistencies down the road.

scripts/prepare_sector_network.py
```diff
@@ -3140,16 +3034,32 @@ def add_heat(
...
+                    tes_supplemental_heating_profile = xr.open_dataarray(
+                        tes_supplemental_heating_profile_file
+                    )
+                    tes_supplemental_heating = (
+                        tes_supplemental_heating_profile.sel(name=nodes)
+                        .to_pandas()
+                        .reindex(index=n.snapshots)
+                    )
+                else:
+                    tes_supplemental_heating = 1  # pd.Series(1.0, index=n.snapshots)
```
amos-schledorn, line 3049: Looks like we have a widow comment here.

scripts/prepare_sector_network.py
```diff
@@ -3140,16 +3034,32 @@ def add_heat(
                     ],
                 )
 
+                if options["district_heating"]["ptes"]["supplemental_heating"][
+                    "enable"
+                ]:
+                    tes_supplemental_heating_profile = xr.open_dataarray(
+                        tes_supplemental_heating_profile_file
+                    )
+                    tes_supplemental_heating = (
```
amos-schledorn, line 3043: I'm splitting hairs here but let's be clearer with the naming: a) as long as `tes_<...>` (here and elsewehere, both variable and rule names) only includes PTES but not all TES technologies, let's call this `ptes_<...>`; b) let's make the name explicit (something like `ptes_supplemental_heating_required`)

amos-schledorn (changes requested):
@TomKae00 this looks great, thanks! I have a few nit-picky coding comments. 

Since the changes seem to render PTES unusable in winter, I'd suggest turning supplemental heating off by default in the config (when everything else is finished, after the CI has passed). Or: we try boosting via resistive heaters.

I know you were already planning this: The PR could also use some more explanations of the feature logic.

scripts/prepare_sector_network.py
TomKae00: I agree! I have updated all cases to use ptes instead of tes and renamed tes_supplemental_heating to ptes_supplemental_heating_required.

scripts/build_tes_supplemental_heating/run.py
TomKae00: That's a good point! I've now merged the ``ptes_capacity_approximator`` and the ``supplemental_heating_approximator`` into a single directory called ``ptes_operations``. This helps ensure a more consistent structure and avoids potential inconsistencies.

TomKae00:
@amos-schledorn Thanks for the review! I’ve implemented your suggestions, and the structure looks cleaner now. The CI appears to be passing, and after your review, I’ll keep ``supplemental_heating`` and ``booster_heat_pump`` disabled.

@cpschau Since I’ve merged the ``ptes_capacity_approximator`` into the ``build_ptes_operations`` directory, it might be helpful if you could have a quick look to see if the changes look alright to you.

amos-schledorn:
Closing this PR, as changes are migrated to #1692

PR #1688, opened 2025-05-13
Fix macOS builds by switching pyscipopt from pip to conda-forge

stefan-oet:
Closes # (if applicable).

The new lock files approach introduced in https://github.com/PyPSA/pypsa-eur/pull/1660 caused macOS builds to [fail](https://github.com/PyPSA/pypsa-eur/actions/runs/14999742758/job/42143208527) because installing `pyscipopt` via pip does not include scip. 

The proposed solution is to fetch `pyscipopt` from conda-forge instead of pip. This package properly bundles scip providing a complete and working environment (see updated locked files included in this PR)

The downside is that we have to drop support for ARM linux as `pyscipopt` does not currently provide builds via conda-forge or pip for this platform. Users will either need to install it from source or get if via a package manager when available (e.g. [Arch Linux ARM package](https://archlinuxarm.org/packages/aarch64/python-pyscipopt))

@lkstrp 

## Checklist


lkstrp (approved):
As already discussed, that's fine. It's unlikely that anyone needs to run Eur on Linux ARM
Thanks for the fixes! @stefan-oet

PR #1690, opened 2025-05-14
Install snakemake plugins via pip instead of bioconda

stefan-oet:
Closes # (if applicable).

This PR defines the snakemake plugins explicitly as pip dependencies to stay [consistent](https://github.com/PyPSA/pypsa-eur/blob/master/envs/linux-pinned.yaml#L507-L510) with how the pinned envs got resolved.

This should also fix [this](https://github.com/PyPSA/pypsa-eur/actions/runs/15018446080/job/42201759810) test workflow, even though the reported missing package (`snakemake-executor-plugin-slurm==1.2.5=pyhdfd78af_0`) does [exist](https://anaconda.org/bioconda/snakemake-executor-plugin-slurm/files) on bioconda.

There is definitely something going on with this package as in the past 2 days there have been 6 releases on [pypi](https://pypi.org/project/snakemake-executor-plugin-slurm/#history) (1.2.10 being the latest) versus 2 on bioconda (1.2.5 being the latest). So there is also some delay there as well.

@lkstrp 

## Checklist


stefan-oet:
Just had a look at the [BiocondaBot](https://github.com/bioconda/bioconda-recipes/pull/56141) which autobumps versions but does not open separate PR's for each version release...

This means that the next version that will be released on bioconda will be 1.2.10. The versions between (1.2.6-1.2.9) will never reach the upstream bioconda channel. So it does not look like relying on bioconda is a good idea.

euronion:
I don't see why that would be a problem per se?
As long as the dependencies are defined and we get them from the specified source...
Do we need to have access to all releases that have ever been made?

I believe they are intentionally not releasing `1.2.6-1.2.9` on `bioconda` because they are just testing things and not making any changes to the package:

image

`snakemake` has been released through `bioconda` since

What we shouldn't do is mix `pypi` and `bioconda` for `snakemake`.
`bioconda` is the [recommended installation vector](https://snakemake.readthedocs.io/en/stable/getting_started/installation.html#installation-via-conda-mamba), so I do not anticipate the plugins there to get neglected. I am still in favor for it for the `plugins`.

stefan-oet:
Sure, makes sense

PR #1711, opened 2025-06-05
feat: resistive_heaters and air_sourced_hp for ptes temperature boosting

TomKae00:

This feature introduces resistive heaters and air-sourced heat pumps as temperature boosting technologies for PTES. The boosting methodology is based on a temperature_boost_ratio, which calculates the additional temperature lift required between the store’s current top temperature and the target forward temperature. This is done relative to the lift already achieved within the store, using the return temperature of the district heating system as a reference.

---

### 1. Boosting-ratio derivation  
Let  
- Ṽ = volumetric flow rate (constant),  
- ρ = fluid density (constant),  
- cₚ = specific heat capacity (constant),  
- T_return = district heating return temperature,  
- T_max,store = temperature in the PTES,  
- T_forward = forward temperature.  

We define two heat-transfer rates:  
1. Energy supplied by the PTES (source)
Q_source = Ṽ·ρ·cₚ·(T_max,store − T_return)

2. Additional energy required (boost)  
Q_boost  = Ṽ·ρ·cₚ·(T_forward − T_max,store)

The total heating demand is  
Q_total  = Q_source + Q_boost

Define the boosting ratio α as the lift-ratio between what the pump must add versus what the PTES already delivers:  
α = (T_forward − T_max,store) / (T_max,store − T_return)  =  Q_boost / Q_source

In other words,  
Q_boost  = α·Q_source

Proceeding from Q_boost  = P_el·η, one finds  
Q_source = (P_el·η) / α
Q_total  = Q_source + Q_boost  =  (P_el·η)/α + P_el·η  =  P_el·η·(1/α + 1)

Thus:  
1. PTES output per unit electricity
Q_source / P_el  =  η / α

2. Total output per unit electricity
Q_total  / P_el  =  η·(1/α + 1)

---

### 2. COP calculation adjustment for heat pumps  
Previously, we calculated the heat-pump COP using the district heating forward/return temperatures. To accurately model air-sourced ptes heat pumps, the COP is now based on the pump’s **sink inlet and outlet temperatures**. This change ensures consistency whenever an external heat source (e.g., ambient air) is used for boosting.

---

### 3. Passing heat-pump parameters correctly  
Prior, heat-pump parameters (set in config) were not used and default value instead. Now, all provided parameters are properly passed through the config and used.

---

### 4. Improved handling in `logarithmic_mean`  
The function `logarithmic_mean(t_hot, t_cold)` failed (division-by-zero) when t_hot = t_cold. It now returns t_hot in that case (L’Hôpital’s rule), instead of throwing an error. This specifically handles instances where the PTES temperature equals the forward temperature.

---

### 5. @amos-schledorn discussion about heat-source flags
Because the air-sourced heat pump is now used in two contexts—  
1. **Standard heating**: raising district-heating return temperature up to forward temperature, and  
2. **Boosting**: raising PTES outlet temperature up to forward temperature—  
both cases would have “heat source = Air”, but this would break the logic we have at the moment. Therefore I still used “heat source = ptes” to manually distinguish the two scenarios. It might be cleaner to introduce a separate identifier for “boosting heat source.” Do you have a quick suggestion on how best to differentiate these two cases in the codebase?

---

## Testing

```yaml
scenario:
  clusters:
    - 8

planning_horizons:
  - 2030

foresight: overnight
countries: ['DE', 'PL']

clustering:
  temporal:
    resolution_sector: 12h
```

## Results 

1. System-cost comparison  
- With supplemental heating:
  System cost = 141.071 trillion
- With supplemental heating disabled:
  System cost = 141.054 trillion
2. supplemental heating 

Ran the analysis for DE and PL, both of which require relatively high forward temperatures (over 90°C). Since the PTES maximum storage temperature is 90 °C, supplemental boosting is necessary whenever the forward temperature exceeds that limit.

3. Preferred boosting technology

In both countries, resistive heaters were chosen over air-sourced heat pumps for boosting. The main reasons are:
- The air heat pump’s COP at high outlet temperatures is relatively low.
- Resistive heaters have much lower capital costs.
- The full-load hours for each booster technology are limited, favoring resistive heaters.

4. Interpretation of the plots

- The blue line shows electricity consumption for boosting (either by resistive heater or heat pump).
- The combined heat output is Q_total = Q_store + P_el·η
- The boosting_ratio determines how much extra energy (blue area) must be added beyond what the PTES alone can supply to achieve the forward temperature

### With supplemental heating and booster technologies for PL 01 (resistive heaters)

overview6plots_resitive_PL0_1

### With supplemental heating and booster technologies for PL 01 (HP)

overview6plots_hp_PL0_1

### With supplemental heating and booster technologies for DE 01 (resistive heaters)

overview6plots_resitive_DE0_1

### With supplemental heating and booster technologies for DE 01 (HP)

overview6plots_hp_DE0_1

## Checklist

### DIFF
--- a/doc/configtables/sector.csv
+++ b/doc/configtables/sector.csv
@@ -21,6 +21,9 @@ district_heating,--,,
 -- -- relative_annual_temperature_reduction,, float, Relative annual reduction of district heating forward and return temperature - defaults to 0.01 (1%)
 -- ptes,,,
 -- -- dynamic_capacity,--,"{true, false}",Add option for dynamic temperature-dependent energy capacity of pit storage in district heating
+-- -- supplemental_heating,,,
+-- -- -- enable,--,"{true, false}",Add option to enable supplemental heating of pit storage in district heating
+-- -- -- booster_heat_pump: true,--,"{true, false}",Add option to enable a booster heat pump for supplemental heating of pit storage in district heating
 -- -- max_top_temperature,C,float,The maximum top temperature of the pit storage according to DEA technology catalogue (2018)
 -- -- min_bottom_temperature,C,float,The minimum bottom temperature of the pit storage according to DEA technology catalogue (2018)
 -- ates,,,
--- a/doc/release_notes.rst
+++ b/doc/release_notes.rst
@@ -39,6 +39,8 @@ Release Notes
 
 * Added aquifer thermal energy storage (ATES) to district heating. Some parameters (CAPEX, standing losses) might require tuning by the user. Eligibility computation is relatively basic. Turned off by default.
 
+* Added supplemental heating of thermal energy storages (currently implemented for PTES). This can be enabled by setting: ``sector: district_heating: ptes: supplemental_heating: true`` . To enable a boosting heat pump as the supplemental heating technology, use: ``sector: district_heating: ptes: supplemental_heating: booster_heat_pump: true``
+
 * Non-sequestered HVC (plastic waste) is now allocated based on the population instead of production. It can be either burned without energetic utilization or in CHPs to support the district heating system.
 
 * Developer note: Scripts now use absolute imporys. When using `mock_snakemake` this 
--- a/rules/build_sector.smk
+++ b/rules/build_sector.smk
@@ -1379,6 +1379,15 @@ rule prepare_sector_network:
             )(w)
             else []
         ),
+        ptes_direct_utilisation_profiles= lambda w: (
+            resources(
+                "ptes_direct_utilisation_profiles_s_{clusters}_{planning_horizons}.nc"
+            )
+            if config_provider(
+                "sector","district_heating","ptes","supplemental_heating","enable"
+            )(w)
+            else[]
+        ),
         solar_thermal_total=lambda w: (
             resources("solar_thermal_total_base_s_{clusters}.nc")
             if config_provider("sector", "solar_thermal")(w)


scripts/prepare_sector_network.py
```diff
@@ -3053,34 +3056,72 @@ def add_heat(
...
+                        nodes,
+                        suffix=f" {heat_system} water pits discharger",
+                        bus0=nodes + f" {heat_system} water pits",
+                        bus1=nodes + f" {heat_system} heat",
+                        carrier=f"{heat_system} water pits discharger",
+                        efficiency=costs.at[
+                            "central water pit discharger",
+                            "efficiency",
+                        ]
+                        * ptes_supplemental_heating_required,
```
cpschau: Couldn't we just save the variable `ptes_supplemental_heating_required` in this else-clause?

scripts/prepare_sector_network.py
```diff
@@ -3264,33 +3305,46 @@ def add_heat(
...
-                and options["district_heating"]["ptes"]["supplemental_heating"][
-                    "booster_heat_pump"
+                and "heat_pump"
+                in options["district_heating"]["ptes"]["supplemental_heating"][
+                    "booster_technologies"
                 ]
             ):
                 raise ValueError(
-                    "'booster_heat_pump' is true, but 'enable' is false in 'supplemental_heating'."
+                    "Supplemental heating: 'booster_technologies' contains 'heat_pump', but 'enable' is false."
```
cpschau: I guess this is a matter of taste: As it is, we incentivize the user to change both `enable` and `boosting_technologies`. Personally, I find it practical being able to switch a feature on/off by just setting the `enable` parameter in the config file, especially with a cascaded config setup. Compromise would be a warning instead of an error.

scripts/prepare_sector_network.py
```diff
@@ -3331,6 +3385,56 @@ def add_heat(
...
+            if (
+                not options["district_heating"]["ptes"]["supplemental_heating"][
+                    "enable"
+                ]
+                and "resistive_heaters"
+                in options["district_heating"]["ptes"]["supplemental_heating"][
+                    "booster_technologies"
+                ]
+            ):
+                raise ValueError(
```
cpschau: See above

scripts/build_cop_profiles/CentralHeatingCopApproximator.py
```diff
@@ -153,7 +170,7 @@ def approximate_cop(self) -> Union[xr.DataArray, np.array]:
             Union[xr.DataArray, np.array]: The calculated COP values.
         """
         return xr.where(
-            self.t_source_in_kelvin >= self.t_sink_out_kelvin,
+            (self.delta_t_lift < self.min_delta_t_lift),
```
cpschau, line 173: This means, that for the default `min_delta_t_lift` of 10 K, PTES could not be discharged with heat pumps to the district heating system if the forward flow temperature is between 90 and 100 °C, right?

1. Is this realistic?
2. This would suppress the capacity factors of booster heat pumps further down and give another reason why there not favored for boosting applications given the higher CAPEX compared to resistive heaters.

scripts/prepare_sector_network.py
```diff
@@ -3264,33 +3305,46 @@ def add_heat(
...
             if (
                 heat_source in params.temperature_limited_stores
                 and options["district_heating"]["ptes"]["supplemental_heating"][
                     "enable"
                 ]
-                and options["district_heating"]["ptes"]["supplemental_heating"][
-                    "booster_heat_pump"
+                and "heat_pump"
+                in options["district_heating"]["ptes"]["supplemental_heating"][
+                    "booster_technologies"
```
cpschau: Why do the booster heat pumps stand in an if-else-relation with normal heat pumps (see else-clause below in L.3355 and following)? We want the latter ones to be available independent of the boosting feature.

scripts/prepare_sector_network.py
```diff
@@ -3331,6 +3385,56 @@ def add_heat(
...
+                    xr.open_dataarray(ptes_temperature_boost_ratio_profile_file)
+                    .sel(name=nodes)
+                    .to_pandas()
+                    .reindex(index=n.snapshots)
+                )
+
+                n.add(
+                    "Link",
+                    nodes,
+                    suffix=f" {heat_system} ptes resistive heater",
```
cpschau: Maybe change the "ptes" part to "pits" or even "water pits" to be more consistent, avoid abbreviations and allow a more convenient filtering of links.

config/plotting.default.yaml
```diff
@@ -470,11 +471,13 @@ plotting:
     CHP electric: '#8a5751'
     district heating: '#e8beac'
     resistive heater: '#d8f9b8'
+    ptes resistive heater: '#a0e78d'
     residential rural resistive heater: '#bef5b5'
     residential urban decentral resistive heater: '#b2f1a9'
     services rural resistive heater: '#a5ed9d'
     services urban decentral resistive heater: '#98e991'
     urban central resistive heater: '#8cdf85'
+    urban central ptes resistive heater: '#71c96b'
```
cpschau: Do we need both entries?

cpschau (changes requested):
Hey @TomKae00 ,
first of all: Kudos to the nice mathematical description and the overall documentation of the PR. I left a couple of comments, most of them questions to better understand what is happening or stylistic remarks. Some of them might be worth discussing together with @amos-schledorn.

scripts/build_cop_profiles/CentralHeatingCopApproximator.py
TomKae00: You're right, this limitation applies when the source temperature is around 90 °C, as would be the case if the PTES were used as the heat source. However, in our setup, we're boosting the PTES temperature using air as source on the evaporator side, since modeling heat extraction from PTES on the evaporator side has proven challenging. So in our case, PTES is connected to the condenser side, and ambient air serves as the heat source.

A minimum temperature lift of 10 K is technically reasonable due to constraints in heat pump circuit design. This value is also supported in the literature: Aleksandrs et al. (2020) reference it based on findings from Meggers et al. (2010).

scripts/prepare_sector_network.py
TomKae00: The ``if-else`` logic was already part of the original loop structure before my PR. Specifically, it handles the case where the ``heat_source`` is not in ``limited_heat_sources`` and then proceeds with the ``else`` branch. What I introduced is an additional condition: if the ``heat_source`` is part of ``temperature_limited_stores``, a new logic branch for booster heat pumps is triggered.

Currently, the only ``heat_source`` that enters this booster logic is ``"ptes"``. While this is not entirely accurate, since the actual heat source for boosting is typically ambient air. This limitation is already acknowledged in Point 5 of the PR documentation, which discusses the need for clearer heat-source flags.

Given that ``"ptes"`` is the only heat_source entering this logic, and considering that the booster heat pump is dimensioned with specific inlet and outlet temperatures for PTES boosting, we intentionally avoid making this component available outside the boosting context.

config/plotting.default.yaml
TomKae00: you're right, we only need  ``"urban central ptes resistive heater"`` here.

cpschau (approved):
Thanks for the explanations @TomKae00! This looks good to me and can be merged!

doc/configtables/sector.csv
```diff
@@ -47,6 +48,7 @@ district_heating,--,,
 -- -- heat_exchanger_pinch_point_temperature_difference,K,float,Heat pump pinch point temperature difference in heat exchangers assumed for approximation.
 -- -- isentropic_compressor_efficiency,--,float,Isentropic efficiency of heat pump compressor assumed for approximation. Must be between 0 and 1.
 -- -- heat_loss,--,float,Heat pump heat loss assumed for approximation. Must be between 0 and 1.
+-- -- min_delta_t_lift,--,float,Minimum feasible temperature lift for heat pumps assumed for approximation.
```
amos-schledorn: Could we make this a little more concrete? (Maybe by adding an extra sentence)

doc/release_notes.rst
```diff
@@ -29,13 +29,15 @@ Release Notes
 
 **Changes**
 
+* Added resistive heaters as a temperature boosting technology for PTES. Additionally, the heat pump booster now uses ambient air as its heat source for boosting.
```
amos-schledorn: Let's combine this entry with the one further below into one single point.

scripts/build_cop_profiles/run.py
```diff
@@ -22,6 +22,8 @@
                 heat_exchanger_pinch_point_temperature_difference
                 isentropic_compressor_efficiency:
                 heat_loss:
+                min_delta_t_lift:
+                min_delta_t_condenser:
```
amos-schledorn: Might be my bad, but I can't find `min_delta_t_condenser` as a config setting.

scripts/build_ptes_operations/ptes_temperature_approximator.py
```diff
@@ -106,3 +106,19 @@ def e_max_pu(self) -> xr.DataArray:
...
+    def temperature_boost_ratio(self) -> xr.DataArray:
+        """
+        Calculate the additional lift required between the store's
+        current top temperature and the forward temperature with the lift
+        already achieved inside the store.
+
+        Returns
+        -------
+        xr.DataArray
+            The resulting fraction of PTES charge that must be further heated.
```
amos-schledorn: Should we add your derivation of the ratio from the PR to the `Notes` in the docstring?

scripts/prepare_sector_network.py
```diff
@@ -2752,7 +2752,8 @@ def add_heat(
     direct_heat_source_utilisation_profile_file: str,
     hourly_heat_demand_total_file: str,
     ptes_e_max_pu_file: str,
-    ptes_direct_utilisation_profile: str,
+    ptes_direct_utilisation_profile_file: str,
+    ptes_temperature_boost_ratio_profile_file: str,
```
amos-schledorn: Nitpick: If we rename the parameters, we could make them `..._filename` to be the most accurate.

scripts/prepare_sector_network.py
```diff
@@ -3053,34 +3056,72 @@ def add_heat(
...
+                            "central water pit discharger",
+                            "efficiency",
+                        ]
+                        * ptes_supplemental_heating_required,
+                        efficiency2=costs.at[
+                            "central water pit discharger",
+                            "efficiency",
+                        ]
+                        * (ptes_supplemental_heating_required - 1)
+                        * (-1),
```
amos-schledorn: Couldn't we just use `(1 - ptes_supplemental_heating_required)` in the line above and get rid of the `(-1)`?

scripts/prepare_sector_network.py
```diff
@@ -3331,6 +3385,56 @@ def add_heat(
...
+                not options["district_heating"]["ptes"]["supplemental_heating"][
+                    "enable"
+                ]
+                and "resistive_heaters"
+                in options["district_heating"]["ptes"]["supplemental_heating"][
+                    "booster_technologies"
+                ]
+            ):
+                raise ValueError(
+                    "Supplemental heating: 'booster_technologies' contains 'resistive_heaters', but 'enable' is false."
```
amos-schledorn: Could we simplify the config logic by allowing the boosting technology list to be non-empty while `enable=False` but only throwing an error when `enable=True` and the list is empty?

scripts/build_cop_profiles/run.py
TomKae00: You're right, thanks for pointing that out! I had removed this setting but forgot to update the docstring accordingly.

scripts/prepare_sector_network.py
TomKae00: Good point, your suggestion is cleaner and does the same thing.

scripts/prepare_sector_network.py
TomKae00: Yes, sounds good to me, let's go with your suggestion.

scripts/build_ptes_operations/ptes_temperature_approximator.py
TomKae00: Good idea, this will help users better understand the background of the ratio.

scripts/prepare_sector_network.py
TomKae00: I checked and noticed that similar parameters in prepare_sector_network are consistently named with _file, so for consistency I'll keep it that way.

PR #1712, opened 2025-06-08
Solar aggregation in CCL plus minor fixes

Aym-brz:
Improved the behavior of ``agg_p_nom_limits``: added the ability to aggregate all ``solar`` electric technologies.
Improved the handling of heating capacities from before the first planning horizon: if no older heating capacity is available, a warning is displayed instead of causing an error later.

## Checklist

### DIFF
--- a/config/config.default.yaml
+++ b/config/config.default.yaml
@@ -987,6 +987,7 @@ solving:
 
   agg_p_nom_limits:
     agg_offwind: false
+    agg_solar: false
     include_existing: false
     file: data/agg_p_nom_minmax.csv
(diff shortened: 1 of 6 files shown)


JuliaSvn:
The solar aggregation add-on works fine on my side (no error) + the minor fixes solve some bugs, however the CCL constraints are still not working. Does any maintainer have an idea why it is not working ? When we print the constraint on p_nom they seem ok, however after the optimisation, the network does not respect the constraint anymore (which calls into question whether the model respects the other constraints?).

PR #1717, opened 2025-06-17
bugfix: add existing renewables consistent with renewable_carriers list

daniel-rdt:
This PR in adding existing renewables. It proposes that existing renewables are only added to the model in `add_existing_baseyear` if they are also listed in [`renewable_carriers`](https://github.com/PyPSA/pypsa-eur/blob/1161a7c8e0f0e2952253df04cafdf01464d67985/config/config.default.yaml#L111) in the configuration file.
Additionally, a similar bug is addressed so that solar rooftop potentials are only loaded again if it is included in [`renewable_carriers`](https://github.com/PyPSA/pypsa-eur/blob/1161a7c8e0f0e2952253df04cafdf01464d67985/config/config.default.yaml#L111).

## Checklist

### DIFF
--- a/scripts/add_existing_baseyear.py
+++ b/scripts/add_existing_baseyear.py
@@ -100,46 +100,47 @@ def add_existing_renewables(
     irena = irena.unstack().reset_index()
 
     for carrier, tech in tech_map.items():
-        if carrier in renewable_carriers:
-            df = (
-                irena[irena.Technology.str.contains(tech)]
-                .drop(columns=["Technology"])
-                .set_index("Country")
-            )
-            df.columns = df.columns.astype(int)
+        if carrier not in renewable_carriers:
+            continue
+        df = (
+            irena[irena.Technology.str.contains(tech)]
+            .drop(columns=["Technology"])
+            .set_index("Country")
+        )
+        df.columns = df.columns.astype(int)
 
-            # calculate yearly differences
-            df.insert(loc=0, value=0.0, column="1999")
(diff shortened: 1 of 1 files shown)


scripts/prepare_sector_network.py
```diff
@@ -1589,23 +1589,24 @@ def insert_electricity_distribution_grid(
     n.generators.loc[solar, "capital_cost"] = costs.at["solar-utility", "capital_cost"]
 
     fn = solar_rooftop_potentials_fn
-    potential = pd.read_csv(fn, index_col=["bus", "bin"]).squeeze()
-    potential.index = potential.index.map(flatten) + " solar"
+    if len(fn) > 0:
```
FabianHofmann, line 1592: just for context of this check: is fn an empty string if solar_rooftop is disabled?

daniel-rdt: It is actually an empty list (see [here](https://github.com/PyPSA/pypsa-eur/blob/478b7abcd846b6483f8379773676c776791e85bf/rules/build_sector.smk#L1409-L1413)), so `len(fn)` works with both.

daniel-rdt:
As @coroa rightfully pointed out [here](https://github.com/open-energy-transition/open-tyndp/pull/53#discussion_r2158704612), we can reduce the diff for these changes by following a
```
if carrier not in renewable_carriers:
   continue
```
convention in the for loop. I'll make the adjustment before we can merge

PR #1720, opened 2025-06-25
fix: make workflow compatible with optimization framework changes

FabianHofmann:
With PyPSA API changes coming from https://github.com/PyPSA/PyPSA/pull/1154 index names will be called "name" instead of "Bus"/"Generator" etc. This backwards compatible change account for this.

## Checklist

### DIFF
--- a/scripts/_helpers.py
+++ b/scripts/_helpers.py
@@ -29,7 +29,7 @@ logger = logging.getLogger(__name__)
 
 REGION_COLS = ["geometry", "name", "x", "y", "country"]
 
-PYPSA_NEW_API = bool(re.match(r"^0\.35\.0\.post1\.dev\d{3}", pypsa.__version__))
+PYPSA_V1 = bool(re.match(r"^0\.35\.0\.post1\.dev\d{3}", pypsa.__version__))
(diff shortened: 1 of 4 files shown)


scripts/_helpers.py
```diff
@@ -29,6 +29,8 @@
 
 REGION_COLS = ["geometry", "name", "x", "y", "country"]
 
+PYPSA_NEW_API = bool(re.match(r"^0\.35\.0\.post1\.dev\d{3}", pypsa.__version__))
```
lkstrp: I would call it `PYPSA_V1` even if this is the hacky `0.35post1.dev100` match right now

lkstrp (approved):
I suppose the parts where it's renamed to `'name'` without a v1 check, are consequent scripts on already renamed dataframes?

FabianHofmann:
> I suppose the parts where it's renamed to `'name'` without a v1 check, are consequent scripts on already renamed dataframes?

exactly

scripts/_helpers.py
FabianHofmann: done

PR #1732, opened 2025-07-03
refactor: remove deprecated old pinned env files

lkstrp:


fneum:
To merge after next release? The deprecation was not released yet, but fine to remove hard.

PR #1734, opened 2025-07-03
feat: add config validation and enable config key deletion

coroa:

We are encountering during extension of pypsa-eur in soft-forks that we cannot account for all possible config options. Therefore i am proposing a central place in maybe `scripts/config.py > validate_config` where one would check for consistency of configuration settings.

This `validate_config` function is called from `scenario_config` for config providers and for the initial config (at the top of `Snakefile`), it is expected to raise an exception when there is a conflict.

At the moment the function is a stub, a good schema with a detailed report of all the failed validations should be added before this is merged. But i would like to get a quick idea of whether you think a mechanism like that is valid? This might also be a good place for a jsonschema or pydantic model evaluation?

The second problem this PR is addressing is that the config is recursively pruned from `null` values so that one can remove keys from dictionaries in higher priority configs.

@tgilon @daniel-rdt

## Checklist

### DIFF
--- a/scripts/_helpers.py
+++ b/scripts/_helpers.py
@@ -201,7 +201,7 @@ def mute_print():
             yield
 
 
-def prune_config_deletes(config: dict) -> dict:
+def prune_config_deletes(config: dict) -> None:
     """
     Removes keys with `null` values to allow removing keys
     """
@@ -214,8 +214,6 @@ def prune_config_deletes(config: dict) -> dict:
 
     for key in remove:
         del config[key]
-
-    return config
 
 
 def set_scenario_config(snakemake):
--- a/rules/common.smk
+++ b/rules/common.smk
@@ -45,9 +45,7 @@ def scenario_config(scenario_name):
     merged = merge_configs(config, scenario[scenario_name])
     prune_config_deletes(merged)
     validate_config(merged)
-    return validate_config(
-        prune_config_deletes(merge_configs(config, scenarios[scenario_name]))
-    )
+    return merged
 
 
 def static_getter(wildcards, keys, default):


scripts/_helpers.py
```diff
@@ -214,6 +231,7 @@ def set_scenario_config(snakemake):
             with open(root_dir / scenario["file"]) as f:
                 scenario_config = yaml.safe_load(f)
         update_config(snakemake.config, scenario_config[snakemake.wildcards.run])
+        prune_config_deletes(snakemake.config)
```
coroa, line 232: maybe we would want to validate here as well? in principle the DAG creation should have validated already, so i decided against it.

lkstrp:
Ref #1547

scripts/_helpers.py
lkstrp: In theory, extending the pruning could make the configuration invalid, couldn't it? Which means it would be best to validate the configuration whenever it is changed

coroa: During DAG generation the same pruning is also applied and then validated, so i think it would still be fine.

coroa:
> Ref #1547

Very true, exactly the same concept, should have reviewed again. Only thing that is new here is null-value pruning and application to scenario_config.

What is the state there?

lkstrp:
The state is pretty much still as described in the PR. JSON Schema is quite verbose, difficult to set up, and unable to handle complicated validations. Since pydantic now supports full export to JSON Schema, I would switch to a minimal pydantic implementation, which we could then extend. The issue with the minimal implementation is that we cannot replace all the doc tables at once.

I'm not sure if pruning should be done with Pydantic as well. But could be moved also later. Otherwise, if you're only defining a placeholder here to have something for soft forks, go ahead

coroa:
> implementation, which we could then extend. The issue with the minimal implementation is that we cannot replace all the doc tables at once.
> 
> I'm not sure if pruning should be done with Pydantic as well. But could be moved also later. Otherwise, if you're only defining a placeholder here to have something for soft forks, go ahead

Thanks, indeed. 

This here is mainly thought of as a placeholder where soft-forks can add custom validation functions. Good to know about pydantic as new plan and i prefer the `lib/validation.py` placement you introduced over in the other PR.

Will adapt here.

fneum:
To be closed by #1912.

Issue #1731, opened 2025-07-03
Grid extraction from OSM fails in `clean_osm_data` due to shapely error


nesnoj:
Hey community,

not sure if this is a bug:
I'm trying to extract the transmission grid without running pypsa itself.

What I did:
- create `config.yaml` from `config.default.yaml`
- change `config.yaml`: `electricity` -> `base_network` to "osm-raw" to activate rule [`build_osm_network`](https://github.com/PyPSA/pypsa-eur/blob/v2025.04.0/rules/build_electricity.smk#L803)
- run `snakemake -j1 build_osm_network`

While rule [`clean_osm_data`](https://github.com/PyPSA/pypsa-eur/blob/v2025.04.0/rules/build_electricity.smk#L758) is running I get the following shapely exception right here in [`_extend_lines_to_substations()`](https://github.com/PyPSA/pypsa-eur/blob/v2025.04.0/scripts/clean_osm_data.py#L1551-L1556)

<details>

<summary>Click to expand log</summary>

```
[....]
INFO:__main__: - Importing cables 33/34: data/osm-raw/SK/cables_way.json
INFO:__main__: - Skipping cables 34/34 (empty): data/osm-raw/XK/cables_way.json
INFO:__main__:---
INFO:__main__:Dropping duplicate lines.
INFO:__main__:Dropped 535 duplicate elements. Keeping 344346 elements.
INFO:__main__:Dropping 9005 OSM ways (AC lines and cables) for their parent OSM relations.
INFO:__main__:Cleaning voltages.
INFO:__main__:Filtering dataframe by voltage. Only keeping rows above and including 220000 V.
INFO:__main__:Dropped 281372 elements with voltage below 220000. Keeping 53969 elements.
INFO:__main__:Cleaning circuits.
INFO:__main__:Cleaning cables.
INFO:__main__:Cleaning frequencies.
INFO:__main__:Cleaning wires.
INFO:__main__:Cleaning lines and determining circuits.
INFO:__main__:Setting circuits of remaining 5 lines to 1...
INFO:__main__:Lines affected: way/907802873-1; way/1340163909-1; way/1348591977; way/1348591978; way/905085926
INFO:__main__:Dropped 411 DC lines. Keeping 54155 AC lines.
INFO:__main__:Creating lines geometry.
INFO:__main__:Finalising lines column types.
INFO:__main__:Identifying and removing lines within substation polygons...
[...]

PR #1737, opened 2025-07-07
Bug fix for non-linestrings

bobbyxng:
Closes https://github.com/PyPSA/pypsa-eur/issues/1731

- Fixing bug that broke the `clean_osm_data` workflow
- Dropping geometries that are neither LineStrings nor MultiLineStrings
- (Minor changes to import of helper functions)

## Checklist

### DIFF
--- a/scripts/build_osm_network.py
+++ b/scripts/build_osm_network.py
@@ -11,7 +11,6 @@ import networkx as nx
 import numpy as np
 import pandas as pd
 import pypsa
-from _helpers import configure_logging, set_scenario_config
 from pyproj import Transformer
 from shapely import prepare
 from shapely.algorithms.polylabel import polylabel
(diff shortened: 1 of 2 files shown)


scripts/build_osm_network.py
```diff
@@ -11,15 +11,14 @@
 import numpy as np
 import pandas as pd
 import pypsa
+from _helpers import configure_logging, set_scenario_config
```
lkstrp: No, stick to absolute imports. If your `mock_snakemake` setup causes issues, just add the project directory to PATHS

bobbyxng: done!

nesnoj:
Thx @bobbyxng for the fix! The task now finishes successfully - I haven't checked the resulting model though..

scripts/build_osm_network.py
lkstrp: I still see a `from _helpers` instead of `from scripts._helpers` here

bobbyxng: My bad, merge conflict gone wrong :) It is now reverted to absolute paths.

PR #1742, opened 2025-07-07
Allow country-specific district heating potentials

cpschau:

This PR extends the `sector:district_heating:potential` config parameter to accept either a single float (as before) or a dictionary with country-specific district heating potentials.

* **Default behavior**: If no dictionary is provided, the parameter remains a float (default: `0.6`).
* **Country-specific potentials**: Users can supply a dictionary mapping country codes to potential values:

  * For **partial coverage**, the dictionary must include:

    * country codes for countries with specific potentials
    * a `'default'` key for all other countries as a fallback value
  * For **full coverage**, the dictionary must contain a value for each country code in the model.

 
## Checklist

### DIFF
--- a/scripts/build_district_heat_share.py
+++ b/scripts/build_district_heat_share.py
@@ -66,6 +66,20 @@ if __name__ == "__main__":
 
     # maximum potential of urban demand covered by district heating
     central_fraction = snakemake.config["sector"]["district_heating"]["potential"]
+    if isinstance(central_fraction, dict):
+        if pd.Index(pop_layout.ct.unique()).difference(central_fraction.keys()).any():
+            default_value = central_fraction.get("default")
+            if default_value is None:
+                raise ValueError(
+                    "No default district heating potential was provided in the config."
+                )
+            logger.warning(
+                "Some countries do not have a district heating potential defined. "
+                f"Using default value {default_value:.2%} for these countries."
+            )
+        else:
+            default_value = None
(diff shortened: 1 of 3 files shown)


scripts/build_district_heat_share.py
```diff
@@ -66,6 +66,25 @@
 
     # maximum potential of urban demand covered by district heating
     central_fraction = snakemake.config["sector"]["district_heating"]["potential"]
+    if isinstance(central_fraction, dict):
+        other_countries = set(pop_layout.ct.unique()).difference(
+            central_fraction.keys()
+        )
+        if other_countries:
```
amos-schledorn, line 74: Could we get one or two comments here explaining the logic?

scripts/build_district_heat_share.py
```diff
@@ -66,6 +66,25 @@
...
+                )
+            logger.warning(
+                "Some countries do not have a district heating potential defined. "
+                f"Using default value {default_value:.2%} for these countries."
+            )
+            central_fraction = {
+                **central_fraction,
+                **{ct: default_value for ct in other_countries},
+            }
+        central_fraction = pop_layout.ct.map(central_fraction)
```
amos-schledorn, line 91: Same

scripts/build_district_heat_share.py
cpschau: Thanks, I added some comments!

PR #1675, opened 2025-05-07
data: Intermediate layer for versioning of datasets

euronion:
This PR suggest an implementation method for an intermediate data layer, that allows to retrieve a specific version of a dataset, rather than always the latest.

The current status is implemented on the OSM dataset, as this one was already tracking versions and working very similarly. I'll expand it to two more datasets in the upcoming days.

## Motivation

Upstream data dependencies are not always versioned or fixed, meaning they may change unexpectedly without a way to revert to a different version. This causes reproducibility issues.

## Approach

This PR solves the problem in the following way:

* We create archived versions of all external datasets (if we are allowed -> question of licensing) on e.g. Zenodo
* The URL for retrieving each combination of (dataset x version) is stored in `data/versions.csv`. This allows us to switch to a different data plattform or provider if necessary; or use a versioned URL directly from a data provider if available.
* `data/versions.csv` also records the license and the description for the dataset. I plan on utilising this information to automatically create new versions of datasets and distribute the license text + metadata information based, as well as utilise this file for the [documentation here](https://pypsa-eur.readthedocs.io/en/latest/data_sources.html)
* I imagine that all externally retrieved data will get two rules: (1) a rule for retrieving the `upstream` version, which I called `source: "build"` in the `config.default.yaml`, and (2) a rule to retrieve an archived version of the data. Both rules yield the same files that are then consumed by the model.

## TODO

- [ ] Implement for two more datasets
- [ ] Helper and instructions for creating a new version of a dataset from `upstream` to `archive`
- [ ] Update documentation to utilise `data/versions.csv`

Comments are already welcome!

## Checklist

- [ ] I tested my contribution locally and it works as intended.
- [ ] Code and workflow changes are sufficiently documented.
- [ ] Changed dependencies are added to `envs/environment.yaml`.
- [ ] Changes in configuration options are added in `config/config.default.yaml`.
- [ ] Changes in configuration options are documented in `doc/configtables/*.csv`.
- [ ] Sources of newly added data are documented in `doc/data_sources.rst`.
- [ ] A release note `doc/release_notes.rst` is added.

PR #1752, opened 2025-07-11
feat: overwrite costs through file

tgilon:

This PR proposes a new way to overwrite costs. The current implementation is backwards compatible and still support both legacy implementations (through `costs:overwrites` and `costs:capital_cost`/`costs:marginal_cost`).

Modifications to the default technico-economic assumptions can now be enabled via the configuration using the  `costs:custom_costs` parameter (default value: `false`). If this is enabled, the modifications from the file `data/custom_costs.csv` are applied to the default `resources/costs_{planning_horizons}.csv` files. The `data/custom_costs.csv` file has the same structure as `resources/costs_{planning_horizons}.csv`. A new column, `planning_horizon`, has been added to allow values to be overwritten with time-varying values. Custom costs can target all planning horizons and / or technologies using the `all` identifier.

To maintain backwards compatibility, overwrites using the configuration file are still permitted and are applied as the final step in the processing. The precedence order is as follows:
1. Default cost assumptions from `resources/costs_{planning_horizons}.csv`
2. Modifications from `data/custom_costs.csv`
3. Overwrites from the configuration file using both `costs:overwrites` and `costs:capital_costs`/`costs:marginal_costs` (deprecated)

The resulting set of costs is written in a new file: `resources/costs_{planning_horizons}_prepped.csv`. This wide-format file lists  technologies as rows and paramaters as columns. It contains the final values to be used by the model. I believe this improves the transparency and user-friendliness.

## Worflow
* `retrieve_cost_data` retrieves the default cost assumptions across all planning horizons.
* (new) `build_cost_data`prepares and extends the default cost data with custom cost modifications.
    * Inputs: default costs assumptions from `resources/costs_{planning_horizons}.csv` and `data/custom_costs.csv`
    * Outputs: `resources/costs_{planning_horizons}_prepped.csv`
    * This script now includes the `load_costs` features. To maintain backwards compatibility, overwriting using the configuration file is supported and are applied as the final step of the processing.
* `load_costs` only reads the prepared files and has been moved to `scripts/_helpers.py`. This affects:
    * `add_electricity`
    * `prepare_network`
    *  `prepare_sector_network`
    *  `add_existing_baseyear` (myopic and perfect)
    *  `add_brownfield` (myopic and perfect)
    *  `solve_sector_network_myopic`
    * `solve_sector_network_perfect`
    * `make_summary_perfect`
* (new) `build_costs` collects all prepared cost assumptions.

## Test

The implementation was tested using the latest objective values from the `master` test: https://github.com/PyPSA/pypsa-eur/actions/runs/16173248041. 

```yaml
# config/config.electricity.yaml
solving:
    check_objective:
      enable: true
      expected_value: 3.4086454773e+07

# config/config.overnight.yaml
solving:
    check_objective:
      enable: true
      expected_value: 5.8513952542e+08

# config/config.perfect.yaml
solving:
    check_objective:
      enable: true
      expected_value: 1.5396533001e+10

# config/config.myopic.yaml (2030 only)
solving:
    check_objective:
      enable: true
      expected_value: 5.9022555338e+08
```

## Checklist


tgilon:
cc @euronion @daniel-rdt @coroa

config/config.default.yaml
```diff
@@ -923,20 +923,10 @@ costs:
...
-    offwind: 0.015
-    hydro: 0.
-    H2: 0.
-    electrolysis: 0.
-    fuel cell: 0.
-    battery: 0.
-    battery inverter: 0.
-    home battery storage: 0
-    water tank charger: 0.03
-    central water pit charger: 0.025
```
euronion, line 949: Have these values been previously applied in the default configuration? If yes, we should also keep them as defaults (maybe move them into the `data/custom_costs.csv` (like you did) and enable them by default with `custom_costs: enable: true`

data/custom_costs.csv
```diff
@@ -0,0 +1,13 @@
...
+all,offwind,marginal_cost,0.015,EUR/Mwh,Default,
+all,hydro,marginal_cost,0,EUR/Mwh,Default,
+all,H2,marginal_cost,0,EUR/Mwh,Default,
+all,electrolysis,marginal_cost,0,EUR/Mwh,Default,
+all,fuel cell,marginal_cost,0,EUR/Mwh,Default,
+all,battery,marginal_cost,0,EUR/Mwh,Default,
+all,battery inverter,marginal_cost,0,EUR/Mwh,Default,
+all,home battery storage,marginal_cost,0,EUR/Mwh,Default,
+all,water tank charger,marginal_cost,0.03,EUR/Mwh,Default,
+all,central water pit charger,marginal_cost,0.025,EUR/Mwh,Default,
```
euronion: * "Default" as "further description" is a bit unclear.
* The unit should be `MWh`, not `Mwh`

config/config.default.yaml
```diff
@@ -923,20 +923,10 @@ costs:
     lifetime: 25
     "CO2 intensity": 0
     "discount rate": 0.07
+  custom_costs:
+    enable: false
+    file: data/custom_costs.csv
```
euronion: Rather than having the path in the config file, I'd prefer to have it in the `Snakefile` (like for `custom_powerplants.csv`) - keep filenames out of the `config` as much as possible.

rules/build_electricity.smk
```diff
@@ -575,6 +575,31 @@ rule build_hac_features:
...
+        costs=config_provider("costs"),
+        max_hours=config_provider("electricity", "max_hours"),
+        snapshots=config_provider("snapshots"),
+        drop_leap_day=config_provider("enable", "drop_leap_day"),
+    input:
+        costs=resources("costs_{planning_horizons}.csv"),
+    output:
+        resources("costs_{planning_horizons}_prepped.csv"),
+    wildcard_constraints:
+        planning_horizons=r"\d{4}",  # Constrain to 4-digit years only
```
euronion: The wildcard has already been restricted [here](https://github.com/PyPSA/pypsa-eur/blob/67f0919ad47c77863d29ba6d8a3aeba9854b8c6d/Snakefile#L54), it shouldn't be necessary to repeat the definition here.

config/config.default.yaml
tgilon: Correct, it was default. I've changed the default to true.

data/custom_costs.csv
tgilon: I've adjusted the unit. The "Default" is written in the `source` column. I've changed it to "Default value to prevent mathematical degeneracy in optimisation". Good for you?

config/config.default.yaml
tgilon: I've introduced the path here to make it dependent of the scenario. Would that work for you?

tgilon:
> I like the structure, nicely done!
> 
> Not so happy about the name `costs_{planning_horizons}_prepped.csv`.

Thank you for the feedback! I'm not a fan of `prepped` either. Couldn't find a better name but I would be happy to change it.

data/custom_costs.csv
euronion: Yes!

config/config.default.yaml
euronion: I see - yes, since it is for scenario dependency, that should be fine.

euronion:
> > I like the structure, nicely done!
> > Not so happy about the name `costs_{planning_horizons}_prepped.csv`.
> 
> Thank you for the feedback! I'm not a fan of `prepped` either. Couldn't find a better name but I would be happy to change it.

Maybe move it original data into `data/costs_{year}.csv` and then have in `resources("costs_{year}.csv)` the processed version named the same way, i.e. `data` for raw data inputs, `resources` for processed stuff.

Not sure why we moved `costs_{year}.csv` into `resources(...)`. Maybe @FabianHofmann can help?

rules/build_electricity.smk
tgilon: You're right. I fixed an ambiguity in Snakemake between `retrieve_cost_data` and `build_cost_data` in the wrong way. I've added a constraint for `{year}`.

tgilon:
> Maybe move it original data into `data/costs_{year}.csv` and then have in `resources("costs_{year}.csv)` the processed version named the same way, i.e. `data` for raw data inputs, `resources` for processed stuff.
> 
> Not sure why we moved `costs_{year}.csv` into `resources(...)`. Maybe @FabianHofmann can help?

No clue but that's fine for me.

fneum:
It was moved because it can depend on config settings. All files that depend on config settings need to go in resources.

euronion:
> It was moved because it can depend on config settings. All files that depend on config settings need to go in resources.

Thanks for the explanation off-list!

@tgilon : Then let's keep the `costs_*.csv` files in `resources(...)`. How about `costs_{planning_horizon}_processed.csv` instead of `prepped`?

tgilon:
> Thanks for the explanation off-list!
> 
> @tgilon : Then let's keep the `costs_*.csv` files in `resources(...)`. How about `costs_{planning_horizon}_processed.csv` instead of `prepped`?

That works for me as well. I also considered `costs_{planning_horizon}_custom.csv`. However, I don't have a strong preference. Happy to use whatever the community prefers.

daniel-rdt:
> > Thanks for the explanation off-list!
> > @tgilon : Then let's keep the `costs_*.csv` files in `resources(...)`. How about `costs_{planning_horizon}_processed.csv` instead of `prepped`?
> 
> That works for me as well. I also considered `costs_{planning_horizon}_custom.csv`. However, I don't have a strong preference. Happy to use whatever the community prefers.

`costs_{planning_horizon}_custom.csv` seems very similar to the newly introduced `custom_costs.csv` which could be confusing. I feel like `_processed` works? This also does not imply that any custom changes are added since this is not case if the `custom_costs.csv` and overwrites are empty, so the costs are only processed

coroa:
Hi @fneum, @euronion ,

i understand from @tgilon that all comments were addressed. Could we have a final review here before it's rebased on top of the latest changes and then make it ready for merging!

Thanks

doc/configtables/costs.csv
```diff
@@ -3,6 +3,9 @@ year,--,YYYY; e.g. '2030',Year for which to retrieve cost assumptions of ``resou
 version,--,vX.X.X or <user>/<repo>/vX.X.X; e.g. 'v0.5.0',Version of ``technology-data`` repository to use. If this string is of the form <user>/<repo>/<version> then costs are instead retrieved from ``github.com/<user>/<repo>`` at the <version> tag.
 social_discountrate,p.u.,float,Social discount rate to compare costs in different investment periods. 0.02 corresponds to a social discount rate of 2%.
 fill_values,--,float,Default values if not specified for a technology in ``resources/costs.csv``.
+custom_costs
+-- enable,bool,true or false, Enable overwriting default techno-economic assumptions with custom values defined in {file} (default value: ``data/custom_costs.csv``).
```
euronion:
```suggestion
-- enable,bool,true or false,Enable overwriting default techno-economic assumptions with custom values defined in {file} (default value: ``data/custom_costs.csv``).

```

doc/release_notes.rst
```diff
@@ -11,6 +11,8 @@ Release Notes
 Upcoming Release
 ================
 
+* Feature: Introduce a new method to overwrite costs (https://github.com/PyPSA/pypsa-eur/pull/1752). Modifications to the default technico-economic assumptions can now be enabled in the configuration using the `costs:custom_costs` setting (default value: `false`). When enabled, modifications specified in the file `data/custom_costs.csv` are applied to the default `resources/costs_{planning_horizons}.csv` files. The implementation maintains backwards compatibility with existing approaches via `costs:overwrites` and `costs:capital_cost`/`costs:marginal_cost` parameters.
```
euronion: My understanding was that we are deprecating `costs:overwrites` in favour of this new overwrite file?

rules/build_electricity.smk
```diff
@@ -575,6 +575,29 @@ rule build_hac_features:
         "../scripts/build_hac_features.py"
 
 
+rule build_cost_data:
```
euronion: For reasons of interactions with #1675 , can we call the rule and scripts `process_cost_data` instead?
(The convention we want to introduce is:
* `build_X` = build primary data from scratch
* `retrieve_X` = retrieve primary data from archive

and here you are processing primary data that was retrieved (the technology-data), so that's why `build_` would not be consistent.

It also fits more nicely with the file postfix `_processed` that you are adding :)

Snakefile
```diff
@@ -52,6 +52,7 @@ wildcard_constraints:
     opts=r"[-+a-zA-Z0-9\.]*",
     sector_opts=r"[-+a-zA-Z0-9\.\s]*",
     planning_horizons=r"[0-9]{4}",
+    year=r"[0-9]{4}",
```
euronion: I don't see this wildcard constraint being used anywhere?

rules/build_electricity.smk
```diff
@@ -575,6 +575,29 @@ rule build_hac_features:
...
+        costs=resources("costs_{planning_horizons}.csv"),
+    output:
+        resources("costs_{planning_horizons}_processed.csv"),
+    log:
+        logs("build_cost_data_{planning_horizons}.log"),
+    benchmark:
+        benchmarks("build_cost_data_{planning_horizons}")
+    threads: 1
+    resources:
+        mem_mb=10000,
```
euronion: Necessary or excessive?

rules/collect.smk
```diff
@@ -12,6 +12,19 @@ localrules:
     solve_sector_networks,
 
 
+rule build_costs:
```
euronion:
```suggestion
rule process_costs:
```

Rename suggestion as earlier

rules/build_sector.smk
```diff
@@ -1346,9 +1345,11 @@ rule prepare_sector_network:
             "biomass_potentials_s_{clusters}_{planning_horizons}.csv"
         ),
         costs=lambda w: (
-            resources("costs_{}.csv".format(config_provider("costs", "year")(w)))
+            resources(
+                "costs_{}_processed.csv".format(config_provider("costs", "year")(w))
```
euronion: Can we write this consistentnly as in `rules/build_electricity.smk` without `.format(...)`?

rules/solve_myopic.smk
```diff
@@ -21,7 +21,7 @@ rule add_existing_baseyear:
         busmap=resources("busmap_base_s_{clusters}.csv"),
         clustered_pop_layout=resources("pop_layout_base_s_{clusters}.csv"),
         costs=lambda w: resources(
-            "costs_{}.csv".format(
+            "costs_{}_processed.csv".format(
```
euronion: Can we write this consistentnly as in `rules/build_electricity.smk` without `.format(...)`?

rules/solve_perfect.smk
```diff
@@ -19,7 +19,7 @@ rule add_existing_baseyear:
         busmap=resources("busmap_base_s_{clusters}.csv"),
         clustered_pop_layout=resources("pop_layout_base_s_{clusters}.csv"),
         costs=lambda w: resources(
-            "costs_{}.csv".format(
+            "costs_{}_processed.csv".format(
```
euronion: Can we write this consistentnly as in `rules/build_electricity.smk` without `.format(...)`?

rules/solve_perfect.smk
```diff
@@ -104,7 +104,7 @@ rule solve_sector_network_perfect:
         network=resources(
             "networks/base_s_{clusters}_{opts}_{sector_opts}_brownfield_all_years.nc"
         ),
-        costs=resources("costs_2030.csv"),
+        costs=resources("costs_2030_processed.csv"),
```
euronion, line 105: Hard-coded year intentional? Doesn't seem right

rules/solve_perfect.smk
```diff
@@ -146,7 +146,7 @@ def input_networks_make_summary_perfect(w):
 rule make_summary_perfect:
     input:
         unpack(input_networks_make_summary_perfect),
-        costs=resources("costs_2020.csv"),
+        costs=resources("costs_2020_processed.csv"),
```
euronion, line 147: Hard-coded year intentional? Doesn't seem right. Also inconsistent with the `2030` earlier that was hard-coded

scripts/_helpers.py
```diff
@@ -1079,3 +1079,21 @@ def load_cutout(
...
+    cost_file : str
+        Path to the CSV file containing cost data
+
+    Returns
+    -------
+    costs : pd.DataFrame
+        DataFrame containing the prepared cost data
+    """
+
+    return pd.read_csv(cost_file, index_col=0).sort_index()
```
euronion: `.sort_index()` is on the auto-generated numerical index, right? Do you want to introduce/sort by a more meaningful order instead? e.g. technology name + parameter name?

scripts/add_electricity.py
```diff
@@ -1176,14 +1064,7 @@ def attach_stores(
...
 
-    Nyears = n.snapshot_weightings.objective.sum() / 8760.0
-
-    costs = load_costs(
-        snakemake.input.tech_costs,
-        params.costs,
-        max_hours,
-        Nyears,
-    )
+    costs = load_costs(snakemake.input.tech_costs)
```
euronion: Maybe a good opportunity to harmonise the naming of the cost file in all Snakefiles?
```suggestion
    costs = load_costs(snakemake.input.costs)
```

(consistent with e.g. `add_existing_baseyear.py`; could also instead do e.g. `snakemake.input.tech_assumptions`, but `costs` is fine.)

scripts/add_existing_baseyear.py
```diff
@@ -745,12 +746,7 @@ def add_heating_capacities_installed_before_baseyear(
     spatial = define_spatial(n.buses[n.buses.carrier == "AC"].index, options)
     add_build_year_to_new_assets(n, baseyear)
 
-    Nyears = n.snapshot_weightings.generators.sum() / 8760.0
```
euronion, line 748: Was this functionality replaced somehow?

scripts/build_cost_data.py
```diff
@@ -0,0 +1,247 @@
...
+    return costs
+
+
+if __name__ == "__main__":
+    if "snakemake" not in globals():
+        from _helpers import mock_snakemake
+
+        snakemake = mock_snakemake("build_cost_data", planning_horizons=2030)
+
+    config = snakemake.params.costs
```
euronion: I'd prefer it if we don't call `params` as `config`, because `config` is usually something else.

scripts/build_cost_data.py
```diff
@@ -0,0 +1,247 @@
...
+    )
+    nyears = len(snapshots) / 8760
+    planning_horizon = str(snakemake.wildcards.planning_horizons)
+
+    # Retrieve costs assumptions
+    costs = pd.read_csv(snakemake.input.costs)
+    custom_costs = (
+        pd.read_csv(
+            snakemake.params.costs["custom_costs"]["file"],
+            dtype={"planning_horizon": "object"},
```
euronion:
```suggestion
            dtype={"planning_horizon": "str"},
```

Is probably what you wanted

scripts/process_cost_data.py
euronion, line 1: Small style nit:

For `pandas` I prefer seeing accessing of column values using `df[...]` notation and accessing of row values using `df.loc[...]` notation, e.g.:

```python
# org
specific_entries = custom_costs[custom_costs["technology"] != "all"]

# my preference
specific_entries = custom_costs.loc[custom_costs["technology"] != "all"]
```

`pandas` has a special alias in this case, that allows your code to work, but it is inconsistent with the remaning syntax. If you don't mind adjusting your code to use `.loc` when accessing `rows`, I'd be grateful!

scripts/build_cost_data.py
```diff
@@ -0,0 +1,247 @@
...
+    expanded_df = pd.merge(
+        all_entries,
+        available_technologies,
+        on="_merge_key",
+    ).drop("_merge_key", axis=1)
+
+    # Remove duplicates based on technology and parameter, keeping the last (specific entries)
+    custom_costs_expanded = pd.concat(
+        [expanded_df, specific_entries], ignore_index=True
+    ).drop_duplicates(subset=["technology", "parameter"], keep="last")
```
euronion: This should implicitly give priority to specific entries over the `all` entries.
Is that intentional and what you want?

scripts/build_cost_data.py
```diff
@@ -0,0 +1,247 @@
...
+        .assign(_merge_key=1)
+    )
+    specific_entries = custom_costs[custom_costs["technology"] != "all"]
+
+    if all_entries.empty:
+        return specific_entries
+
+    # Perform cartesian product merge to expand all technologies
+    available_technologies = (
+        costs[["technology"]].drop_duplicates().assign(_merge_key=1)
```
euronion:
```suggestion
        costs[["technology"]].unique().assign(_merge_key=1)
```

I find that a bit easier to read.

scripts/build_cost_data.py
```diff
@@ -0,0 +1,247 @@
...
+    )
+    expanded_df = pd.merge(
+        all_entries,
```
euronion:
```suggestion
    expanded_df = all_entries.merge(
```

`df.merge` is prefered over `pd.merge(df, ...`

scripts/build_cost_data.py
```diff
@@ -0,0 +1,247 @@
...
+    custom_costs = (
+        pd.read_csv(
+            snakemake.params.costs["custom_costs"]["file"],
+            dtype={"planning_horizon": "object"},
+        )
+        .query("planning_horizon in [@planning_horizon, 'all']")
+        .drop("planning_horizon", axis=1)
+    )
+
+    if config["custom_costs"].get("enable", False) and not custom_costs.empty:
```
euronion: IIRC we wanted at some point remove all `<config>.get(...)` entries to avoid unintentional default behaviour. Not sure if that ever took off, but how about:
```suggestion
    if not config["custom_costs"]:
```

which fail if the key `custom_costs` is not in the `configfile` (i.e. the person did upgrade their file) and it will not skip this part if the option is enabled but `custom_costs` is empty (also unexpected)

scripts/prepare_network.py
```diff
@@ -308,12 +309,7 @@ def set_line_nom_max(
 
     n = pypsa.Network(snakemake.input[0])
     Nyears = n.snapshot_weightings.objective.sum() / 8760.0
```
euronion, line 311: Functionality for `Nyears` removed?

euronion (changes requested):
Thanks @tgilon , a few comments. Some more important, some only about style. Happy to also receive input from @coroa or @fneum on some of them about general aspects.

doc/configtables/costs.csv
martacki: with file  either `str` or `None`, we could avoid an extra setting for `enable`

rules/build_electricity.smk
```diff
@@ -575,6 +575,29 @@ rule build_hac_features:
...
+rule build_cost_data:
+    params:
+        costs=config_provider("costs"),
+        max_hours=config_provider("electricity", "max_hours"),
+        snapshots=config_provider("snapshots"),
+        drop_leap_day=config_provider("enable", "drop_leap_day"),
+    input:
+        costs=resources("costs_{planning_horizons}.csv"),
+    output:
+        resources("costs_{planning_horizons}_processed.csv"),
```
martacki, line 587: also with the upper suggestion this is more sparse in changes (no need to update which `costs.csv` to use in all future rules) - we don't need to save both the original costs & the updated one. seems unnecessary (to me), but maybe I'm missing sth?

scripts/build_cost_data.py
martacki: @tgilon I think this way would be a bit more efficient / readable, but you'd need to re-index 

https://github.com/open-energy-transition/pypsa-wal/blob/764415fe77a355f6cbf8ac99f74425be10109c8a/scripts/walloon_scripts/update_costs.py#L39

Snakefile
martacki: It was used in `retrieve_cost_data`, but I replaced it now with `planning_horizons`... not sure the reason. Seems to work. It was also used in the new `build_costs` rule. But I think in both cases, `planning_horizons` makes more sense.

scripts/build_cost_data.py
martacki: changed the whole structure of this, effectively removing priotization

rules/build_electricity.smk
martacki: Wrong, we still need it -- unless we want to change sth in a retrieve script, which I think is a but ugly...

rules/build_electricity.smk
martacki: imo we could also add a function within the existing retrieve_cost_data , and update the retrieved cost with the custom_costs right behind, just using a logger.info.

Would be consistent with how the powerplants are being built & updated.
Also, the retrieval of cost data is currently a very short script

martacki: update: kept it there to avoid making changes to data that is purely retrieved.

scripts/build_cost_data.py
martacki: which would be an indirect default behaviour...? ;) anyway, solved this differently

rules/solve_perfect.smk
martacki: I think it's intentional if I look at the outputs -- I guess it starts with 2030 & does perfect foresight then for all other years, using other costs... I think this might be an issue for another PR, in case this is a bug

scripts/_helpers.py
martacki: not needed for this functionality

scripts/add_electricity.py
martacki: good idea - luckily not many `tech_costs` appearances

scripts/process_cost_data.py
martacki: not sure if I resolved this / what the request is

scripts/prepare_network.py
martacki: Not sure what you mean

doc/release_notes.rst
martacki: Yes, this is why it's just maintaining backward compatibility for sth that'll be deprecated soon... ?

euronion:
```suggestion
* Feature: Introduce a new method to overwrite costs (https://github.com/PyPSA/pypsa-eur/pull/1752). Modifications to the default technico-economic assumptions can now be enabled in the configuration using the `costs:custom_costs` setting (default value: `false`). When enabled, modifications specified in the file `data/custom_costs.csv` are applied to the default `resources/costs_{planning_horizons}.csv` files. The existing implementation via `costs:overwrites` and `costs:capital_cost`/`costs:marginal_cost` parameters is still available but will be deprecated in the near future.
```

Ok, then I suggest to make it add the deprecation note also here.

doc/configtables/costs.csv
euronion: Good idea, but there's a cross-interaction to account for. See other comment.

config/config.default.yaml
```diff
@@ -932,20 +932,8 @@ costs:
     "CO2 intensity": 0
     "discount rate": 0.07
     "standing losses": 0
+  custom_cost_fn: data/custom_costs.csv # fn or None
```
euronion:
```suggestion
  custom_cost_fn: data/custom_costs.csv
```

euronion: No comments in config file for cleanness :)

data/custom_costs.csv
euronion, line 1: nit: "Default value to prevent mathematical degeneracy in optimisation" should go into the `further_description` column, not the `source` column

doc/configtables/costs.csv
```diff
@@ -3,6 +3,7 @@ year,--,YYYY; e.g. '2030',Year for which to retrieve cost assumptions of ``resou
 version,--,vX.X.X or <user>/<repo>/vX.X.X; e.g. 'v0.5.0',Version of ``technology-data`` repository to use. If this string is of the form <user>/<repo>/<version> then costs are instead retrieved from ``github.com/<user>/<repo>`` at the <version> tag.
 social_discountrate,p.u.,float,Social discount rate to compare costs in different investment periods. 0.02 corresponds to a social discount rate of 2%.
 fill_values,--,float,Default values if not specified for a technology in ``resources/costs.csv``.
+custom_cost_fn,str or None,Path to the custom costs file or None if it should not be used.
```
euronion:
```suggestion
custom_cost_fn,--,str or None,Path to the custom costs file. None if it should not be used. Default ``data/custom_costs.csv`` contains minor adjustments for stabilising the optimisation results.
```

rules/build_electricity.smk
euronion:
```suggestion
rule process_cost_data:
```

To address my comment above. Seperating retrieval and modification (powerplants are a special case, should address that in the future)

rules/build_electricity.smk
```diff
@@ -575,6 +575,29 @@ rule build_hac_features:
...
 
 
+rule build_cost_data:
+    params:
+        costs=config_provider("costs"),
+        max_hours=config_provider("electricity", "max_hours"),
+        snapshots=config_provider("snapshots"),
+        drop_leap_day=config_provider("enable", "drop_leap_day"),
+    input:
+        costs=resources("costs_{planning_horizons}.csv"),
```
euronion, line 584: Add `custom_costs_fn` as optional input that is either the file or None. All file dependencies should be exposed in the Snakefiles

Snakefile
euronion: Yes, using `planning_horizon` makes more sense to me as well!

rules/solve_perfect.smk
euronion: Opened separate bug report.

scripts/prepare_network.py
euronion: `Nyears` is far as I understand it used here if the model spans more than a year - annualised costs then need to be adjusted to represent `Nyear > 1`. Looking at the new code, it does not keep this functionality and always assumes the model spanning a single year.

scripts/process_cost_data.py
```diff
@@ -0,0 +1,197 @@
...
+    nyears = len(snapshots) / 8760
+    planning_horizon = str(snakemake.wildcards.planning_horizons)
+
+    # Retrieve costs assumptions
+    # If the index is the unique pair (technology, parameter),
+    # it can be easily overwritten.
+    costs = pd.read_csv(snakemake.input.costs, index_col=[0, 1])
+    if snakemake.input.costs is not None:
+        custom_costs = pd.read_csv(
+            snakemake.params.costs["custom_cost_fn"],
```
euronion: See other comment: Needs to be a `snakemake.input["custom_costs"]`, not a `snakemake.params`.

rules/build_electricity.smk
euronion: Also rename the script please.

scripts/process_cost_data.py
```diff
@@ -0,0 +1,197 @@
...
+    return costs
+
+
+if __name__ == "__main__":
+    if "snakemake" not in globals():
+        from _helpers import mock_snakemake
+
+        snakemake = mock_snakemake("build_cost_data", planning_horizons=2030)
+
+    config = snakemake.config["costs"]
```
euronion:
```suggestion
    cost_params = snakemake.params["costs"]
```

I meant the comment the other way around: Keep using the `params`, but don't call them `config` ;)
(need to rename other occurrences in script)

scripts/prepare_network.py
euronion: It confuses me a bit: The `nyears` are (with slightly different code) considered in the `build_cost_data.py`, but they are also considered here. Is this correct? Or is maybe the inclusion here obsolete/double, as the processing for `Nyears` now happens in the cost processing rule?

scripts/process_cost_data.py
euronion: You (unintentionally?) addressed it! :)

rules/build_electricity.smk
euronion: Yeah. We still need it and we want to go to a system where we have:
* unmodified data in `data` (great naming, right?)
* modified data in `resources`

Different file names here help with clarity.

So current code is fine

euronion (changes requested):
Two major things to do:
* Make input file dependency explicit (instead of using a the `params` for the input file
* Clarify/check on `Nyears` functionality, whether it is applied properly / in some cases maybe twice.

rules/build_electricity.smk
martacki: Wanted to do that, but it seemed like `config_provider()` is only used in params?

scripts/prepare_network.py
martacki: It's still used e.g. for creating the co2 limit, gaslimit, etc.

doc/release_notes.rst
martacki: That's at least my understanding! @tgilon can you confirm?

data/custom_costs.csv
martacki: I'm not sure -- I do agree partially, but it also suggests that the source is some random developers mind?

scripts/prepare_network.py
martacki: I didn't see it used accidentally twice within the costs functionality - but I'll double-check that later.

scripts/process_cost_data.py
```diff
@@ -0,0 +1,199 @@
...
+        from _helpers import mock_snakemake
+
+        snakemake = mock_snakemake("build_cost_data", planning_horizons=2030)
+
+    cost_params = snakemake.params["costs"]
+
+    snapshots = get_snapshots(
+        snakemake.params.snapshots, snakemake.params.drop_leap_day, tz="UTC"
+    )
+    nyears = len(snapshots) / 8760
```
euronion: We modify the annualised investment cost by the number of years. In other scripts, `Nyears` was calculated as

```python
years = n.snapshot_weightings.generators.sum() / 8760.0
```

which seems to as if it would be the correct formula here as well, because it returns a year even if we run with e.g. 6H resolution.

(maybe even put it into a `_helpers.py` function... we use it a lot.)

scripts/prepare_sector_network.py
```diff
@@ -6187,11 +6187,7 @@ def add_import_options(
     nhours = n.snapshot_weightings.generators.sum()
     nyears = nhours / 8760
 
-    costs = load_costs(
-        snakemake.input.costs,
-        snakemake.params.costs,
-        nyears=nyears,
-    )
+    costs = load_costs(snakemake.input.costs)
```
euronion, line 6191: Update: I thought this is the place where the `Nyears` are double-counted for the costs, but looking at it with a fresh mind again it looks fine.

If for a last check you could look at the costs that are in a `network` with and without your PR, that'd be enough for last check for me! :)

martacki: I checked the `costs.csv` output files after processing, `prepare_network` and `prepare_sector` network, and they are all the same.

I also checked the netcdf files comparing master branch to this PR, and they also look the same to me (except small changes in `marginal_costs` that are likely due to `noisy_costs` parameter which I forgot to disable - `investment_costs` / `capital_costs` are the same with an accuracy of `1e-9`)

scripts/process_cost_data.py
martacki: I fixed it locally for now - definitely second putting it into a `_helpers` script, however I'm not sure when to use generators, objective etc. so this might need more thorough thinking on architecture. for now, I think all the 3 columns are always the same, right?

martacki: also we don't have a network here, so I just calculated it using a `timedelta` instead

euronion: Hm, I'm not sure how stable that is. It looks fine. Maybe @fneum can give a thumbs up/down?
For generalising it, the network `n` could be an argument to `load_costs(...)`, which then weights the annualised costs based on the passed network.

martacki: I know, just wanted to avoid loading the network in. ;) but if it's more stable maybe worth it

euronion: Certainly easier to maintain if we use the same approach consistently everywhere!

martacki: adjusted

doc/release_notes.rst
tgilon: Also, the release note is no longer up to date. You've changedd the config structure.

rules/build_electricity.smk
```diff
@@ -575,14 +575,14 @@ rule build_hac_features:
         "../scripts/build_hac_features.py"
 
 
-rule build_cost_data:
+rule process_cost_data:
```
tgilon, line 578: This doesn't follow the standard PyPSA-Eur prefixes.

tgilon: And if we agree on this change, we will need to change the definitions of `log`, `benchmark` and `mock`.

Snakefile
```diff
@@ -52,7 +52,6 @@ wildcard_constraints:
     opts=r"[-+a-zA-Z0-9\.]*",
     sector_opts=r"[-+a-zA-Z0-9\.\s]*",
     planning_horizons=r"[0-9]{4}",
-    year=r"[0-9]{4}",
```
tgilon: I see other uses of `{year}` in the code. It would probably be unsafe to remove it.

scripts/process_cost_data.py
```diff
@@ -209,39 +163,35 @@ def costs_for_storage(store, link1, link2=None, max_hours=1.0):
...
 
-    config = snakemake.params.costs
+    cost_params = snakemake.params["costs"]
 
-    snapshots = get_snapshots(
-        snakemake.params.snapshots, snakemake.params.drop_leap_day, tz="UTC"
-    )
-    nyears = len(snapshots) / 8760
+    n = pypsa.Network(snakemake.input.network)
+    nyears = n.snapshot_weightings.generators.sum() / 8760.0
```
tgilon, line 166: Is this really an improvement? It creates an unnecessary dependency on the base network, that could be avoided.

scripts/process_cost_data.py
```diff
@@ -28,59 +28,13 @@
...
 import pandas as pd
-from _helpers import get_snapshots
+import pypsa
 
 from scripts.add_electricity import calculate_annuity
 
 logger = logging.getLogger(__name__)
 
 
-def expand_all_technologies(
```
tgilon: I think that this feature is useful and it shouldn't be removed.

scripts/process_cost_data.py
```diff
@@ -209,39 +163,35 @@ def costs_for_storage(store, link1, link2=None, max_hours=1.0):
...
+    if snakemake.input.custom_costs is not None:
+        custom_costs = pd.read_csv(
+            snakemake.input.custom_costs,
+            dtype={"planning_horizon": "str"},
+            index_col=[1, 2],  # 0 is planning_horizon
+        ).query("planning_horizon in [@planning_horizon, 'all']")
+        custom_costs = custom_costs.drop("planning_horizon", axis=1)
+
+        # Overwrite unique pairs of (technology, parameter)
+        missing_idx = custom_costs.index.difference(costs.index)
```
tgilon, line 198: Leaner but conflicts with the all tech feature.

data/custom_costs.csv
tgilon: Same view as Martha

scripts/process_cost_data.py
```diff
@@ -0,0 +1,197 @@
...
+"""
+Prepare and extend default cost data with custom cost modifications. Custom costs can target all planning horizons and / or technologies using the 'all' identifier.
+
+Preparing the cost data includes:
+- aligning all units to conventional units (i.e. MW / MWh),
+- filling in missing data,
+- computing 'capital_cost' parameter (annualised investment costs and FOM),
+- computing 'marginal_cost' parameter (fuel costs and VOM),
+- computing storage costs for batteries and hydrogen,
+- (deprecated) overwriting attributes using config-based modifications.
```
tgilon, line 13: Depending on the decision, this needs to be updated as well.

scripts/process_cost_data.py
```diff
@@ -0,0 +1,197 @@
+# SPDX-FileCopyrightText: Contributors to PyPSA-Eur <https://github.com/pypsa/pypsa-eur>
+#
+# SPDX-License-Identifier: MIT
+"""
+Prepare and extend default cost data with custom cost modifications. Custom costs can target all planning horizons and / or technologies using the 'all' identifier.
```
tgilon, line 5: If the all tech feature must be removed, this needs to be removed.

doc/release_notes.rst
```diff
@@ -9,6 +9,8 @@ Release Notes
...
```
tgilon: Building on @euronion' suggestion

```suggestion
* Feature: Introduce a new method to overwrite costs (https://github.com/PyPSA/pypsa-eur/pull/1752). Modifications to the default technico-economic assumptions can now be configured via `costs:custom_cost_fn`, which applies changes to the `resources/costs_{planning_horizons}.csv` files. The default configuration includes minor adjustments to stabilize optimization results. The existing implementation via `costs:overwrites` and `costs:capital_cost`/`costs:marginal_cost` parameters remains available but will be deprecated in a future release.
```

rules/build_electricity.smk
euronion: In #1675 we harmonise `retrieve_` and `build_` prefixes, with the meaning of `build_` here not fitting in, that's why I suggested `process_` as kind-of new prefix. Do you have another suggestion?

scripts/process_cost_data.py
euronion: The number of snapshots is not the same as the snapshot weightings, so the meaning of both options is different. If we weight the annualised costs, we need the information from the network somehow.

rules/build_electricity.smk
tgilon: I think it's a broader discussion indeed. Probably not something that should block this PR specifically. I've already seen `clean` and `extract` as two other candidates.

scripts/process_cost_data.py
tgilon: Ok, got it, tks

Snakefile
martacki: the year one comes from the config though, not as a wildcard... or did I oversee sth?

scripts/process_cost_data.py
martacki: does it? where are the conflicts?

scripts/process_cost_data.py
martacki: Can you briefly explain the benefits of it? Given that it wasn't needed before + creates a lot of code to maintain, I'm not sure of the benefits -- I thought this was mainly needed in case custom_costs and costs have different indices, which I resolved with the leaner version you commented on below

scripts/process_cost_data.py
martacki: ah yes. It does. sorry for overseeing. Yes it's leaner, and I think easier to maintain.

scripts/process_cost_data.py
tgilon: An example would be if we wanted to enforce a fixed annuity factor or lifetime across the entire config.

Snakefile
tgilon: The rules `retrieve_seawater_temperature` and `retrieve_hera_data` use `year` as a wildcard. Correct?

scripts/process_cost_data.py
tgilon: "Conflict" was probably too strong. My main concern is ensuring the compatibility with the `all` tech feature.

scripts/process_cost_data.py
martacki: How many technologies are normally missing in the costs? If not in the costs file at this stage but needed later, this would throw an error, wouldn't it? At the moment it's only the marginal_costs that are not in the costs.csv, but they are not subject to annuity

Snakefile
martacki: No, in `retrieve_hera_data` it's not a wildcard but a variable taken from `n.snapshots`.

I changed it for `seawater_temperature` to `planning_horizons`.

scripts/process_cost_data.py
martacki: I think it should be compatible -- the `missing_idx` should be empty after expanding costs to all techs. Not sure though if `expand_all_technologies` adds also missing parameters, such as `marginal_costs`  -- in that case, we'd still need this snippet

scripts/process_cost_data.py
tgilon: It's not about missing technologies, but rather about setting a value for all the techs present in the cost file using a single line in `custom_costs`.

What about this? It's probably easier to maintain and it follows the logic that you introduced.

```python
all_techs = custom_costs.loc["all"]
custom_costs = custom_costs.query("technology != 'all'")

# Overwrite unique pairs

# Overwrite all techs entries
costs = costs.sort_index()
for param in all_techs.index:
    costs.loc[pd.IndexSlice[:, param], all_techs.columns] = all_techs.loc[param].values
```

Snakefile
tgilon: Are we both referring to [this](https://github.com/PyPSA/pypsa-eur/blob/master/rules/retrieve.smk#L763-L783)? I can't see the `n.snapshots` thing.

scripts/process_cost_data.py
tgilon: @martacki This is probably better.

```python
all_techs = custom_costs.query("technology=='all'")
custom_costs = custom_costs.query("technology != 'all'")

# Overwrite unique pairs

# Overwrite all techs entries
if not all_techs.empty:
    all_techs = all_techs.droplevel(0)
    costs = costs.sort_index()
    for param in all_techs.index:
        costs.loc[pd.IndexSlice[:, param], all_techs.columns] = all_techs.loc[param].values

```

martacki: thanks, added!

scripts/process_cost_data.py
martacki: we kept it, thanks for the lean(er) implementation!

scripts/process_cost_data.py
martacki: I guess this is also resolved

tgilon:
@martacki What are the next steps here to get this ready?

euronion:
> @martacki What are the next steps here to get this ready?

fneum wanted to review today, we'll probably have some more comments to address then.

Snakefile
fneum: @tgilon is right. Should stay.

scripts/process_cost_data.py
fneum: Information on the snapshot weightings is essential. In principle, you could run non-consecutive / non-contiguous years and in that case working with time deltas like @martacki won't work.

I agree that loading the network here is clumsy. A general alternative could be to store the snapshot weightings in a separate `resource/*.csv`, which might be worth discussing if there are several such cases where only snapshot information is needed.

Actually, I think at this stage all weightings are still one, so it's just about the number of uniformly weighted snapshots.

doc/release_notes.rst
```diff
@@ -9,6 +9,8 @@ Release Notes
 Upcoming Release
 ================
 
+* Feature: Introduce a new method to overwrite costs (https://github.com/PyPSA/pypsa-eur/pull/1752). Modifications to the default technico-economic assumptions can now be configured via `costs:custom_cost_fn`, which applies changes to the `resources/costs_{planning_horizons}.csv` files. The default configuration includes minor adjustments to stabilize optimization results. The existing implementation via `costs:overwrites` and `costs:capital_cost`/`costs:marginal_cost` parameters remains available but will be deprecated in a future release.
```
fneum:
```suggestion
* Feature: Introduce a new method to overwrite costs (https://github.com/PyPSA/pypsa-eur/pull/1752). Modifications to the default techno-economic assumptions can now be configured via `costs:custom_cost_fn`, which applies changes to the `resources/costs_{planning_horizons}.csv` files. The default configuration includes minor adjustments to stabilize optimization results. The existing implementation via `costs:overwrites` and `costs:capital_cost`/`costs:marginal_cost` parameters remains available but will be deprecated in a future release.
```

scripts/_helpers.py
```diff
@@ -1113,3 +1113,21 @@ def load_cutout(
...
```
fneum:
```suggestion
    return pd.read_csv(cost_file, index_col=0)
```

rules/solve_perfect.smk
fneum: Yes, fine to ignore here. The perfect foresight optimization needs an overhaul.

rules/retrieve.smk
```diff
@@ -732,9 +732,9 @@ if config["enable"]["retrieve"]:
         params:
             default_cutout=config_provider("atlite", "default_cutout"),
         output:
-            seawater_temperature="data/seawater_temperature_{year}.nc",
+            seawater_temperature="data/seawater_temperature_{planning_horizons}.nc",
```
fneum:
```suggestion
            seawater_temperature="data/seawater_temperature_{year}.nc",
```
No, using `planning_horizons` here would be confusing because the wildcard refers to a weather year.

rules/retrieve.smk
```diff
@@ -732,9 +732,9 @@ if config["enable"]["retrieve"]:
...
         log:
-            "logs/retrieve_seawater_temperature_{year}.log",
+            "logs/retrieve_seawater_temperature_{planning_horizons}.log",
```
fneum:
```suggestion
            "logs/retrieve_seawater_temperature_{year}.log",
```
No, using `planning_horizons` here would be confusing because the wildcard refers to a weather year.

rules/collect.smk
```diff
@@ -12,6 +12,19 @@ localrules:
...
+    input:
+        lambda w: (
+            [resources(f"costs_{config_provider('costs', 'year')(w)}_processed.csv")]
+            if config_provider("foresight")(w) == "overnight"
+            else expand(
+                resources("costs_{planning_horizons}_processed.csv"),
+                **config["scenario"],
+                run=config["run"]["name"],
+            )
+        ),
```
fneum, line 25: The "overnight" case also needs the `{run}` wildcard passed. At the moment it's just for the `else` block.

scripts/retrieve_seawater_temperature.py
```diff
@@ -87,13 +87,13 @@
...
-            f"Downloading seawater temperature data for year {snakemake.wildcards.year}"
+            f"Downloading seawater temperature data for year {snakemake.wildcards.planning_horizons}"
         )
 
         _ = copernicusmarine.subset(
             dataset_id="cmems_mod_glo_phy_my_0.083deg_P1D-m",  # Global ocean physics reanalysis
-            start_datetime=f"{snakemake.wildcards.year}-01-01",
-            end_datetime=f"{int(snakemake.wildcards.year)}-12-31",
+            start_datetime=f"{snakemake.wildcards.planning_horizons}-01-01",
+            end_datetime=f"{int(snakemake.wildcards.planning_horizons)}-12-31",
```
fneum: Should be `{year}` as it doesn't refer to the `{planning_horizon}`.
```suggestion
            start_datetime=f"{snakemake.wildcards.year}-01-01",
            end_datetime=f"{int(snakemake.wildcards.year)}-12-31",
```

scripts/retrieve_seawater_temperature.py
```diff
@@ -87,13 +87,13 @@
         # Spatial coverage: European waters (-12°W to 42°E, 33°N to 72°N)
         # Depth range: 5-15m (suitable for heat pump intake depths)
         logger.info(
-            f"Downloading seawater temperature data for year {snakemake.wildcards.year}"
+            f"Downloading seawater temperature data for year {snakemake.wildcards.planning_horizons}"
```
fneum:
```suggestion
            f"Downloading seawater temperature data for year {snakemake.wildcards.year}"
```

Snakefile
tgilon: Added back

rules/collect.smk
tgilon: Fixed

PR #1764, opened 2025-07-21
code: Add empty folders to repo

euronion:
I added the folders that will be automatically created by `snakemake` as empty folders with `.gitkeep`.

The idea is to avoid folder "popping up" when exeucting the workflow. It also makes it easier to talk about those folders in the documentation, because one doesn't always have to add "this folder will be created automatically".

## Checklist


fneum:
It kind of conflicts with the purge rule which deletes resources and results folders. If we go ahead with this, the purge rule should be adjusted to delete the content of resources and results directories (rather than directories themselves).

lkstrp (changes requested):
I am a bit against `.gitkeep`s files, since they indicate git functionality when they are just random files. They also just hide the information if a directory actually contains anything.

Is there any other benefit then just for reference in the documentation? We could use a simple file tree instead. It's more clear anyway

euronion:
There are some minor benefits (like tab completion for the folders) and downsides (like how the folders show up in the editor (w/ content and w/o content). It also makes creating the file trees easier - you don't need to manually edit them to add those folders.

My main concern was them "popping up". From my experience, this behaviour can be a bit surprising for people that are not used to `snakemake`. When they read the documentation and read about folders, and then compare it with the folders they see in their file explorer that don't match the documentation - that's were confusion starts to settle in.

Happy to adjust the `purge` rule if we want to go forward with it.

fneum:
I don't have a strong preference on this matter. @euronion, if you'd like, go ahead and adjust the purge rule.

euronion:
I opted for the simple approach of recreating the folder with the `.gitkeep` files, i.e. this does not preserve the creation timestamps of the folders and the `.gitkeep` files.

euronion:
@lkstrp Can we merge without CI?

PR #1748, opened 2025-07-09
feat:allocate heat pump capex on heat bus

amos-schledorn:
replaces #1714

Issue #1728, opened 2025-07-01
Implement TES temperature boosting as constraints


TomKae00:
## Describe the feature you'd like to see

In PR #1711, we added PTES temperature boosting via resistive heaters and air-sourced heat pumps directly in prepare_sector_network.py.
After discussing with @amos-schledorn, we’ve decided to move this logic into solve_sector_network.py as explicit constraints.
This refactor keeps prepare_sector_network.py clear and makes it easier to generalize boosting for other TES: TTES, ATES

The config forwarding fix for `heat_pump_cop_approximation` is handled separately in PR #1729 .

PR #1765, opened 2025-07-21
Refactor: Implementation of ptes temperature boosting constraints

TomKae00:

Addresses issue #1728.
Incorporates #1748 (with minor fixes).

This PR introduces **temperature boosting** for Pit Thermal Energy Storage (PTES) during both **discharging** (to meet district heating (DH) forward temperature) and **charging** (to raise PTES top temperature and increase usable capacity). Boosting is modeled via linear constraints that couple PTES flows with booster technologies—**resistive heaters**, **air-source heat pumps (ASHP)** for charging and discharging, and a **PTES-sourced heat pump** for discharge boosting.

Additionally, a new class **`TesTemperatureMode`** enables selecting the PTES **top-temperature regime**:
- **`constant`** — keeps the PTES top temperature fixed at the configured maximum.
- **`dynamic`** — tracks the DH forward setpoint (a function of ambient temperature) **until** the setpoint exceeds the PTES maximum; beyond that, the top temperature is clamped at the maximum.

**Model assumptions**
- PTES **max top temperature** `T_top,max` = **90 °C**.  
- PTES **bottom temperature** `T_bottom` = **35 °C**.  
- Both **charge** and **discharge** boosting may be enabled to enforce physically consistent temperature levels.  
- When `charge_boosting_required = true`, `temperature_profile = dynamic` is **not** permitted (the forward temperature is boosted up to `T_top,max`, so the dynamic rule would collapse to the constant case). Use `temperature_profile = constant` in that situation.
- The **charge boosting ratio** is clipped at **1.0** to prevent heat venting, since booster technologies can also be used to charge the PTES.

---

## Boosting ratios

Let

- $\dot V$ … volumetric flow rate (assumed constant),
- $\rho$ … fluid density (assumed constant),
- $c_p$ … specific heat capacity (assumed constant),
- $T_{\text{return}}$ … DH return temperature,
- $T_{\text{fwd}}$ … DH forward temperature setpoint,
- $T_{\text{store}}$ … PTES top temperature at the considered time step,
- $T_{\text{top,max}}$ … configured maximum PTES top temperature,
- $T_{\text{bottom}}$ … PTES bottom temperature.

Generic heat-rate form:

$$
Q = \dot V\ \cdot\ \rho\ \cdot\ c_p\ \cdot\ \Delta\ T
$$

### Discharging (meeting the DH forward temperature)
The **discharge boosting ratio** $\gamma_d$ quantifies the extra lift required from boosters relative to what PTES already supplies:

$$
\gamma_d \equiv \frac{Q_{\text{boost}}}{Q_{\text{discharge}}}
= \frac{T_{\text{fwd}} - T_{\text{store}}}{T_{\text{store}} - T_{\text{return}}}
$$

A value of **0** means **no** boosting ($T_{\text{store}}\ge T_{\text{fwd}}$); larger values imply more boosting is required.

### Charging (raising the PTES top temperature)
To fill the storage from the DH **return** up to the **maximum top temperature**, split the total energy as

Define the **charge boosting ratio** $\gamma_c$ as

$$
\gamma_c \equiv \frac{Q_{\text{boost}}}{Q_{\text{charge}}}
= \frac{T_{\text{top,max}} - T_{\text{fwd}}}{T_{\text{fwd}} - T_{\text{return}}}
$$

Wherever $T_{\text{fwd}} \ge T_{\text{top,max}}$, we set $\gamma_c = 0$ since no further boost is needed.

---

## Constraints
We derive a **direct-utilisation indicator** from the boosting ratio,

$$
\phi(\gamma) = \frac{\gamma}{\gamma+\varepsilon}
$$

with small $\varepsilon>0$. It acts as a smoothed 0/1 **gate** in the constraints: if no boosting is needed ($\gamma=0$), then $\phi=0$ and the RHS vanishes—disabling the coupling so dispatch variables remain free; if boosting is required ($\gamma>0$), $\phi>0$ activates the coupling. The $\varepsilon$ term is only for numerical stability and has no physical meaning.

### Discharging constraint
For **discharger** link $d$:

$$
(COP_{ptes} - 1) \cdot p_{ptes} + \sum_{b \in B} p_b \cdot\ \beta_b \ge\ \phi(\gamma_d)\ \cdot\ p_d
$$

**PTES-HP source cap.** Each PTES-sourced heat pump draws its thermal source from the PTES discharge, so its boost contribution is capped by the portion of discharge that actually requires boosting. For every $h \in H$:

$$
(COP_{ptes} - 1)\cdot p_{ptes} \le \phi(\gamma_d)\cdot p_d
$$

Where:
- $B$: set of booster technologies; $p_b$ their dispatch; $\beta_b$ their **discharge** boosting factor.
- $p_d$, $\gamma_d$: discharger dispatch and its boosting ratio profile  .

**PTES-HP accounting.** In `prepare_sector_network`, the PTES-sourced heat pump is represented as a **link converting 1 unit of electric power to 1 unit of heat** (because its thermal source is the PTES). The energy balance follows

$$
Q_{\text{out}} = Q_{\text{source}} + P_{\text{el}}
$$

where the $Q_{\text{source}}$ term is simultaneously accounted by the **PTES discharger** delivering heat directly to the DH system. This split preserves the correct PTES energy balance for HP operation.

### Charging constraint
For each **charger** link $c$:

$$
\sum_{b \in B} p_b \cdot\ \beta_b \ge\ \phi(\gamma_c)\ \cdot\ p_c
$$

- $B$: set of booster technologies; $p_b$ their dispatch; $\beta_b$: **charge** boosting ratio  
- $p_c$, $\gamma_c$: charger dispatch and its boosting ratio profile  

---

## Limitations & future work

- **PTES-HP sourcing:** The PTES-sourced heat pump can, in principle, exploit temperatures **below 35 °C**. A more detailed implementation could pre-heat DH return via PTES (heat exchanger) and let the HP provide the remaining lift via mixing—expanding usable storage capacity.  
- **Top-temperature trajectory:** The PTES top temperature is likely **time-varying** in practice; extending beyond `constant`/`dynamic` to a more realistic trajectory model would improve fidelity.  
- **Selective charge boosting:** Not all technologies should require charge boosting. Distinguish **high-temperature** sources (e.g., resistive heaters, CHP) that can **charge directly** at $T_{\text{top,max}}$ from **low-temperature** sources (e.g., geothermal at ~65 °C) that either cannot charge directly or would require boosting.

---

## Testing

```yaml
scenario:
  clusters:
  - 16
  planning_horizons:
  - 2030

foresight: overnight
countries: ['DE', 'PL', 'AT' ,'DK']

sector:
  district_heating:
    ptes:
      enable: true
      dynamic_capacity: false
      discharge_boosting_required: true #false
      charge_boosting_required: true #false
      temperature_profile: 'constant'
      booster_technologies: [ "resistive heater", "air heat pump", "water pits heat pump" ]
      max_top_temperature: 90
      min_bottom_temperature: 35
    ttes:
      enable: true
  heat_pump_sources:
    urban central:
      - water pits
      - air
      - geothermal

clustering:
  temporal:
    resolution_sector: 6h
```

## Results 

---

**Total system cost**

| Scenario                                  | Total cost [€ bn] | Δ vs. no boosting [€ bn] | Δ vs. no boosting [%] |
|-------------------------------------------|------------------:|--------------------------:|----------------------:|
| No charge & no discharge boosting         | 169.94            | –                         | –                     |
| Charge **and** discharge boosting         | 170.21            | 0.27                      | 0.16%                 |
| Discharge boosting only (no charge boost) | 169.98            | 0.04                      | 0.03%                 |

**Scenario overviews**

_No charge & no discharge boosting_

overview_master_PL0_0

_Charge & discharge boosting (“bidiboost”)_

overview_bidiboost_PL0_0

_Discharge boosting only_

overview_discharge_PL0_0

**Interpretation**

- With **no boosting**, PTES can be used freely year-round; **resistive heaters** cover the required charging.
- Under **charge & discharge boosting**, resistive-heater capacity is unchanged, but an **air-source heat pump** is additionally built to provide winter boosting (notably **November–December**).
- With **discharge-only boosting**, boosting is handled solely by the **resistive heater**.  
-  The current **PTES-sourced HP** configuration is not competitive against resistive heaters / ASHP for boosting in these runs. If **bottom-temperature cooling** were allowed **only for PTES-HP** (as in practice) and implemented as outlined in *Future work*, relative competitiveness could shift.

**Heat balances**

boost_compare

PTES remains present across scenarios. In **master** (no boosting) and **discharge-only** cases, PTES usage is stronger; with **bidiboost**, PTES capacity declines and **TTES** becomes relatively more favored, though PTES is still built.

## Checklist

### DIFF
--- a/rules/solve_overnight.smk
+++ b/rules/solve_overnight.smk
@@ -2,6 +2,7 @@
 #
 # SPDX-License-Identifier: MIT
 
+
 def ptes_operation_profiles(w):
     """
     Return a dict of only the PTES profiles that are enabled in config,
--- a/scripts/build_ptes_operations/ptes_temperature_approximator.py
+++ b/scripts/build_ptes_operations/ptes_temperature_approximator.py
@@ -2,25 +2,28 @@
 #
 # SPDX-License-Identifier: MIT
 
-import xarray as xr
 from enum import Enum
+
+import xarray as xr
+
 
 class TesTemperatureMode(Enum):
     """
     TES temperature profile assumptions.
-    
+
     CONSTANT: Assumes fixed temperatures at operational limits.
         - Top temperature: constant at max_top_temperature
         - Bottom temperature: constant at min_bottom_temperature
         - Assumes charge-boosting to maintain top temperature.
         - NOTE: Assuming bottom_temperature = min_bottom_temperature ignores that cooling of the return temperature might be necessary in practice.
-        
+
     DYNAMIC: Assumes temperatures follow network conditions.
         - Top temperature: follows forward_temperature (clipped at max_top_temperature)
         - Bottom temperature: follows return_temperature
         - Does not assume charge-boosting.
         - Note: This ignores that the TES temperatures do not match the supply temperatures due to thermal losses or other factors.
     """
+
     CONSTANT = "constant"
     DYNAMIC = "dynamic"
 
@@ -112,7 +115,9 @@ class PtesTemperatureApproximator:
                 self.max_top_temperature,
             )
         else:
-            raise NotImplementedError(f"Temperature profile {self.temperature_profile} not implemented")
+            raise NotImplementedError(
+                f"Temperature profile {self.temperature_profile} not implemented"
+            )
 
     @property
     def bottom_temperature(self) -> xr.DataArray:
@@ -129,7 +134,9 @@ class PtesTemperatureApproximator:
         elif self.temperature_profile == TesTemperatureMode.DYNAMIC:
             return self.return_temperature
         else:
-            raise NotImplementedError(f"Temperature profile {self.temperature_profile} not implemented")
+            raise NotImplementedError(
+                f"Temperature profile {self.temperature_profile} not implemented"
+            )
 
     @property
     def e_max_pu(self) -> xr.DataArray:
@@ -186,9 +193,10 @@ class PtesTemperatureApproximator:
             The resulting fraction of PTES charge that must be further heated.
         """
         if self.discharge_boosting_required:
-            return ((self.forward_temperature - self.top_temperature) / (
-                self.top_temperature - self.bottom_temperature
-            )).where(self.forward_temperature > self.top_temperature, 0)
+            return (
+                (self.forward_temperature - self.top_temperature)
+                / (self.top_temperature - self.bottom_temperature)
+            ).where(self.forward_temperature > self.top_temperature, 0)
         else:
             return xr.zeros_like(self.forward_temperature)
 
@@ -229,8 +237,13 @@ class PtesTemperatureApproximator:
             The fraction of the PTES's available storage capacity already used.
         """
         if self.charge_boosting_required:
-            return ((self.max_top_temperature - self.forward_temperature) / (
-                self.forward_temperature - self.return_temperature
-            )).where(self.forward_temperature < self.max_top_temperature, 0).clip(max=1)
+            return (
+                (
+                    (self.max_top_temperature - self.forward_temperature)
+                    / (self.forward_temperature - self.return_temperature)
+                )
+                .where(self.forward_temperature < self.max_top_temperature, 0)
+                .clip(max=1)
+            )
         else:
             return xr.zeros_like(self.forward_temperature)
--- a/scripts/build_ptes_operations/run.py
+++ b/scripts/build_ptes_operations/run.py
@@ -55,8 +55,8 @@ Approximate thermal energy storage (TES) top temperature and identify need for s
 import logging
 
 import xarray as xr
-from scripts._helpers import set_scenario_config
 
+from scripts._helpers import set_scenario_config
 from scripts.build_ptes_operations.ptes_temperature_approximator import (
     PtesTemperatureApproximator,
     TesTemperatureMode,
@@ -76,15 +76,18 @@ if __name__ == "__main__":
 
     set_scenario_config(snakemake)
 
-    if (snakemake.params.charge_boosting_required and
-            TesTemperatureMode(snakemake.params.ptes_temperature_profile) is TesTemperatureMode.DYNAMIC):
+    if (
+        snakemake.params.charge_boosting_required
+        and TesTemperatureMode(snakemake.params.ptes_temperature_profile)
+        is TesTemperatureMode.DYNAMIC
+    ):
         raise ValueError(
             "Charger boosting cannot be used with 'dynamic' temperature profile"
         )
 
     # Load temperature profiles
     logger.info(
-        f"Loading district heating temperature profiles and approximating PTES temperatures"
+        "Loading district heating temperature profiles and approximating PTES temperatures"
     )
     logger.info(
         f"PTES configuration: temperature_profile={snakemake.params.ptes_temperature_profile}, "
@@ -103,7 +106,9 @@ if __name__ == "__main__":
         ),
         max_top_temperature=snakemake.params.max_ptes_top_temperature,
         min_bottom_temperature=snakemake.params.min_ptes_bottom_temperature,
-        temperature_profile=TesTemperatureMode(snakemake.params.ptes_temperature_profile),
+        temperature_profile=TesTemperatureMode(
+            snakemake.params.ptes_temperature_profile
+        ),
         charge_boosting_required=snakemake.params.charge_boosting_required,
         discharge_boosting_required=snakemake.params.discharge_boosting_required,
         dynamic_capacity=snakemake.params.dynamic_capacity,
@@ -116,7 +121,7 @@ if __name__ == "__main__":
     ptes_temperature_approximator.e_max_pu.to_netcdf(
         snakemake.output.ptes_e_max_pu_profile
     )
-    
+
     ptes_temperature_approximator.boost_per_discharge.to_netcdf(
         snakemake.output.boost_per_discharge_profile
     )
--- a/scripts/prepare_sector_network.py
+++ b/scripts/prepare_sector_network.py
@@ -2972,7 +2972,8 @@ def add_heat(
                     p_nom_extendable=True,
                     marginal_cost=costs.at["water tank charger", "marginal_cost"],
                     lifetime=costs.at[
-                        heat_system.central_or_decentral + " water tank storage", "lifetime"
+                        heat_system.central_or_decentral + " water tank storage",
+                        "lifetime",
                     ],
                 )
 
@@ -2989,12 +2990,14 @@ def add_heat(
                     ],
                     p_nom_extendable=True,
                     lifetime=costs.at[
-                        heat_system.central_or_decentral + " water tank storage", "lifetime"
+                        heat_system.central_or_decentral + " water tank storage",
+                        "lifetime",
                     ],
                 )
 
                 n.links.loc[
-                    nodes + f" {heat_system} water tanks charger", "energy to power ratio"
+                    nodes + f" {heat_system} water tanks charger",
+                    "energy to power ratio",
                 ] = energy_to_power_ratio_water_tanks
 
                 n.add(
@@ -3015,12 +3018,15 @@ def add_heat(
                         "capital_cost",
                     ],
                     lifetime=costs.at[
-                        heat_system.central_or_decentral + " water tank storage", "lifetime"
+                        heat_system.central_or_decentral + " water tank storage",
+                        "lifetime",
                     ],
                 )
 
-            if heat_system == HeatSystem.URBAN_CENTRAL and options["district_heating"]["ptes"]["enable"
-                ]:
+            if (
+                heat_system == HeatSystem.URBAN_CENTRAL
+                and options["district_heating"]["ptes"]["enable"]
+            ):
                 n.add("Carrier", f"{heat_system} water pits")
 
                 n.add(
@@ -3076,9 +3082,7 @@ def add_heat(
                 # (flat ones if dynamic capacity disabled)
                 e_max_pu_data = xr.open_dataarray(ptes_e_max_pu_file)
                 e_max_pu = (
-                    e_max_pu_data.sel(name=nodes)
-                    .to_pandas()
-                    .reindex(index=n.snapshots)
+                    e_max_pu_data.sel(name=nodes).to_pandas().reindex(index=n.snapshots)
                 )
 
                 n.add(
@@ -3254,7 +3258,6 @@ def add_heat(
                     )
 
             if heat_source in {tes_system.value for tes_system in TesSystem}:
-
                 n.add(
                     "Link",
                     nodes,
@@ -3298,8 +3301,7 @@ def add_heat(
                 bus1=nodes,
                 carrier=f"{heat_system} resistive heater",
                 efficiency=1 / costs.at[key, "efficiency"],
-                capital_cost=costs.at[key, "capital_cost"]
-                * overdim_factor,
+                capital_cost=costs.at[key, "capital_cost"] * overdim_factor,
                 p_max_pu=0,
                 p_min_pu=-1,
                 p_nom_extendable=True,
--- a/scripts/solve_network.py
+++ b/scripts/solve_network.py
@@ -43,8 +43,6 @@ import yaml
 from pypsa.descriptors import get_activity_mask
 from pypsa.descriptors import get_switchable_as_dense as get_as_dense
 
-from scripts.definitions.tes_system import TesSystem
-
 from scripts._benchmark import memory_logger
 from scripts._helpers import (
     PYPSA_V1,
@@ -53,6 +51,7 @@ from scripts._helpers import (
     set_scenario_config,
     update_config_from_wildcards,
 )
+from scripts.definitions.tes_system import TesSystem
 
 logger = logging.getLogger(__name__)
 
@@ -1015,7 +1014,9 @@ def add_discharge_boosting_constraints(
     tes_heat_pumps = {f"{t.value} heat pump" for t in TesSystem}
 
     cop = xr.open_dataarray(cop_profiles_file).rename({"name": "node"})
-    ptes_boost_per_discharge_dataarray = xr.open_dataarray(boost_per_discharge_profile_file).rename({"name": "node"})
+    ptes_boost_per_discharge_dataarray = xr.open_dataarray(
+        boost_per_discharge_profile_file
+    ).rename({"name": "node"})
 
     # --- RHS: ptes discharger
     ptes_discharger_links = n.links.index[
@@ -1025,13 +1026,14 @@ def add_discharge_boosting_constraints(
         ptes_discharger_links.str.extract(r"^(\S+\s+\S+)")[0].to_numpy(),
         coords={"Link": ptes_discharger_links},
         dims="Link",
-        name="node"
+        name="node",
     )
 
-    rhs_base = p.sel(Link=ptes_discharger_links).groupby(ptes_discharger_node_da).sum("Link")
+    rhs_base = (
+        p.sel(Link=ptes_discharger_links).groupby(ptes_discharger_node_da).sum("Link")
+    )
     alpha_d = (
-        ptes_boost_per_discharge_dataarray
-        .sel(node=rhs_base.coords["node"])
+        ptes_boost_per_discharge_dataarray.sel(node=rhs_base.coords["node"])
         .sel(time=snapshot)
         .transpose("snapshot", "node")
     )
@@ -1041,14 +1043,17 @@ def add_discharge_boosting_constraints(
     lhs = None
     for tech in ptes_booster_technologies:
         booster_technologies_links = n.links.index[
-            n.links.index.str.contains("urban central") & n.links.index.str.contains(tech)
+            n.links.index.str.contains("urban central")
+            & n.links.index.str.contains(tech)
         ]
         if booster_technologies_links.empty:
             if tech in tes_heat_pumps:
                 raise ValueError(
                     f"Booster technology {tech} not found: it may not be listed as a heat_pump_source for urban central heating"
                 )
-            raise ValueError(f"No links found for booster technology '{tech}', check if the component exists")
+            raise ValueError(
+                f"No links found for booster technology '{tech}', check if the component exists"
+            )
 
         booster_tech_nodes_da = xr.DataArray(
             booster_technologies_links.str.extract(r"^(\S+\s+\S+)")[0].to_numpy(),
@@ -1056,7 +1061,11 @@ def add_discharge_boosting_constraints(
             dims="Link",
             name="node",
         )
-        expr_base = p.sel(Link=booster_technologies_links).groupby(booster_tech_nodes_da).sum("Link")
+        expr_base = (
+            p.sel(Link=booster_technologies_links)
+            .groupby(booster_tech_nodes_da)
+            .sum("Link")
+        )
 
         if tech in tes_heat_pumps:
             cop_heat_pump_da = (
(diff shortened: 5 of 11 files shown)


config/config.default.yaml
```diff
@@ -514,12 +514,15 @@ sector:
...
       relative_annual_temperature_reduction: 0.01
     ptes:
+      enable: true
       dynamic_capacity: true
-      supplemental_heating:
-        enable: false
-        booster_heat_pump: false
+      booster_technologies: ["resistive heater", ] #"air heat pump", ptes
+      storage_temperature_boosting: false  # postheating
+      forward_temperature_boosting: false # preheating
```
amos-schledorn: Let's move config instructions to the docs

rules/solve_myopic.smk
```diff
@@ -126,6 +126,41 @@ rule solve_sector_network_myopic:
...
+        ),
+        ptes_direct_utilisation_profiles= lambda w: (
+            resources(
+                "ptes_direct_utilisation_profiles_s_{clusters}_{planning_horizons}.nc"
+            )
+            if config_provider(
+                "sector","district_heating","ptes","storage_temperature_boosting",
+            )(w)
+            else []
+        ),
```
cpschau: Would maybe unpack these inputs from an external function to keep the rule definition a little bit leaner.

rules/solve_overnight.smk
```diff
@@ -15,6 +15,41 @@ rule solve_sector_network:
...
+        ),
+        ptes_direct_utilisation_profiles= lambda w: (
+            resources(
+                "ptes_direct_utilisation_profiles_s_{clusters}_{planning_horizons}.nc"
+            )
+            if config_provider(
+                "sector","district_heating","ptes","storage_temperature_boosting",
+            )(w)
+            else []
+        ),
```
cpschau: Same here

scripts/build_ptes_operations/ptes_temperature_approximator.py
```diff
@@ -106,3 +109,100 @@ def e_max_pu(self) -> xr.DataArray:
...
+            Raw forward temperature profile.
+
+        Returns
+        -------
+        xr.DataArray
+            Effective forward temperature for PTES.
+        """
+        if self.forward_temperature_boosting:
+            return forward_temperature.clip(min=self.max_ptes_top_temperature)
+        return forward_temperature
```
cpschau: I do not quite understand how this is meant to work. The effective forward temperature is only used to compute the top temperature property of the PTES. The top temperature of the PTES is either kept constant at `max_ptes_top_temperature` or changing according to #1612. Why does it depend on the forward temperature boosting?

scripts/build_ptes_operations/ptes_temperature_approximator.py
```diff
@@ -106,3 +109,100 @@ def e_max_pu(self) -> xr.DataArray:
...
+        """
+        return ((self.forward_temperature - self.return_temperature) / (
+            self.max_ptes_top_temperature - self.forward_temperature
+        )).where(self.forward_temperature < self.max_ptes_top_temperature, 0)
+
+    def _get_effective_forward_temperature(
+        self, forward_temperature: xr.DataArray
+    ) -> xr.DataArray:
+        """
+        Return the forward profile, flooring values below max_ptes_top_temperature
```
cpschau: Values are rather rounded up by clipping, so flooring could be a misleading term.

scripts/build_ptes_operations/run.py
```diff
@@ -97,7 +102,7 @@
         snakemake.output.ptes_top_temperature_profiles
     )
 
-    # if snakemake.params.enable_supplemental_heating:
+    # if snakemake.params.storage_temperature_boosting:
```
cpschau: Commented lines should be removed in final clean-up

scripts/build_ptes_operations/run.py
```diff
@@ -116,3 +121,25 @@
     ptes_temperature_approximator.e_max_pu.to_netcdf(
         snakemake.output.ptes_e_max_pu_profiles
     )
+
+    # if snakemake.params.storage_temperature_boosting:
```
cpschau: Same here

scripts/prepare_sector_network.py
```diff
@@ -1576,10 +1577,10 @@ def insert_electricity_distribution_grid(
     n.links.loc[v2gs, "bus1"] += " low voltage"
 
     hps = n.links.index[n.links.carrier.str.contains("heat pump")]
-    n.links.loc[hps, "bus0"] += " low voltage"
+    n.links.loc[hps, "bus1"] += " low voltage"
 
     rh = n.links.index[n.links.carrier.str.contains("resistive heater")]
-    n.links.loc[rh, "bus0"] += " low voltage"
+    n.links.loc[rh, "bus1"] += " low voltage"
```
cpschau, line 1583: Bus order for resistive heaters is not affected by #1748 or is this planned @amos-schledorn?

scripts/prepare_sector_network.py
```diff
@@ -3228,16 +3229,18 @@ def add_heat(
...
-                    capital_cost=costs.at[costs_name_heat_pump, "efficiency"]
-                    * costs.at[costs_name_heat_pump, "capital_cost"]
+                    efficiency=(1 / cop_heat_pump.clip(lower=0.001)).replace(1000, 0),
+                    efficiency2=1
+                    - (1 / cop_heat_pump.clip(lower=0.001)).replace(1000, 0),
+                    capital_cost=costs.at[costs_name_heat_pump, "capital_cost"]
                     * overdim_factor,
                     p_nom_extendable=True,
+                    p_max_pu=-cop_heat_pump / cop_heat_pump.clip(lower=0.001),
+                    p_min_pu=0,
```
cpschau: This is also part of the other PR #1748 , but I would prefer to swap values of `p_min_pu` and `p_max_pu` to avoid `p_min_pu` > `p_max_pu` .

scripts/prepare_sector_network.py
```diff
@@ -3263,41 +3266,21 @@ def add_heat(
...
-                    efficiency=(-(cop_heat_pump - 1)).clip(upper=0),
-                    efficiency2=cop_heat_pump,
-                    capital_cost=costs.at[costs_name_heat_pump, "efficiency"]
-                    * costs.at[costs_name_heat_pump, "capital_cost"]
+                    efficiency=1,
+                    capital_cost=costs.at[costs_name_heat_pump, "capital_cost"]
                     * overdim_factor,
                     p_nom_extendable=True,
+                    p_max_pu=0,
+                    p_min_pu=-(1 / cop_heat_pump).where(cop_heat_pump > 0, 0),
```
cpschau: Same here

scripts/prepare_sector_network.py
```diff
@@ -3306,13 +3289,14 @@ def add_heat(
...
+                    bus1=nodes,
                     carrier=f"{heat_system} {heat_source} heat pump",
-                    efficiency=cop_heat_pump,
-                    capital_cost=costs.at[costs_name_heat_pump, "efficiency"]
-                    * costs.at[costs_name_heat_pump, "capital_cost"]
+                    efficiency=(1 / cop_heat_pump.clip(lower=0.001)).replace(1000, 0),
+                    capital_cost=costs.at[costs_name_heat_pump, "capital_cost"]
                     * overdim_factor,
+                    p_min_pu=-cop_heat_pump / cop_heat_pump.clip(lower=0.001),
+                    p_max_pu=0,
```
cpschau: Same here

scripts/prepare_sector_network.py
```diff
@@ -3323,13 +3307,14 @@ def add_heat(
             n.add(
                 "Link",
                 nodes + f" {heat_system} resistive heater",
-                bus0=nodes,
-                bus1=nodes + f" {heat_system} heat",
+                bus0=nodes + f" {heat_system} heat",
+                bus1=nodes,
```
cpschau, line 3301: Is there any advantage of reordering the buses for resistive heaters as well? If yes, I would suggest to include it in #1748 instead of here.

scripts/solve_network.py
```diff
@@ -964,6 +966,140 @@ def add_TES_charger_ratio_constraints(n: pypsa.Network) -> None:
...
+    # Get model variable for Link dispatch
+    p = (n.model["Link-p"])
+    rhs = p.loc[:, ptes_discharger_ext] * (1 - ptes_direct_utilisation_profiles)
+    lhs = None
+
+    for tech in ptes_booster_technologies:
+        booster_technologies_links_ext = (
+            n.links[n.links.index.str.contains("urban central") &
+            n.links.index.str.contains(tech)]
+            .query("p_nom_extendable").index
```
cpschau: Why do the booster links have to be extendable? Couldn't brownfield assets also be used for boosting purposes?

scripts/solve_network.py
```diff
@@ -964,6 +966,140 @@ def add_TES_charger_ratio_constraints(n: pypsa.Network) -> None:
...
 
+def add_storage_temperature_boosting_constraints(
+    n: pypsa.Network,
+    ptes_temperature_boost_ratio_profile_file: str,
+    cop_profiles_file: str,
+    ptes_direct_utilisation_profiles_file: str,
+    ptes_booster_technologies: list[str],
+):
+    """
+    Add storage temperature boosting constraints.
```
cpschau: A more detailed description of the constraint within the docstring would be helpful.

scripts/solve_network.py
```diff
@@ -964,6 +966,140 @@ def add_TES_charger_ratio_constraints(n: pypsa.Network) -> None:
...
+):
+    """
+    Add forward temperature boosting constraints.
+    """
+    if not ptes_booster_technologies:
+        return
+
+    ptes_charger_ext = (
+        n.links[n.links.index.str.contains("urban central water pits charger")]
+        .query("p_nom_extendable").index
```
cpschau: Filter for extendables not necessary

scripts/solve_network.py
```diff
@@ -964,6 +966,140 @@ def add_TES_charger_ratio_constraints(n: pypsa.Network) -> None:
...
+):
+    """
+    Add storage temperature boosting constraints.
+    """
+    if not ptes_booster_technologies:
+        return
+
+    ptes_discharger_ext = (
+        n.links[n.links.index.str.contains("urban central water pits discharger")]
+        .query("p_nom_extendable").index
```
cpschau: Filter for extendables not necessary

scripts/solve_network.py
```diff
@@ -964,6 +966,140 @@ def add_TES_charger_ratio_constraints(n: pypsa.Network) -> None:
...
+
+
+def add_forward_temperature_boosting_constraints(
+    n: pypsa.Network,
+    ptes_forward_temperature_boost_ratio_profile_file: str,
+    ptes_booster_technologies: list[str],
+    ptes_direct_utilisation_profiles_file: str,
+):
+    """
+    Add forward temperature boosting constraints.
```
cpschau: Also here some further explanations would be great.

scripts/solve_network.py
```diff
@@ -964,6 +966,140 @@ def add_TES_charger_ratio_constraints(n: pypsa.Network) -> None:
...
+    # Get model variable for Link dispatch
+    p = (n.model["Link-p"])
+    rhs = p.loc[:, ptes_charger_ext] * ptes_direct_utilisation_profiles
+    lhs = None
+
+    for tech in ptes_booster_technologies:
+        booster_technologies_links_ext = (
+            n.links[n.links.index.str.contains("urban central") &
+            n.links.index.str.contains(tech)]
+            .query("p_nom_extendable").index
```
cpschau: Filter for extendables not necessary

scripts/prepare_sector_network.py
cpschau: This point was discussed in person. Reordering the buses for resistive heaters was found to be practical to simplify boosting constraint formulations in `solve_network`, as the power signs of boosting technologies are uniformly defined (negative `p_min_pu`/`p_max_pu` at the moment) .

scripts/prepare_sector_network.py
cpschau: This point was discussed in person. Reordering the buses for resistive heaters was found to be practical to simplify boosting constraint formulations in solve_network, as the power signs of boosting technologies are uniformly defined (negative p_min_pu/p_max_pu at the moment) .

scripts/solve_network.py
```diff
@@ -964,6 +966,140 @@ def add_TES_charger_ratio_constraints(n: pypsa.Network) -> None:
...
+
+    ptes_temperature_boost_ratio_dataaray = xr.open_dataarray(ptes_temperature_boost_ratio_profile_file)
+    ptes_direct_utilisation_profiles_dataary = xr.open_dataarray(ptes_direct_utilisation_profiles_file)
+    cop = xr.open_dataarray(cop_profiles_file)
+    tes_systems = {t.value for t in TesSystem}
+
+    ptes_direct_utilisation_profiles = (
+        ptes_direct_utilisation_profiles_dataary.to_pandas()
+        .reindex(index=n.snapshots)
+        .set_axis(ptes_discharger_ext, axis=1)
```
cpschau: This part needs to be adjusted for myopic test runs. The direct utilisation profiles have different column names than the discharger indices, as the latter ones also comprise a technology name and a build year. Also during test runs, I was encountering problems with regions that do not have any district heating. These won't have TES components available, but the utilisation profile is computed anyway leading to a length mismatch.

scripts/build_ptes_operations/ptes_temperature_approximator.py
TomKae00: The term ``forward_temperature_boosting`` might be misleading, I will change it to ``charger_temperature_boosting_required``. This means that if the forward temperature is below ``max_ptes_top_temperature`` and ``charger_temperature_boosting_required`` (preheating) is set to true, we will boost the top layer temperature to ``max_ptes_top_temperature``. This results in a higher ``max_capacity`` during summer and a higher COP for the PTES heat pump.

scripts/solve_network.py
TomKae00: Yes, you’re right, I’ll remove the ``p_nom_extendable`` part so that we catch all relevant links, including brownfield assets.

scripts/solve_network.py
cpschau: In general, I would be careful to use set_axis, as it overrides the indices by order without providing a clear mapping.

scripts/prepare_sector_network.py
```diff
@@ -3264,40 +3253,20 @@ def add_heat(
...
                     carrier=f"{heat_system} {heat_source} heat pump",
-                    efficiency=(-(cop_heat_pump - 1)).clip(upper=0),
-                    efficiency2=cop_heat_pump,
-                    capital_cost=costs.at[costs_name_heat_pump, "efficiency"]
-                    * costs.at[costs_name_heat_pump, "capital_cost"]
+                    efficiency=1,
+                    capital_cost=costs.at[costs_name_heat_pump, "capital_cost"]
                     * overdim_factor,
+                    p_max_pu=0,
+                    p_min_pu=-(1 / cop_heat_pump.clip(lower=0.001)).replace(1000, 0),
```
cpschau, line 3272: Why is the COP not written into the efficiency but as p_min_pu instead? Also why is the replace-call necessary?

TomKae00: You’re right, the replace call is unnecessary. 
We keep efficiency = 1 because the ptes heat pump link only carries the electric input P_el, the PTES discharger already provides Q_source, so putting COP into efficiency would double count. By setting p_min_pu = -(1 / COP), we ensure that the invest_cost reflects actual thermal capacity.

scripts/solve_network.py
```diff
@@ -972,6 +974,230 @@ def add_TES_charger_ratio_constraints(n: pypsa.Network) -> None:
...
+
+        ptes_boost_per_charge = (
+            ptes_boost_per_charge_dataaray
+            .to_pandas()
+            .loc[n.snapshots, booster_nodes]
+            .dropna(axis=1, how="all")
+            .rename(columns=booster_node_to_link)
+            .reindex(columns=booster_technologies_links)
+        )
+        gamma = (1 / ptes_boost_per_charge).where(ptes_boost_per_charge > 0, 0)
```
cpschau: Seems to be inconsistent with calculation of ration in `ptes_temperature_approximator`

config/config.default.yaml
```diff
@@ -514,12 +514,16 @@ sector:
...
-      dynamic_capacity: true
-      supplemental_heating:
-        enable: false
-        booster_heat_pump: false
+      enable: true
+      dynamic_capacity: false
+      discharge_boosting_required: false
+      charge_boosting_required: false
+      temperature_profile: 'constant'
+      booster_technologies: ["resistive heater", "air heat pump", "water pits heat pump"]
```
amos-schledorn:
```suggestion
      enable: true
      dynamic_capacity: false
      discharge_boosting_required: false
      charge_boosting_required: false
      temperature_profile: 'constant'
      booster_technologies: ['resistive heater', 'air heat pump', 'water pits heat pump']
```

doc/configtables/sector.csv
```diff
@@ -20,10 +20,18 @@ district_heating,--,,
...
--- -- -- enable,--,"{true, false}",Add option to enable supplemental heating of pit storage in district heating
--- -- -- booster_heat_pump: true,--,"{true, false}",Add option to enable a booster heat pump for supplemental heating of pit storage in district heating
+-- -- discharge_boosting_required,--,"{true, false}",Require temperature boosting on discharge when the storage’s top temperature falls below the district heating forward temperature
+-- -- charger_boosting_required,--,"{true, false}",Require temperature boosting on charge when the forward temperature is below the storage’s maximum top temperature (only if temperature_profile = 'constant')
+-- -- temperature_profile,--,"{'constant', 'dynamic'}",Select the storage top and bottom temperature profile:
+        'constant' - fixed at max_top_temperature and min_bottom_temperature
+        'dynamic'  - temperature is bounded to distirct heatings forward and return temperature
+-- -- booster_technologies,--,List of available boosting technologies (note: ‘water pits heat pump’ is only used for discharge boosting). Must be subset of ['resistive heater';
+       'air heat pump';
+       'water pits heat pump'],
```
fneum, line 32: I think they need to be single line.

config/config.default.yaml
```diff
@@ -545,8 +549,10 @@ sector:
         ignore_missing_regions: false
     direct_utilisation_heat_sources:
     - geothermal
-    temperature_limited_stores:
-    - ptes
+  tes: true
+  tes_tau:
+    decentral: 3
+    central: 180
```
fneum, line 555: Intentional? Undocumented?

doc/release_notes.rst
```diff
@@ -21,6 +19,14 @@ Upcoming Release
 
 * Allow expandable CCGTs by default
 
+* Updated standing losses for PTES, central TTES, and decentral TTES, previously calculated using the ``tes_tau`` parameter, to the latest DEA technology data, and updated costs version to v0.13.3.
```
fneum, line 22: but in `config.default.yaml` some `tes_tau` settings are added?

doc/release_notes.rst
```diff
@@ -21,6 +19,14 @@ Upcoming Release
...
+
+* Introduce temperature boosting for pit thermal energy storage (PTES) while discharging and charging, configurable via ``sector: ptes:`` in the configs. Available boosting technologies include resistive heaters, air heat pump, and water pit heat pump. This update supersedes and replaces PyPSA-Eur pull request #1692 (https://github.com/PyPSA/pypsa-eur/pull/1692).
```
fneum, line 24:
```suggestion
* Introduce temperature boosting for pit thermal energy storage (PTES) while discharging and charging, configurable via ``sector: ptes:`` in the configs. Available boosting technologies include resistive heaters, air heat pump, and water pit heat pump.
```

rules/solve_overnight.smk
```diff
@@ -3,6 +3,32 @@
...
+        )
+
+    if config_provider(
+        "sector", "district_heating", "ptes", "charge_boosting_required"
+    )(w):
+        profiles["boost_per_charge_profile"] = resources(
+            "boost_per_charge_profile_base_s_{clusters}_{planning_horizons}.nc"
+        )
+
+    return profiles
```
fneum, line 29: Can move to `common.smk` to prevent code duplication of this function in overnight and myopic scenarios.

scripts/prepare_sector_network.py
```diff
@@ -2944,83 +2942,91 @@ def add_heat(
         )
 
         if options["tes"]:
-            n.add("Carrier", f"{heat_system} water tanks")
+            if options["district_heating"]["ttes"]["enable"]:
```
fneum, line 2945: It would be nicer if indentation level could be reduced by one level for this section.

scripts/solve_network.py
```diff
@@ -972,6 +973,233 @@ def add_TES_charger_ratio_constraints(n: pypsa.Network) -> None:
...
+    ptes_boost_per_discharge_dataarray = xr.open_dataarray(
+        boost_per_discharge_profile_file
+    ).rename({"name": "node"})
+
+    # --- RHS: ptes discharger
+    ptes_discharger_links = n.links.index[
+        n.links.index.str.contains("urban central water pits discharger")
+    ]
+    ptes_discharger_node_da = xr.DataArray(
+        ptes_discharger_links.str.extract(r"^(\S+\s+\S+)")[0].to_numpy(),
```
fneum, line 1026: maybe comment on what regex matches here

fneum (commented):
Looks mostly good to me. Some smaller things. Happy to merge when @amos-schledorn or @cpschau double-checked and are fine with it.

It would be good if explanations in PR make it into documentation, not just docstrings. Otherwise it will be easily lost / overlooked.

fneum:
superseded by #1893

PR #1766, opened 2025-07-21
fix:add geothermal-sourced HP to HP cost names and turn on geothermal heating in test configs

### LINKED ISSUE #1763: No geothermal in default configuration, + a geothermal error in the new pypsa-eur v2025.07.0
With the older pypsa-eur (v2025.04.0) there is no geothermal energy used by default. This happens because geothermal direct utilisation is only activated when geothermal is also a heat_source in params.heat_pump_sources. However, by default we have 
```
heat_pump_sources:
    urban central:
    - air
    urban decentral:
    - air
    rural:
    - air
    - ground
```
I thought it would be a good idea to allow direct utilisation without also having to allow geothermal as a heat pump source. Or at least be clearer in the documentation about the fact that geothermal needs to be put under heat_pump_sources for it to be installed, and also if it can be urban central, urban decentral and/or rural, because that is unclear.

I wanted to check if this problem still occurred in the newest pypsa-eur version (v2025.07.0 (11th July 2025)) and it is actually worse. The build_geothermal_heat_potential.py script seems to work fine, but when running a model with 
```
scenario:
  clusters:
  - 16
  sector_opts:
  - 3H
  planning_horizons:
  - 2050
countries: ['NL', 'BE', 'DE', 'NO', 'DK']
sector:
  heat_pump_sources:
    urban central:
    - air
    - geothermal
```
I get this error from prepare_sector_network.py: 
.snakemake\scripts\tmp2f3ng4ho.prepare_sector_network.py", line 3240, in add_heat
    capital_cost=costs.at[costs_name_heat_pump, "efficiency"]
                 ~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
Resulting in KeyError: 'central geothermal-sourced heat pump'
This configuration worked in the previous pypsa-eur version, so something must have changed in the code, but I don't know what change would cause this error and if it was done in purpose. I wouldn't know how else to ensure geothermal is part of the energy system.

amos-schledorn:
Closes # (if applicable).
Partially closes #1763. 

@mareenjp raised that there is no heat-pump-cost name for geothermal heating. This is fixed by assuming the same costs as for other water-water heat pumps (DEA costs for excess-heat-sourced heat pumps).

## Checklist

### DIFF
--- a/config/test/config.myopic.yaml
+++ b/config/test/config.myopic.yaml
@@ -40,6 +40,15 @@ sector:
       supplemental_heating:
         enable: true
         booster_heat_pump: true
+    heat_pump_sources:
+        urban central:
+        - air
+        - geothermal
+        urban decentral:
+        - air
+        rural:
+        - air
+        - ground
   hydrogen_turbine: false
   regional_oil_demand: false
   regional_co2_sequestration_potential:
--- a/config/test/config.myopic.yaml
+++ b/config/test/config.myopic.yaml
@@ -41,14 +41,14 @@ sector:
         enable: true
         booster_heat_pump: true
     heat_pump_sources:
-        urban central:
-        - air
-        - geothermal
-        urban decentral:
-        - air
-        rural:
-        - air
-        - ground
+      urban central:
+      - air
+      - geothermal
+      urban decentral:
+      - air
+      rural:
+      - air
+      - ground
   hydrogen_turbine: false
   regional_oil_demand: false
   regional_co2_sequestration_potential:
--- a/config/test/config.overnight.yaml
+++ b/config/test/config.overnight.yaml
@@ -69,14 +69,14 @@ sector:
     ates:
       enable: true
     heat_pump_sources:
-        urban central:
-        - air
-        - geothermal
-        urban decentral:
-        - air
-        rural:
-        - air
-        - ground
+      urban central:
+      - air
+      - geothermal
+      urban decentral:
+      - air
+      rural:
+      - air
+      - ground
   hydrogen_turbine: false
   regional_oil_demand: false
   regional_co2_sequestration_potential:
--- a/config/test/config.overnight.yaml
+++ b/config/test/config.overnight.yaml
@@ -68,6 +68,15 @@ sector:
         booster_heat_pump: true
     ates:
       enable: true
+    heat_pump_sources:
+        urban central:
+        - air
+        - geothermal
+        urban decentral:
+        - air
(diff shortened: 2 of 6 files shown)


mareenjp:
@amos-schledorn Thank you for looking into the bug! I added if heat_source in ["ptes", "geothermal"]: to my code and indeed the error disappeared, however, the amount of geothermal heat installed is very low compared to what was installed with the same configuration in the previous pypsa-eur version. Perhaps this is a mistake somewhere on my side, so maybe you can check this as well. I looked for example at the nodal energy balance and in the old version I got: 
Generator,urban central geothermal heat,,urban central geothermal heat,160618879.01212
With this new version it is only:
Generator,urban central geothermal heat,,urban central geothermal heat,7.90357
This was for a 2030 model with NL, DE, BE, NO and DK. Also as you might see, something that I forgot to mention before, but geothermal generation and urban central geothermal heat direct utilisation are not assigned to specific nodes, so this is another bug in the code. I think the problem there is that usually when we add a bus, location=nodes is used, which is not the case here.

amos-schledorn:
> @amos-schledorn Thank you for looking into the bug! I added if heat_source in ["ptes", "geothermal"]: to my code and indeed the error disappeared, however, the amount of geothermal heat installed is very low compared to what was installed with the same configuration in the previous pypsa-eur version. Perhaps this is a mistake somewhere on my side, so maybe you can check this as well. I looked for example at the nodal energy balance and in the old version I got: Generator,urban central geothermal heat,,urban central geothermal heat,160618879.01212 With this new version it is only: Generator,urban central geothermal heat,,urban central geothermal heat,7.90357 This was for a 2030 model with NL, DE, BE, NO and DK. Also as you might see, something that I forgot to mention before, but geothermal generation and urban central geothermal heat direct utilisation are not assigned to specific nodes, so this is another bug in the code. I think the problem there is that usually when we add a bus, location=nodes is used, which is not the case here.

@mareenjp I've added the location attribute to the heat-source bus creation (this shouldn't change your results though).

The lower geothermal build-out might have to do with updated investment costs (see [technology-data](https://github.com/PyPSA/technology-data)). However, we're considering another update, which would switch to the newly introduced large-scale geothermal projects suggested in the DEA technology catalogue. 
If you're interested in geothermal heating specifically, I'd recommend looking into the DEA data and potentially scaling the investment costs to your liking.

cpschau (approved):
LGTM

PR #1787, opened 2025-07-31
define extra_funcitonality inside solve_network to capture snakemake object

lindnemi:
At the moment the solve_network function can not be imported from other scripts, because the snakemake object required by extra_functionality is a global variable.

This PR defines the extra_functionality within solve_network, and adds the snakemake object as an additional argument to solve_network, s.t. the snakemake object can be captured from the scope of solve_network instead of the global scope.
This allows imports of the `solve_network` function from other scripts.

For an application see: https://github.com/PyPSA/pypsa-de/pull/103

Before completing the PR i'd like to hear some opinions @fneum @lkstrp 

Closes # (if applicable).

## Checklist


FabianHofmann (commented):
since moving this will likely lead to a lot of merge conflicts in downstream forks, what you can do alternatively is to pass the snakemake object to `n._snakemake` and change
```python
        custom_extra_functionality(n, snapshots, snakemake)  # pylint: disable=E0601
```
to 
```python
        custom_extra_functionality(n, snapshots, n._snakemake)
```

coroa:
Or you make an explicit argument to `extra_functionality`:
```python
def extra_functionality(
    n: pypsa.Network, snapshots: pd.DatetimeIndex, planning_horizons: str | None = None, snakemake: Snakemake
) -> None:
```

and then pass it to `n.optimize` with:
```python
n.optimize(..., extra_functionality=partial(extra_functionality, snakemake=snakemake))
```

lindnemi:
I like the `partial` suggestion

While we are at it: If snakemake is passed to solve_network should we get rid of all the other keywords as well?

fneum (approved):
This looks good! Let's start with that.

fneum (approved):

PR #1840, opened 2025-09-23
refactor: remove hotfix and check that the directory exists

tgilon:
Closes # (if applicable).

This PR removes the hotfix in `progress_retrieve` and checks that the directory exists. I tested it with the latest env in the master for `run:disable_progressbar` with both `true` and `false`. I wasn't able to reproduce the bug introduced in https://github.com/PyPSA/pypsa-eur/commit/54e39dfbbe7eea26588825cf9334ea5dab85f6a7 in the first place.

## Checklist


scripts/_helpers.py
coroa: I agree that this `disable=True` should not be necessary. @fneum Do you remember the circumstances of the hotfix?

The `disable` code should also be changed to:
```python
response = requests.get(url, headers=headers, stream=True)
with open(file, "wb") as f:
    shutil.copyfileobj(response.raw, f)
```

Since currently `response.content` is happily loading everything into memory.

fneum: No, I don't remember. @bobbyxng changed it when introducing OSM.

tgilon: @coroa I introduced the streaming copy for `disable`.

PR #1842, opened 2025-09-25
implement national CO2 emissions reduction targets

ebbekyhl:

I have seen more and more PyPSA-Eur studies including national targets for CO2 emissions reductions. Some use approaches that might be unhandy to implement in the master branch. But with Linopy this can be done relatively easy, with a small addition to the config file and the solve_network script. Maybe something to consider?

I tested this for UK, and it does the job. I have only tested it for a myopic optimization - needs to be tested for other foresight modes.

## Checklist

### DIFF
--- a/config/config.default.yaml
+++ b/config/config.default.yaml
@@ -105,7 +105,7 @@ co2_budget:
   2045: 0.050
   2050: 0.000 # climate-neutral by 2050
 
-local_co2_budget: False
+local_co2_budget: false
 
 # docs in https://pypsa-eur.readthedocs.io/en/latest/configuration.html#electricity
 electricity:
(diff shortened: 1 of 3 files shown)


pworschischek-aggmag:
Hi @ebbekyhl, over at PyPSA-AT we also needed national CO2 budgets and implemented it recently in https://github.com/AGGM-AG/pypsa-at/pull/30. It's a simplified version of the PyPSA-DE constraint. I did not know about the `n.optimize.expressions` so some refactoring needs to be done. Happy to join forces and bring this to PyPSA-EUR

lindnemi:
It's a great idea to have national emissions budgets directly in PyPSA-Eur. Let me share a few points from our experience at PyPSA-DE.

One tricky part, which the constraint in PyPSA-DE tries to address, is the ability to account for e-fuel/carbon trade.

If a country imports a lot of e-/biofuels the emissions from burning them will show up in its emissions balance. If the national budget does not take into account, that will create a counter-incentive to importing e-/biofuels.

That is especially tricky if the gas pipeline network is activate, because in the gas links the fossil and the renewable gas gets mixed up, which means it's challenging to track its carbon content.

Maybe there could be several modes for the national budget, which differ in how they treat e-fuel imports.

PR #1846, opened 2025-10-03
Add user friendly messages to snakemake workflow

willu47:
Closes #1844 .

Blocked by PyPSA/snakemake-logger-plugin-pypsa#2

## Changes proposed in this Pull Request

- Adds new bioconda dependency PyPSA/snakemake-logger-plugin-pypsa which provides user-readable messages describing what each rule does in understandable language.
- Adds `message` annotation to each rule

Instead of:

```python
[Fri Oct  3 17:11:26 2025]
Job 0: plot_power_networks_clustered
Reason: Input files updated by another job: resources/test-elec-tyndp/maps/power-network-s-all.pdf
[Fri Oct  3 17:11:26 2025]
Finished jobid: 0 (Rule: plot_power_networks_clustered)
```
we see
```python
[Fri Oct  3 17:11:26 2025]
Job 0: Plotting clustered power network topology
Reason: Input files updated by another job: resources/test-elec-tyndp/maps/power-network-s-all.pdf
[Fri Oct  3 17:11:26 2025]
Finished jobid: 0 (Rule: plot_power_networks_clustered)
```

## Checklist

- [x] I tested my contribution locally and it works as intended.
- [x] Code and workflow changes are sufficiently documented.
- [x] Changed dependencies are added to `envs/environment.yaml`.
- [x] Changes in configuration options are added in `config/config.default.yaml`.
- [x] Changes in configuration options are documented in `doc/configtables/*.csv`.
- [x] Sources of newly added data are documented in `doc/data_sources.rst`.
- [x] A release note `doc/release_notes.rst` is added.

## TODO

- [ ] Integrate with [PyPSA-Eur documentation](https://pypsa-eur.readthedocs.io/en/latest/preparation.html)

PR #1855, opened 2025-10-10
feat: Add simple descriptive logger plugin and use by default

coroa:
Follow-up/variant to #1846 .

The logger plugin registers a derived class of snakemake's DefaultFormatter
which shows the bold message in front of the regular job info block.

### DIFF
--- a/Snakefile
+++ b/Snakefile
@@ -79,12 +79,6 @@ if config["foresight"] == "perfect":
     include: "rules/solve_perfect.smk"
 
 
-# Appends the name of the rule to each rule's message for easier identification in logs
-for name, r in rules._rules.items():
-    if f"{name}" not in r.rule.message:
-        r.rule.message += f" (Rule: {name})"
-
-
 rule all:
     input:
         expand(RESULTS + "graphs/costs.svg", run=config["run"]["name"]),


willu47:
Hi @coroa - I think this is a cleaner implementation than manually adding information to each message. However it adds an extra dependency to the repo, and a configuration step. Is there a way to automate/avoid the need to manually install the logger plug-in?

Also, can you change the target of your pull request to my branch?  I can then continue the work against the original issue?

coroa:
> Hi @coroa - I think this is a cleaner implementation than manually adding information to each message. However it adds an extra dependency to the repo, and a configuration step. Is there a way to automate/avoid the need to manually install the logger plug-in?

In a way it is already automated, since i added it to `environment.yaml`. so that when people update their environment the plugin is pulled in automatically as well, the pinned environments are usually updated based on environment.yaml by a CI process. What is the configuration step? If you refer to writing `--logger descriptive`. That is not necessary, because i added it to `profiles/default/config.yaml`.

> Also, can you change the target of your pull request to my branch? I can then continue the work against the original issue?

That was my plan, but i can't select your branch since it is in your fork, you will have to merge it into your branch manually.

coroa:
I am not fond of the name `descriptive`. It does show up as the `--logger descriptive` argument, and if we want to add new options to choose the amount of output `--logger-descriptive-verbose` could be used as a flag (ie. it always starts with `--logger-<packagename>-<optionname>`).

I was thinking maybe just naming it `--logger pypsa` would be more obvious.

coroa:
Pulled into #1846 manually.

PR #1371, opened 2024-10-18
initial dsm implementation for the heating sector

martacki:
Closes # (if applicable).

## Changes proposed in this Pull Request
adding demand side response for the heating sector

## Checklist

- [x] I tested my contribution locally and it works as intended.
- [ ] Code and workflow changes are sufficiently documented.
- [x] Changed dependencies are added to `envs/environment.yaml`.
- [x] Changes in configuration options are added in `config/config.default.yaml`.
- [ ] Changes in configuration options are documented in `doc/configtables/*.csv`.
- [x] Sources of newly added data are documented in `doc/data_sources.rst`.
- [ ] A release note `doc/release_notes.rst` is added.

PR #1857, opened 2025-10-14
Heat dsm 2025

lprieto1409:
Closes # (if applicable).

This PR introduces residential heat demand-side management (DSM) for electric heat pumps based on the smartEn / DNV methodology.

It enables flexible heat demand that shifts within configurable intraday periods, improving renewable integration and reducing system costs.

The new formulation models building thermal mass as an energy-storage resource, allowing load shifting within **user-defined windows** (default = 12 h) via the` restriction_time parameter,` while maintaining thermal comfort.

### Implementation

 - Adds `residential_heat` section in `config/config.default.yaml` with `dsm`, `restriction_value`, and `restriction_time` parameters (`restriction_value` reflects projected EC heat-pump rollout, ≈ 27% by 2030)
  - Applies DSM to rural, urban decentral, and urban central (district-heating) systems
  - Introduces `heat_dsm_profile()` in `scripts/build_hourly_heat_demand.py` and DSM stores in `scripts/prepare_sector_network.py`
  - Updates `rules/build_sector.smk`, documentation (`doc/configtables/sector.csv`, `doc/supply_demand.rst`),
   and release notes

## Validation & Impact Analysis
The following analysis demonstrates the flexibility feature is working correctly using a test scenario (EU27, TY: 2030, 38 clusters). Comparison: **Baseline (No DSM)** vs. **With DSM**.

 ### 1. DSM Dispatch & Renewable Response

The plot validates dynamic optimization: DSM shifts from wind-following in winter to solar-driven midday charging in summer, confirming responsiveness to renewable availability.

02_dsm_stores_dispatch

### 2. Heat Pump Efficiency Improvement

DSM improves heat pump efficiency by shifting operation to higher COP periods. The time series shows consistently higher COP for DSM, particularly in summer, with the distribution revealing more frequent operation at higher efficiency levels.

04_cop_improvement

### 3. Generation Mix Changes

06_generation_by_technology

The flexibility provided by DSM dynamically shifts the optimal generation mix, leading to the largest proportional increases in solar thermal and biogas dispatch:

### 4. Generation Dispatch Patterns 

07_generation_dispatch_patterns

The bottom panel shows DSM's net impact on generation. During summer, DSM reduces generation through better COP utilization and solar alignment. In winter, it increases generation due to thermal storage losses and pre-heating during high-wind periods.

### 5. Renewable Curtailment 

08_curtailment_by_technology

DSM reduces offshore wind curtailment by absorbing surplus generation, but increases onshore wind curtailment due to higher capacity deployment driven by storage losses.

### 6. System Cost Impact

10_system_cost_comparison

Heat DSM reduces total system cost by 1.3 billion EUR (0.22%). Generator capacity investment remains unchanged at 156 B€, while storage and link investments decrease slightly due to improved system flexibility.

**Definition of done**
- [x] Update the code to current `master` and resolve the merge conflict
- [x] Check the code is still working as intended
- [x] Check the feature implementation, if the implementation is reasonable and confirmed by literature/assumptions supported by the referenced document
- [x] Check if the code matches the style and requirements of the remaining PyPSA-Eur code
- [x] Add some documentation (PyPSA-Eur style, i.e. the configuration options, but also to the body of the PyPSA-Eur docs where the logic of implemented features is explained)
- [x] Update the assessment of the impact this PR has on the model figure/qualitatively

## Checklist

### DIFF
--- a/config/test/config.overnight.yaml
+++ b/config/test/config.overnight.yaml
@@ -62,7 +62,17 @@ sector:
   gas_network: true
   H2_retrofit: true
   residential_heat:
-    dsm: true
+    dsm:
+      enable: true
+      direction:
+      - overheat
+      restriction_value:
+        2020: 0.06
+        2025: 0.16
+        2030: 0.27
+      restriction_time:
+      - 10
+      - 21
   district_heating:
     ptes:
       supplemental_heating:


euronion:
Supersedes #1856 and #1371

lprieto1409:
**Check the code is still working as intended**
- [ ] I run `snakemake -call --configfile config/test/config.overnight.yaml` and the the code
  executes without errors
- [ ] DSM components are successfully created: 15 heating flexibility stores added to the network (11,732 MWh total capacity)
- [ ] However, there's a bug with temporal aggregation: The DSM profile is generated at hourly resolution but the test uses resolution_sector: 24h. When e_max_pu=heat_dsm_profile is set in scripts/prepare_sector_network.py:2997, timestamps don't match and PyPSA reindexes with zeros, preventing DSM stores from being utilized. The profile needs to be reindexed to match n.snapshots after temporal aggregation.

**Check the code is still working as intended**

config/test/config.overnight.yaml
euronion, line 1: why the changes to this file?

config/config.default.yaml
```diff
@@ -561,6 +561,9 @@ sector:
     rural:
     - air
     - ground
+  residential_heat_dsm: false
+  residential_heat_restriction_value: 0.27
+  residential_heat_restriction_time: [10, 22] # 9am and 9pm
```
euronion: These three options are related to each other, aren't they? They should then be grouped into one option

euronion: What's the timezone for this entry? The comment (btw. no need to have that comment in the config file) mentions 9am/9pm, but the numbers say 10 and 22.
Is this timezone adjusted / local time? Because if the model stretches over multiple time zones, buses will have different time zones. Will it always be local time, or maybe UTC, with some buses seeing flexibility between 7-19 local time while others see flex between 11-23?

config/plotting.default.yaml
euronion, line 1: What are the changes in these files for?

doc/release_notes.rst
```diff
@@ -53,6 +53,8 @@ Upcoming Release
 
 * Running perfect foresight is now marked as unstable and may not work as expected.
 
+* Add residential heat demand-side management (DSM) feature that allows electric heating systems to provide flexibility by shifting demand within daily time windows. The implementation models building thermal mass as energy stores with time-varying availability constraints based on SmartEn research. Configuration options include ``residential_heat_dsm`` (default: false), ``residential_heat_restriction_value`` (default: 0.27), and ``residential_heat_restriction_time`` (default: [10, 22] for 9am and 9pm checkpoint hours). Thermal losses in buildings are modeled using the same time constant as decentralized water tanks. (https://github.com/PyPSA/pypsa-eur/pull/1371)
```
euronion: Please add link to the SmartEn study here and in the docs.
Reduce verbosity in the release notes, no need for all implementation details.

doc/configtables/sector.csv
euronion, line 1: Your changes do not work in the docs, the table doesn't render (see here: https://pypsa-eur--1857.org.readthedocs.build/en/1857/configuration.html#sector ) due to the syntax beeing wrong.

doc/supply_demand.rst
```diff
@@ -169,6 +169,65 @@ Activated in Config from the `tes <https://github.com/PyPSA/pypsa-eur-sec/blob/3
...
 A thermal energy density of 46.8 kWh :math:`_{th}`/m3 is assumed, corresponding to a temperature difference of 40 K. The decay of thermal energy in the stores: 1- :math:`e^{-1/24τ}` is assumed to have a time constant  of  τ=180 days for central TES and  τ=3 days for individual TES, both modifiable through `tes_tau <https://github.com/PyPSA/pypsa-eur-sec/blob/3daff49c9999ba7ca7534df4e587e1d516044fc3/config.default.yaml#L229>`__ in config file. Charging and discharging efficiencies are 90% due to pipe losses.
 
+**Residential Heat Demand-Side Management (DSM)**
+
+Can be activated in the `config file <https://github.com/PyPSA/pypsa-eur/blob/master/config/config.default.yaml>`__ through the ``residential_heat_dsm`` option (default: false).
+
+Residential heat demand-side management allows electric heating systems to provide flexibility to the energy system by shifting heat demand within daily time windows while maintaining thermal comfort. This feature models the thermal mass of buildings as an energy storage capability, enabling heat pumps to operate more flexibly in response to electricity price signals and renewable generation variability.
```
euronion: only operation of heat pumps, or any heat source?

doc/supply_demand.rst
```diff
@@ -169,6 +169,65 @@ Activated in Config from the `tes <https://github.com/PyPSA/pypsa-eur-sec/blob/3
...
+
+*Implementation approach*
+
+The implementation is based on the smartEn/DNV methodology for quantifying demand-side flexibility benefits in the European energy system. Building thermal mass is represented as energy stores connected to residential heat buses, with time-varying availability constraints that enforce consumption requirements within defined periods.
```
euronion: Please link somewhere to the smartEn study and the precise location the assumptions are coming from, from within the study doc.

doc/supply_demand.rst
```diff
@@ -169,6 +169,65 @@ Activated in Config from the `tes <https://github.com/PyPSA/pypsa-eur-sec/blob/3
...
+
+*Time windows and constraints*
+
+Heat demand can be shifted within 12-hour periods to avoid buildings acting as long-term seasonal storage:
```
euronion: 12-hours is only default, right? Can be arbitrarily adjusted.

doc/supply_demand.rst
```diff
@@ -169,6 +169,65 @@ Activated in Config from the `tes <https://github.com/PyPSA/pypsa-eur-sec/blob/3
...
+
+- **Day period**: 9am to 9pm (21:00)
+- **Night period**: 9pm (21:00) to 9am
+
+At the boundaries between these periods (configured via `residential_heat_restriction_time <https://github.com/PyPSA/pypsa-eur/blob/master/config/config.default.yaml>`__, default: [10, 22] corresponding to 9am and 9pm), the thermal storage state of charge must return to its baseline level. This ensures that heat consumption requirements are met within each 12-hour window while allowing temporal load shifting for demand response.
+
+*Storage capacity and flexibility magnitude*
+
+The flexibility storage capacity is sized based on the maximum residential space heating demand at each node. The actual available flexibility is constrained by the `residential_heat_restriction_value <https://github.com/PyPSA/pypsa-eur/blob/master/config/config.default.yaml>`__ parameter (default: 0.27), which sets the maximum state of charge as a fraction of the storage capacity. A value of 0.27 represents a conservative estimate of the thermal buffer capacity available in residential buildings.
```
euronion: Where is the default 0.27 coming from?

doc/supply_demand.rst
```diff
@@ -169,6 +169,65 @@ Activated in Config from the `tes <https://github.com/PyPSA/pypsa-eur-sec/blob/3
...
+
+*Thermal characteristics*
+
+Thermal losses in the building thermal mass are modeled using the same standing loss rates as decentralized water tank storage, which provides a reasonable proxy for the thermal inertia of building envelopes. The storage is cyclic, meaning the state of charge at the end of the optimization period must equal that at the beginning.
```
euronion: Do we have anything in the SmartEn study that underlines this assumption? I'd say a hot water storage tank is likely better insulated that most buildings.

scripts/build_cop_profiles/BaseCopApproximator.py
euronion, line 1: what's the reason for the modifications here? I'm fine with keeping them, but why make them in this PR? Was there something throwing an error?

scripts/build_hourly_heat_demand.py
```diff
@@ -24,6 +25,57 @@
...
+
+    Parameters
+    ----------
+    nodes : pd.Index or array-like
+        Node identifiers for which to generate profiles.
+    options : dict
+        Configuration dictionary containing:
+        - 'residential_heat_restriction_time': list of int
+            Hours at which storage must be empty (checkpoint hours).
+            Default: [10, 22] corresponding to 9am and 9pm.
```
euronion: There is no default for the method... The default is always from the config and shouldn't be mentioned here (increases maintenance efforts)

scripts/build_hourly_heat_demand.py
```diff
@@ -40,6 +92,8 @@
         snakemake.params.snapshots, snakemake.params.drop_leap_day
     )
 
+    options = snakemake.params.sector
```
euronion: Please name the var differently or don't use this intermediate var at all.
`options` is too generic in the `main` block of this script - it could be any option for anything here - no linkage to the `sector` or `dsm` in the name of the variable.

scripts/plot_balance_map.py
euronion, line 1: Why all these unrelated changes? Were they necessary?

scripts/make_summary.py
euronion, line 1: Why all these unrelated changes? Were they necessary?

scripts/plot_gas_network.py
euronion, line 1: Why all these unrelated changes? Were they necessary?

IMPACT_ASSESSMENT_HEAT_DSM.md
euronion, line 1: We always report the changes in the PRs, not as files in the repo. Do you also have a figure that shows how it behaves?

config/config.default.yaml
lprieto1409: done as suggested

config/test/config.overnight.yaml
lprieto1409: I made some changes as some values are already on default, residential_heat: dsm: true - is the only change needed to activate the residential heat DSM feature in the test configuration

config/config.default.yaml
lprieto1409: The values [10, 22] are UTC hours. This means different local times across Europe (10am/10pm in Portugal, 11am/11pm in Germany, 12pm/midnight in Greece). This is a simplification—the SmartEn study (Appendix A, section 1.1.2.5) specifies 9am-9pm but doesn't indicate timezone. Implementing per-bus local time would require coordinate-based timezone lookup and complicate the model. I've removed the misleading comment from the config file and clarified the documentation.

config/plotting.default.yaml
lprieto1409: I reverted to match master

doc/configtables/sector.csv
lprieto1409: Thanks for the comment, this is fixed now!

doc/release_notes.rst
lprieto1409: done as suggested

doc/supply_demand.rst
lprieto1409: SmartEn study was heat pumps only, but the implementation covers all residential electric heating. Clarified in the docs.

doc/supply_demand.rst
lprieto1409: Added precise references: Appendix A, Section 1.1.2.5 and page 17 Section 3.6 of the smartEn study.

doc/supply_demand.rst
lprieto1409: Correct, 12-hour is the default. Updated docs to clarify that checkpoint hours are configurable

doc/supply_demand.rst
lprieto1409: It represents max thermal storage capacity (fraction of heating demand). Likely a conservative estimate for comfort limits. This value was in the original branch with no related documentation.

doc/supply_demand.rst
lprieto1409: No, the smartEn study doesn't support this assumption. Agreed that water tanks are better insulated than buildings. Updated docs to clarify this is a simplifying assumption.

scripts/build_cop_profiles/BaseCopApproximator.py
lprieto1409: You're right - nothing was failing. The changes were purely aesthetic (Union → |syntax for Python 3.10+, np.array → np.ndarray for correctness). Since they're not essential to the feature, I've reverted them to keep the PR focused.

scripts/build_hourly_heat_demand.py
lprieto1409: done as suggested

scripts/build_hourly_heat_demand.py
lprieto1409: done as suggested

scripts/make_summary.py
lprieto1409: Yes, they were necessary because pypsa.options.set_option() caused AttributeError: Invalid category 'set_option'

scripts/plot_balance_map.py
lprieto1409: Yes, they were necessary because PyPSA 0.33.2 introduced breaking API changes: (1) pypsa.options API was removed causing AttributeError, (2) geomap_colors was renamed to color_geomap, (3) separate flow parameters were consolidated into single flow parameter, (4) .fillna("lightgrey") prevents ValueError: Invalid RGBA argument: nan when matplotlib receives NaN colors from missing carrier definitions.

scripts/plot_gas_network.py
lprieto1409: Removed invalid geomap=False parameter that caused TypeError:  plot() got an unexpected keyword argument 'geomap_colors'. This parameter was removed in PyPSA 0.33.2 API - omitting it allows proper overlay rendering, consistent with how plot_hydrogen_network.py works.

euronion:
@lprieto1409 The `oet/master` got merged into this branch instead of the `pypsa-eur/master`.
Can you please fix it? It makes it difficult to review and we don't want the OET changes in here.

fneum:
Thanks for the PR. For getting it reviewed here, could you prepare:

1. An overview of what new functionality was added and how it can be used (configuration settings, new data dependencies, etc.). Please, not just an auto-generated AI summary.

2. Could you add a before (`master`) and after (this PR) comparison of system costs and energy balances, as well as any other aspects relevant to the feature? This allows us to assess whether the proposed changes are plausible quickly.

Thank you!

lprieto1409:
> @lprieto1409 The `oet/master` got merged into this branch instead of the `pypsa-eur/master`. Can you please fix it? It makes it difficult to review and we don't want the OET changes in here.

Thanks @euronion, I fixed it and also address the conflicts, let me know if you have any additional comments

doc/supply_demand.rst
```diff
@@ -169,6 +169,67 @@ Activated in Config from the `tes <https://github.com/PyPSA/pypsa-eur-sec/blob/3
...
+
+*Implementation approach*
+
+The implementation is based on the `smartEn/DNV methodology <https://smarten.eu/wp-content/uploads/2022/09/SmartEn-DSF-benefits-2030-Report_DIGITAL.pdf>`__ (see Appendix A, Section 1.1.2.5 "Residential electric heating" and page 17 Section 3.6) for quantifying demand-side flexibility benefits from residential heat pumps in the European energy system. The study estimates 195.5 TWh of annual flexibility potential for EU27 heat pumps using 12-hour load shifting constraints. Building thermal mass is represented as energy stores connected to residential heat buses, with time-varying availability constraints that enforce consumption requirements within defined periods. In the PyPSA-Eur implementation, this methodology is applied to all residential electric heating sources.
```
martacki: for which year do they estimate this amount? and does our initial implementation cover for this? I think it's hard to say, as we assume 27% of total demand that can be "stored" and pre-heated, plus the value would differ by the assumed `energy_totals_year`

doc/supply_demand.rst
```diff
@@ -169,6 +169,67 @@ Activated in Config from the `tes <https://github.com/PyPSA/pypsa-eur-sec/blob/3
...
+
+*Configuration example*
+
+.. literalinclude:: ../config/config.default.yaml
+   :language: yaml
+   :lines: 564-566
+
+*References*
+
+The methodology is based on: smartEn and DNV (2022). "Demand-side flexibility in the EU: Quantification of benefits in 2030." The study quantifies how residential heat pumps can provide 195.5 TWh of activated flexibility (both upward and downward) across the EU27 through load shifting within daily time constraints.
```
martacki: you could add that this was applied/developed first in https://www.sciencedirect.com/science/article/pii/S0306261925001515

doc/configtables/sector.csv
```diff
@@ -62,12 +62,15 @@ district_heating,--,,
 -- -- urban central,--,"List of heat sources for heat pumps in urban central heating. Must be one of [air, river_water, sea_water, geothermal]",
 -- -- urban decentral,--,"List of heat sources for heat pumps in urban decentral heating. Must be one of [air]",
 -- -- rural,--,"List of heat sources for heat pumps in rural heating. Must be one of [air, ground]",
-
+residential_heat,--,,Configuration options for residential heat demand-side management (DSM). See `smartEn DSM study <https://smarten.eu/wp-content/uploads/2022/09/SmartEn-DSF-benefits-2030-Report_DIGITAL.pdf>`_ (Appendix A) for methodology.
+-- dsm,--,"{true, false}",Enable residential heat demand-side management that allows electric heating systems to provide flexibility by shifting demand within configurable time periods. Models building thermal mass as energy storage.
+-- restriction_value,--,float,"Maximum state of charge (as fraction) for heat flexibility storage representing available thermal buffer capacity in buildings. Set to 0 for no flexibility; 1.0 for maximum flexibility. Default: 0.27 (conservative assumption for building thermal comfort constraints)."
```
martacki: might be good to motivate the 0.27 choice - I don't 100% recall our initial choice, it was a bit related to the 190 something TWh SmartEn estimate for 20X? we certainly assumed bigger values for 2040/50

doc/supply_demand.rst
martacki: year assumed is 2030 - - the potential they see would essentially be model result in our setting, not an input

doc/configtables/sector.csv
martacki: Ok I did some digging why we assume 27%:

Figure 44 (page 55/56) in Part 3 of the documents presented here: https://eur-lex.europa.eu/resource.html?uri=cellar:6c154426-c5a6-11ee-95d9-01aa75ed71a1.0001.02/DOC_3&format=PDF shows approx 60 mio HPs by 2030 (meaning 60 mio households electrify their heating demand), which is approx. 27% of households, thus 27% of demand.

thus, we might adapt the numbers for the other planning horizons as follows:
2020: 0.06
2025: 0.16
2030: 0.27
2035: 0.36
2040: 0.38
2045: 0.39
2050: 0.4

lprieto1409: Thanks @martacki, I'm working on a small test, I'll add these changes :)

IMPACT_ASSESSMENT_HEAT_DSM.md
lprieto1409: I created a new config file to test the approach (covers EU27), you can run the last config file (`config.heat_dsm_test.yaml`), some results are:
heat_dsm_utilization

euronion: From the figure it looks like the implementation is working but not used everywhere (good sign). The one on the right seems to be flipped in order (Bottom 15, not Top 15?).

For a comparison, we need a `master` compared with this PR though.

config/config.default.yaml
euronion: Can you check if this is indeed the case? I in the past `generate_periodic_profiles(...)` accounted for timezone, but  maybe it has changed since then. Can you compare the profiles for buses in different timezones, e.g. IE and GR?

I think we should properly model this. You can also get timezone information from the buses locations, e.g. 
```python
import tzfpy
n.buses.apply(lambda row: tzfpy.get_tz(row["x"], row["y"]), axis=1)
```

IMPACT_ASSESSMENT_HEAT_DSM.md
lprieto1409: Yes, you are right about the plot
image

lprieto1409: ## Comparison Baseline (master) vs Heat DSM branch

### System Costs
- Heat DSM reduces total system costs slightly:
  - Baseline: 32.266 B€
  - Heat DSM: 32.245 B€
  - Savings: 20.14 M€ (≈0.06%)
- Energy Balance:

| Component  | Diff (DSM – Master) | Interpretation |
|-----------|----------------------------|----------------|
| Generator | –4,497 MWh | Slight reduction in thermal/RES generation due to temporal shifting. |
| Store| +4,192 MWh | Increased charge/discharge activity driven by heating flexibility. |
| Link| +305 MWh | Minor shifts in heat/electricity conversion flows. |
| Line| +0.44 MWh| Negligible change in electricity grid flows. |
| Load| 0| Heat demand unchanged (as expected). |

To test this, the following command was used: `snakemake --configfile config/test/config.heat_dsm_test.yaml -call all --cores 4`

euronion: Which config for master are you comparing to?

Since we are introducing flexibility to the system, energy generation should not change significantly.
* I would expect it to increase due to standing losses for the flexibility stores.
* I would expect it to decrease due to shifting HP usage to times of better COP (higher outdoor temperature)
* A shift in generation technologies could also be happening. If this is happening cannot be established from your results.
* What would also be relevant is the dispatch pattern for generation assets and the charging/discharging of the flexibility stores, whether those patterns show the effects of the heat dsm

With your config, you are only evaluating 7 days in central European winter, right?

config/config.default.yaml
euronion: Looking at the `dsm_profile` generated by `build_hourly_heat_demand.py`, it looks indeed like the local timezones are correctly adjusted for:
```python
dsm_profile["residential space"].iloc[:24][["SK0 0", "RO0 0", "PT0 0"]].hvplot()
```
image

Can you confirm if my understanding is correct and then adjust the docs please?

lprieto1409: Yes, I think that makes sense, I'll adjust the docs

IMPACT_ASSESSMENT_HEAT_DSM.md
lprieto1409: I added the new tests and added plots in the PR definition, let me know if you have any comments

config/config.default.yaml
lprieto1409: I just adjust the docs

euronion:
Thanks a lot for the updated figures @lprieto1409 , this helps discussion the PR impacts!

I'm not fully happy with the PR results.
1. Looking at the figures, it seems that most of the DSF serves the integration of `solar thermal` outside of winter/the main heating season
2. The impact of DSF during winter/heating season on the electricity generation profile seems small.
3. The DSF only works in pulling demand forward in time, not by delaying demand, i.e. it allows to heat earlier, but not to defer the heating decision. This is a modelling decision, by assuming `e_min_pu = 0` for the DSM store.

Related to 1. and 2.:
This could either be an artifact of the model results you show only being for `6h` resolution, or they could be and unexpected interaction with other aspects. If I look at the earlier results from @martacki 's implementation at https://github.com/PyPSA/pypsa-eur/pull/1371 , I would have exepcted a more pronounced impact during winter/heating season.
Can you please check with `1h` results again. If then it doesn't change have a look and see if there is indeed not the behaviour that @martacki saw and what could be causing it. If it does change with different time resolution, then best to add a note to the documentation and hint at this PR with different results for different time resolutions.

Related to 3. I would argue that we should have `e_min_pu = (-1) * e_max_pu`, representing the thermal inertia of buildings (`e_max_pu > 0` -> building get's heated above regular temperature and cools down slowly, `e_min_pu < 0` -> building is allowed to cool down below regular temperature and then is heated up again). I suggest to make this configurable with a a dedicated config key, e.g. `shift: {back-and-forth,back,forth}`.

lprieto1409:
> Thanks a lot for the updated figures @lprieto1409 , this helps discussion the PR impacts!
> 
> I'm not fully happy with the PR results.
> 
>     1. Looking at the figures, it seems that most of the DSF serves the integration of `solar thermal` outside of winter/the main heating season
> 
>     2. The impact of DSF during winter/heating season on the electricity generation profile seems small.
> 
>     3. The DSF only works in pulling demand forward in time, not by delaying demand, i.e. it allows to heat earlier, but not to defer the heating decision. This is a modelling decision, by assuming `e_min_pu = 0` for the DSM store.
> 
> 
> Related to 1. and 2.: This could either be an artifact of the model results you show only being for `6h` resolution, or they could be and unexpected interaction with other aspects. If I look at the earlier results from @martacki 's implementation at #1371 , I would have exepcted a more pronounced impact during winter/heating season. Can you please check with `1h` results again. If then it doesn't change have a look and see if there is indeed not the behaviour that @martacki saw and what could be causing it. If it does change with different time resolution, then best to add a note to the documentation and hint at this PR with different results for different time resolutions.
> 
> Related to 3. I would argue that we should have `e_min_pu = (-1) * e_max_pu`, representing the thermal inertia of buildings (`e_max_pu > 0` -> building get's heated above regular temperature and cools down slowly, `e_min_pu < 0` -> building is allowed to cool down below regular temperature and then is heated up again). I suggest to make this configurable with a a dedicated config key, e.g. `shift: {back-and-forth,back,forth}`.

Thanks for your comments @euronion, to look even further at points 1 and 2, I run the same test on hourly resolution from January to March (3 months) here are some of the results comparing both:

_1 hour resolution_
04_cop_improvement

_6 hour resolution_
04_cop_improvement

_1 hour resolution_
06_generation_by_technology

_6 hour resolution_
06_generation_by_technology

_1 hour resolution_
07_generation_dispatch_patterns

_6 hour resolution_
07_generation_dispatch_patterns

_1 hour resolution_
08_curtailment_by_technology

_6 hour resolution_
08_curtailment_by_technology

_1 hour resolution_
11_dsm_soc_selected_stores

_6 hour resolution_
11_dsm_soc_selected_stores

euronion:
Thanks for the updated plots.
I was still puzzled by the little effects, but the 6h does smooth over some of the benefits of the 1h.
I had a look myself and made a few adjustments:

* Implemented directionality: Structures can be cooled or heated below/above requirements (`overheat` and `undercool` or `overheat-undercool`). The previous implementation was only `overheat`.
* Reduced the test scope to 5 countries, 2 months, hourly resolution for faster solving

Here are the results I'm getting:

The effect on the heat pump dispatch pattern is very noticeable:
image

Although the annual COP is not much affected:
image

The electricity generation / dispatch pattern is also affected, in peak by up to 27%:
image

Interestingly enough, the `undercool`-ing seems to be more prominent than the overheating. I think that change makes most of the difference. You can see this in the SoC of the heat dsm  / virtual stores, as they are more often negative than positive. This corresponds to letting the housing structure cool below target temperature during the morning/midday and to re-heat it to target temperature in the early afternoon/evening.
image

euronion:
I'm happy with the results now.

@martacki and @lprieto1409 feel free to have another look. I'd consider it RTM then

lprieto1409:
> I'm happy with the results now.
> 
> @martacki and @lprieto1409 feel free to have another look. I'd consider it RTM then

Thanks @euronion it does look good to me, I only saw a minor typo in sector.csv but other than that looks great :)

doc/configtables/sector.csv
```diff
@@ -62,12 +62,17 @@ district_heating,--,,
...
+-- -- direction,--,"{overheat-undercool, overheat, undercool}",'overheat-undercool' means both pre-heating and delayed heating are allowed. 'overheat' allows only pre-heating where buildings are heated up above target temperature and then allowed to cool down, while 'undercool' allows only delayed heating where buildings can cool below target temperature and then be heated up again.
+-- -- restriction_value,--,Dictionary with planning horizons as keys.,Maximum state of charge (as fraction) for heat flexibility storage representing available thermal buffer capacity in buildings. Set to 0 for no flexibility or to 1.0 to assume that the entire heating demand can contribute to flexibility.
+-- -- restriction_time,--,list of int,"Checkpoint hours (0-23) at which heat flexibility storage must return to baseline state of charge, automatically adjusted to each country's local timezone. Default: [10 22] creates 12-hour periods with checkpoints at 9am and 9pm local time in each region."
```
fneum:
```suggestion
-- -- restriction_time,--,list of int,"Checkpoint hours (0-23) at which heat flexibility storage must return to baseline state of charge, automatically adjusted to each country's local timezone. Default: [10 22] creates 12-hour periods with checkpoints at 10am and 22pm UTC time in each region."

```

config/test/config.heat_dsm_test.yaml
fneum, line 1: Please remove. We are at capacity regarding tests and the `test` folder should only include scenarios actually tested in CI.

config/test/config.no_heat_dsm_test.yaml
fneum, line 1: Please remove. We are at capacity regarding tests and the `test` folder should only include scenarios actually tested in CI.

config/config.default.yaml
```diff
@@ -565,6 +565,19 @@ sector:
...
+      direction: "overheat-undercool"
+      restriction_value:
+        2020: 0.06
+        2025: 0.16
+        2030: 0.27
+        2035: 0.36
+        2040: 0.38
+        2045: 0.39
+        2050: 0.4
+      restriction_time: [10, 22]
```
fneum:
```suggestion
      restriction_time:
      - 10
      - 22
```

doc/release_notes.rst
```diff
@@ -90,6 +90,8 @@ Upcoming Release
 
 * Running perfect foresight is now marked as unstable and may not work as expected.
 
+* Add residential heat demand-side management (DSM) based on `smartEn study <https://smarten.eu/wp-content/uploads/2022/09/SmartEn-DSF-benefits-2030-Report_DIGITAL.pdf>`_ methodology.
```
fneum:
```suggestion
* Add residential heat demand-side management (DSM) based on `smartEn study <https://smarten.eu/wp-content/uploads/2022/09/SmartEn-DSF-benefits-2030-Report_DIGITAL.pdf>`_ methodology. See new settings under `sector: residential_heat`.
```

scripts/build_hourly_heat_demand.py
```diff
@@ -24,6 +25,56 @@
...
+    Notes
+    -----
+    The checkpoint approach operationally enforces the constraint that heat
+    consumption requirements must be met within each 12-hour period, preventing
+    the building thermal mass from acting as long-term seasonal storage while
+    allowing short-term load shifting for demand-side flexibility.
+
+    See Also
+    --------
+    prepare_sector_network.add_heat : Uses this profile to constrain heat flexibility stores
```
fneum:
```suggestion
```
bit of trimming

scripts/build_hourly_heat_demand.py
```diff
@@ -24,6 +25,56 @@
...
+        DSM availability profile indexed by timestamp with columns for each node.
+        Values are 1.0 (storage available) for most hours and 0.0 at checkpoint
+        hours to force storage depletion and enforce consumption periods.
+
+    Notes
+    -----
+    The checkpoint approach operationally enforces the constraint that heat
+    consumption requirements must be met within each 12-hour period, preventing
+    the building thermal mass from acting as long-term seasonal storage while
+    allowing short-term load shifting for demand-side flexibility.
```
fneum: Could be integrated in description above.

scripts/make_summary.py
```diff
@@ -315,8 +315,8 @@ def calculate_market_values(n: pypsa.Network) -> pd.Series:
     assign_carriers(n)
     assign_locations(n)
 
-    pypsa.options.params.statistics.nice_names = False
-    pypsa.options.params.statistics.drop_zero = False
+    pypsa.set_option("params.statistics.nice_names", False)
+    pypsa.set_option("params.statistics.drop_zero", False)
```
fneum, line 319: The original solution should also be good, but requires PyPSA v0.35. Was there another reason to change?

doc/supply_demand.rst
```diff
@@ -169,6 +169,77 @@ Activated in Config from the `tes <https://github.com/PyPSA/pypsa-eur-sec/blob/3
...
+
+*Implementation approach*
+
+The implementation is based on the `smartEn/DNV methodology <https://smarten.eu/wp-content/uploads/2022/09/SmartEn-DSF-benefits-2030-Report_DIGITAL.pdf>`__ (see Appendix A, Section 1.1.2.5 "Residential electric heating" and page 17 Section 3.6) for quantifying demand-side flexibility benefits from residential heat pumps in the European energy system. The study estimates 195.5 TWh of annual flexibility potential for EU27 heat pumps using 12-hour load shifting constraints. Building thermal mass is represented as energy stores connected to residential heat buses, with time-varying availability constraints that enforce consumption requirements within defined periods. In the PyPSA-Eur implementation, this methodology is applied to all residential electric heating sources.
```
fneum: The link did not work for me. Needed to be:

https://smarten.eu/wp-content/uploads/2022/10/SmartEN-DSF-benefits-2030-Report_DIGITAL-1.pdf

fneum: > In the PyPSA-Eur implementation, this methodology is applied to all residential electric heating sources.

I don't think it's clear what is meant here, since the DSM option is on the demand side, not the supply side.

doc/supply_demand.rst
```diff
@@ -169,6 +169,77 @@ Activated in Config from the `tes <https://github.com/PyPSA/pypsa-eur-sec/blob/3
...
+
+*Time windows and constraints*
+
+Heat demand can be shifted within configurable time periods to avoid buildings acting as long-term seasonal storage. By default, 12-hour periods are used:
+
+- **Day period**: 9am to 9pm (21:00)
+- **Night period**: 9pm (21:00) to 9am
+
+At the boundaries between these periods (configured via `residential_heat_restriction_time <https://github.com/PyPSA/pypsa-eur/blob/master/config/config.default.yaml>`__, default: [10, 22] corresponding to 9am and 9pm local time), the thermal storage state of charge must return to its baseline level. The implementation automatically adjusts these checkpoint hours to each country's local timezone, ensuring consistent behavioral patterns across different regions. This ensures that heat consumption requirements are met within each time window while allowing temporal load shifting for demand response. Users can adjust the checkpoint hours to create different period lengths as needed.
```
fneum:
```suggestion
At the boundaries between these periods (configured via `residential_heat_restriction_time <https://github.com/PyPSA/pypsa-eur/blob/master/config/config.default.yaml>`__, default: [10, 22] corresponding to 10am and 10pm UTC time), the thermal storage state of charge must return to its baseline level. The implementation automatically adjusts these checkpoint hours to each country's local timezone, ensuring consistent behavioral patterns across different regions. This ensures that heat consumption requirements are met within each time window while allowing temporal load shifting for demand response. Users can adjust the checkpoint hours to create different period lengths as needed.
```
Which local time? I think it's clearer to go with UTC. Even CET/CEST has some unclarity given daylight saving time.

doc/supply_demand.rst
```diff
@@ -169,6 +169,77 @@ Activated in Config from the `tes <https://github.com/PyPSA/pypsa-eur-sec/blob/3
...
+
+*Storage capacity and flexibility magnitude*
+
+The flexibility storage capacity is sized based on the maximum residential space heating demand at each node. The actual available flexibility is constrained by the `residential_heat_restriction_value <https://github.com/PyPSA/pypsa-eur/blob/master/config/config.default.yaml>`__ parameter (default: 0.27), which sets the maximum state of charge as a fraction of the storage capacity.
```
fneum:
```suggestion
The flexibility storage capacity is sized based on the maximum residential space heating demand at each node. The actual available flexibility is constrained by the `residential_heat_restriction_value <https://github.com/PyPSA/pypsa-eur/blob/master/config/config.default.yaml>`__ parameter, which sets the maximum state of charge as a fraction of the storage capacity.
```
The default value is time-varying and only 0.27 for 2030.

doc/supply_demand.rst
```diff
@@ -169,6 +169,77 @@ Activated in Config from the `tes <https://github.com/PyPSA/pypsa-eur-sec/blob/3
...
+
+The default value of 0.27 represents a conservative assumption for the thermal buffer capacity available in residential buildings, balancing flexibility potential against thermal comfort constraints. This parameter is user-configurable and can be adjusted based on specific assumptions about building thermal mass, insulation quality, and acceptable temperature variations. Setting it to 0 disables heat flexibility, while values closer to 1.0 allow greater load shifting (though this may require stronger assumptions about building characteristics and occupant comfort tolerance).
```
fneum:
```suggestion
For instance, a value of 0.27 would represent a conservative assumption for the thermal buffer capacity available in residential buildings, balancing flexibility potential against thermal comfort constraints. This parameter is user-configurable and can be adjusted based on specific assumptions about building thermal mass, insulation quality, and acceptable temperature variations. Setting it to 0 disables heat flexibility, while values closer to 1.0 allow greater load shifting (though this may require stronger assumptions about building characteristics and occupant comfort tolerance).
```

doc/supply_demand.rst
```diff
@@ -169,6 +169,77 @@ Activated in Config from the `tes <https://github.com/PyPSA/pypsa-eur-sec/blob/3
...
+
+*Thermal characteristics*
+
+Thermal losses are modeled using the same standing loss rates as decentralized water tank storage (a simplifying assumption that likely underestimates actual building losses). The storage is cyclic, meaning the state of charge at the end of the optimization period must equal that at the beginning.
```
fneum, line 199: I would expect standing losses of buildings to be lower than that of a water tank. This is perhaps something to get a feeling for, but not a showstopper for merging.

doc/supply_demand.rst
```diff
@@ -169,6 +169,77 @@ Activated in Config from the `tes <https://github.com/PyPSA/pypsa-eur-sec/blob/3
...
+
+In this framework, heat pumps are assumed to provide flexibility to the energy system within daily operational constraints, represented by two time windows: from 9 am to 9 pm, and 9 pm to 9 am. These time windows are parametrizable, allowing sensitivity analyses of different flexibility periods.
+
+The potential for heat demand shifting is derived from the projected roll-out of heat pumps, as outlined in the European Commissions *Impact Assessment Report on the Heat Pump Action Plan* (`see page 55, Fig. 44 <https://eur-lex.europa.eu/resource.html?uri=cellar:6c154426-c5a6-11ee-95d9-01aa75ed71a1.0001.02/DOC_3&format=PDF>`__). The projected number of installed heat pumps is translated into the share of electrified households capable of participating in demand-side management (DSM).
+
+For example, a projected **58 million heat pumps** in 2030 corresponds to:
+
+``58 million / 202 million households in the EU ≈ 28 %``
+
+Thus, approximately 28 % of EU households are assumed to contribute to the residential DSM potential by 2030.
```
fneum, line 241: This section does not add much (if done equally detailed for other settings, the documentation would get very long). I would suggest: Keep it short and remove this part, maybe shift the reference up.

scripts/prepare_sector_network.py
```diff
@@ -2964,6 +2964,64 @@ def add_heat(
...
+            HeatSystem.URBAN_CENTRAL,
+        ]:
+            factor = heat_system.heat_demand_weighting(
+                urban_fraction=urban_fraction[nodes], dist_fraction=dist_fraction[nodes]
+            )
+
+            heat_dsm_profile = pd.read_csv(
+                snakemake.input.heat_dsm_profile,
+                header=[1],
+                index_col=[0],
```
fneum:
```suggestion
                header=1,
                index_col=0,
```
This should work as well.

config/config.default.yaml
```diff
@@ -565,6 +565,19 @@ sector:
     rural:
     - air
     - ground
+  residential_heat:
+    dsm:
+      enable: false
+      direction: "overheat-undercool"
```
fneum:
```suggestion
      direction:
      - overheat
      - undercool
```
could avoid encoding that as string (implementation already works for this, just the documentation would need to be adjusted).

scripts/prepare_sector_network.py
```diff
@@ -2964,6 +2964,64 @@ def add_heat(
...
+                    "decentral water tank storage", "standing_losses"
+                ]
+                / 100,  # convert %/hour into unit/hour
+                e_cyclic=True,
+                e_nom=e_nom,
+                e_max_pu=e_max_pu,
+                e_min_pu=e_min_pu,
+            )
+
+            logger.info(f"adding heat dsm in {heat_system} heating.")
```
fneum:
```suggestion
            logger.info(f"Adding DSM in {heat_system} heating.")
```

scripts/prepare_sector_network.py
```diff
@@ -2964,6 +2964,64 @@ def add_heat(
...
+                parse_dates=True,
+            )[nodes].reindex(n.snapshots)
+
+            e_nom = (
+                heat_demand[["residential space"]]
+                .T.groupby(level=1)
+                .sum()
+                .T[nodes]
+                .multiply(factor)
+            )
```
fneum, line 2992: Assumes hourly resolution. More robust would be to use `n.snapshot_weightings`.

doc/supply_demand.rst
fneum: I think the standing losses in a hot water storage are likely much higher than in a building.

fneum (approved):

doc/configtables/sector.csv
euronion: I'll revert this: The checkpoint times are adjusted to each buses' timezone, i.e. 10am/10pm are local and not UTC time.

euronion:
Tests are failing because Zenodo retrieval for `seawater_temperature` is not consistent with the data retrieval from #1675 . will be fixed in https://github.com/open-energy-transition/pypsa-eur/pull/67/

fneum:
> Tests are failing because Zenodo retrieval for seawater_temperature is not consistent with the data retrieval from https://github.com/PyPSA/pypsa-eur/pull/1675 . will be fixed in https://github.com/open-energy-transition/pypsa-eur/pull/67

Good but then it's unrelated to this PR.

PR #1865, opened 2025-10-20
add `max_requests_per_second` for storage provider with zenodo

lkstrp:
Follow up on #1861 

Another quick fix: Adds `max_requests_per_second=1.5` to snakemake storage provider requests to zenodo.
Zenodos rate limit is as 133, so 1.5 keeps a large buffer with 90 per min.
```
Response code: 200
Rate limit headers:
{'x-ratelimit-limit': '133',
 'x-ratelimit-remaining': '131',
 'x-ratelimit-reset': '1760970155'}
Rate limit resets at: 2025-10-20 16:22:35
```

To check rate limits:
``` python
import requests
from pprint import pprint
from datetime import datetime

r = requests.get("https://zenodo.org/api/records/15198744")
print(f"Response code: {r.status_code}")
# inspect the response header for X-RateLimit-Remaining
print("Rate limit headers:")
pprint({k: v for k, v in r.headers.items() if "X-RateLimit".casefold() in k.casefold()})

# convert the `x-ratelimit-reset` header to a human-readable timestamp from unix epoch
reset_time = int(r.headers.get("X-RateLimit-Reset", 0))
print(f"Rate limit resets at: {datetime.fromtimestamp(reset_time)}")
```

### DIFF
--- a/rules/retrieve.smk
+++ b/rules/retrieve.smk
@@ -85,7 +85,7 @@ if config["enable"]["retrieve"]:
     rule retrieve_nuts_2013_shapes:
         input:
             shapes=storage(
-                "https://gisco-services.ec.europa.eu/distribution/v2/nuts/download/ref-nuts-2013-03m.geojson.zip"
+                "https://gisco-services.ec.europa.eu/distribution/v2/nuts/download/ref-nuts-2013-03m.geojson.zip",
             ),
         output:
             shapes_level_3="data/nuts/NUTS_RG_03M_2013_4326_LEVL_3.geojson",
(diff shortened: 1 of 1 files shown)


euronion:
1. I don't think this will change much, since the `storage` plugin seems to access Zenodo multiple times
2. In the PR #1675 you can see an implementation that manages that configration centrally, rather than each time `storage` is used.
3. In #1675 coroa suggested something for debugging. I didn't get to it yet. If you have capacity, maybe you can have a look (make the modifications suggested to `storage` and see how often it accesses the remote. Alternatively maybe you can also `wireshark` that information?)

lkstrp:
Ok, so we need #1675 before it becomes any less annoying. Then I'll stop trying to hotfix this

coroa:
> * I don't think this will change much, since the `storage` plugin seems to access Zenodo multiple times

And the head request is not subject to the rate limit.

PR #1874, opened 2025-11-03
Move config dump in `n.meta` to own key

lkstrp:
The full snakemake config is currently saved in `n.meta`. This PR moves it instead to `n.meta['snakemake_config']`. I want to add additional metadata to the Network to use somewhere else, so having it's own key is more organised.

But this is breaking if people rely on it somewhere.

### DIFF
--- a/doc/release_notes.rst
+++ b/doc/release_notes.rst
@@ -9,6 +9,9 @@ Release Notes
 Upcoming Release
 ================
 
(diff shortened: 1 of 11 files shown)


fneum:
Probably wise, but I definitely have used `n.meta` for accessing snakemake configuration options in post-processing scripts in the past. It's only a mild annoyance. Just add a note in release notes.

PR #1879, opened 2025-11-04
fix: fix capital and marginal costs overwrite

### LINKED ISSUE #1877: Marginal and capital costs not overwritten

- [ ] I have confirmed this bug exists on the lastest [release](https://github.com/pypsa/pypsa-eur/releases) of PyPSA-Eur.

### Issue Description

https://github.com/PyPSA/pypsa-eur/pull/1752 introduced a bug for `marginal_cost` and `capital_cost`. These are not being overwritten. This is because they are overwritten before their being assigned in `prepare_costs`.

cc @martacki 

### Reproducible Example

```python
Validate the default values in the processed file
```

### Expected Behavior

Get the default values in the processed file

### Installed Versions

Master

tgilon:
Closes #1877. Fix https://github.com/PyPSA/pypsa-eur/pull/1752.

Ensure that we overwrite costs in two stages and that fill values do not overwrite custom values.

### Test

Using this `custom_costs.csv` file, we cover all cases:

planning_horizon | technology | parameter | value | unit | source | further description
-- | -- | -- | -- | -- | -- | --
all | solar | lifetime | 123456789 | EUR/MWh | test |  
all | solar | FOM | 123456789 | EUR/MWh | test |  
all | offwind | FOM | 123456789 | EUR/MWh | test |  
all | onwind | marginal_cost | 123456789 | EUR/MWh | test |  
all | all | VOM | 123456789 | EUR/MWh | test |  
all | all | capital_cost | 123456789 | EUR/MWh | test |  

We introduce tests before writing the output. The following tests pass:
```python
test_value = 123456789
assert (
    costs_processed.query("technology not in ['H2', 'battery']")["VOM"]
    == test_value
).all()
assert costs_processed.loc["solar", "lifetime"] == test_value
assert costs_processed.loc["solar", "FOM"] == test_value
assert costs_processed.loc["offwind", "FOM"] == test_value

assert costs_processed.loc["onwind", "marginal_cost"] == test_value
assert (costs_processed["capital_cost"] == test_value).all()
```

## Checklist

### DIFF
--- a/scripts/solve_network.py
+++ b/scripts/solve_network.py
@@ -40,7 +40,7 @@ import pandas as pd
 import pypsa
 import xarray as xr
 import yaml
-from linopy.remote.oetc import OetcCredentials, OetcSettings, OetcHandler
+from linopy.remote.oetc import OetcCredentials, OetcHandler, OetcSettings
 from pypsa.descriptors import get_activity_mask
 from pypsa.descriptors import get_switchable_as_dense as get_as_dense
 


rules/collect.smk
```diff
@@ -15,7 +15,12 @@ localrules:
 rule process_costs:
     input:
         lambda w: (
-            [resources(f"costs_{config_provider('costs', 'year')(w)}_processed.csv")]
+            expand(
+                resources(
+                    f"costs_{config_provider('costs', 'year')(w)}_processed.csv"
+                ),
+                run=config["run"]["name"],
```
euronion, line 22: How does adding this line make a difference? Where is the `{run}` placeholder coming from that is `expand`-ed with the value from `config["run"]["name"]`?

scripts/process_cost_data.py
```diff
@@ -35,8 +36,72 @@
...
+            costs = pd.concat([costs, custom_costs.loc[missing_idx]])
+
+        for col in custom_costs.columns:
+            custom_col = custom_costs[col].dropna()
+            costs.loc[custom_col.index, col] = custom_col
+
+        # If technology "all" exists, propagate its parameter values to all other technologies
+        if not all_techs.empty:
+            for param in all_techs.index:
+                costs.loc[:, param] = all_techs.loc[param, "value"]
```
martacki: could we instead make sure to always pass costs in the right format?

scripts/process_cost_data.py
```diff
@@ -51,13 +116,30 @@ def prepare_costs(
...
+            dtype={"planning_horizon": "str"},
+            index_col=["technology", "parameter"],
+        ).query("planning_horizon in [@planning_horizon, 'all']")
+        custom_costs = custom_costs.drop("planning_horizon", axis=1)
+        custom_costs_1 = custom_costs.query(
+            "parameter not in ['marginal_cost', 'capital_cost']"
+        )
+        custom_costs_2 = custom_costs.query(
+            "parameter in ['marginal_cost', 'capital_cost']"
+        )
```
martacki: can we rename 1 + 2 to make it more intuitive, and add a comment why this is needed? (`prepare_costs` overwrite them again otherwise) ... or we add this functionality directly into `prepare_costs` ?

scripts/process_cost_data.py
```diff
@@ -51,13 +116,30 @@ def prepare_costs(
...
+    # Load custom costs
+    if custom_costs_fn is not None:
+        custom_costs = pd.read_csv(
+            snakemake.input.custom_costs,
+            dtype={"planning_horizon": "str"},
+            index_col=["technology", "parameter"],
+        ).query("planning_horizon in [@planning_horizon, 'all']")
+        custom_costs = custom_costs.drop("planning_horizon", axis=1)
+        custom_costs_1 = custom_costs.query(
+            "parameter not in ['marginal_cost', 'capital_cost']"
```
martacki: maybe we could create a variable `prepared_attributes` for this list `["capital_cost", "marginal_cost"]` and the one below `raw_attributes` `["investment", "lifetime", "FOM", "VOM", "efficiency", "fuel", "standing losses"]` (or alike) to avoid hard-coding them twice (prone to error) ?

scripts/process_cost_data.py
martacki: Ok looking at the code, I'd go for rename + comment / explanation

scripts/process_cost_data.py
tgilon: As discussed, we prioritise reducing the number of lines of code rather than optimising the performance of stacking/unstacking. We assume here that `costs` is small and that the impact will be marginal.

scripts/process_cost_data.py
tgilon: Done

scripts/process_cost_data.py
tgilon: Done

scripts/process_cost_data.py
```diff
@@ -35,8 +36,51 @@
...
+    Parameters
+    ----------
+    costs : pd.DataFrame
+        Base cost assumptions.
+    custom_costs : pd.DataFrame
+        Custom cost modifications.
+
+    Returns
+    -------
+    costs : pd.DataFrame
```
euronion:
```suggestion
    pd.DataFrame
```

Here not so relevant, but in general mind that in some cases this method modifies the original `costs` dataframe and returns it, and in other cases it returns a new dataframe:

`custom_costs.empty` -> original dataframe
`len(missing_idx) == 0` -> modified original dataframe
`len(missing_idx) == 0` -> a completely new dataframe

scripts/process_cost_data.py
```diff
@@ -35,8 +36,51 @@
...
+    custom_costs = custom_costs.query("technology != 'all'").dropna(axis=1, how="all")
+
+    # Overwrite unique pairs of (technology, parameter)
+    missing_idx = custom_costs.index.difference(costs.index)
+    if len(missing_idx) > 0:
+        costs = pd.concat([costs, custom_costs.loc[missing_idx]])
+
+    for param in custom_costs.columns:
+        custom_col = custom_costs[param].dropna()
+        costs.loc[custom_col.index, param] = custom_col
```
euronion, line 69: I do not understand the purpose of this code block. Can we have a comment/explanation here?

scripts/process_cost_data.py
```diff
@@ -35,8 +36,51 @@
...
+
+    # If technology "all" exists, propagate its parameter values to all other technologies
+    if not all_techs.empty:
+        for param in all_techs.columns:
```
euronion, line 73: Has the dataframe been transposed such that it is in a wide format (parameters = columns)? Wasn't it a long format before (parameters = index)? Or is the use of `param` here just an unlucky choice, because we have `parameter` also as a column name.

euronion (commented):
I don't really like the two stage approach. Why would that be necessary?
When does a user rather specify `capital_cost` than overwriting `investment` for the technology? Especially since `capital_cost` is usually not an output, but calculated.

And why is `marginal_cost` overwritten after the preparation?

If we go with this two stage approach we'd need a sentence or two also in the docs or the script docstring about the overwrite logic.

scripts/process_cost_data.py
tgilon: I've slightly modified the docstring: 

```
    pd.DataFrame
        Updated cost data with custom modifications applied (if applicable).

scripts/process_cost_data.py
tgilon: Here is the new version

```python
    # Add technologies that don't already exist
    missing_idx = custom_costs.index.difference(costs.index)
    if len(missing_idx) > 0:
        costs = pd.concat([costs, custom_costs.loc[missing_idx]])

    # Overwrite unique pairs of (technology, parameter)
    for param in custom_costs.columns:
        custom_col = custom_costs[param].dropna()
        costs.loc[custom_col.index, param] = custom_col

scripts/process_cost_data.py
tgilon: I decided to transpose to a wide format at the beginning to reduce the number of manipulations: https://github.com/PyPSA/pypsa-eur/blob/3ef8fe573f9f45feb2e494de883378a2e3c19074/scripts/process_cost_data.py#L117-L119

rules/collect.smk
tgilon: The `{run}` placeholder will be introduced by `get_rdir` when scenarios are enabled. This comes from the `resources()` function, which is created by `path_provider()`.

https://github.com/PyPSA/pypsa-eur/blob/3ef8fe573f9f45feb2e494de883378a2e3c19074/scripts/_helpers.py#L55-L68

This addresses Fabian's [comment](https://github.com/PyPSA/pypsa-eur/pull/1752/files#r2481479953) that I wasn't able to push before the PR was merged.

scripts/process_cost_data.py
euronion: ok, thanks

scripts/process_cost_data.py
euronion: I see. Feels a bit strange to deviate from the structure that is used elsewhere, but ok.

euronion:
Ok, LGTM.

> I do agree that `prepare_costs` can cause confusion. However, refactoring it is beyond the scope of this PR.

I'll create an issue for refactoring `preapre_costs`. Could you add a few comments on what you would improve given the time, such that future contributors don't start at 0? Thanks!

tgilon:
> I'll create an issue for refactoring `preapre_costs`. Could you add a few comments on what you would improve given the time, such that future contributors don't start at 0? Thanks!

Done, see my [comment](https://github.com/PyPSA/pypsa-eur/issues/1885#issuecomment-3491095589)

I'm sure @fneum would like to take a look as well.

PR #1883, opened 2025-11-05
Wave and floating solar updates

lmezilis:
Closes # (if applicable).

## Checklist


lmezilis:
Changes in configuration options are added in config/config.electricity.yaml.

fneum:
Great, could you try and resolve the merge conflicts so that we can have a look? Generally, keeping the `diff` as small as possible will help us review.

lmezilis:
Can you let me know if the conflicts are resolved? Apologies for delays, but I am a bit new to github and its procedures.

lkstrp:
@lmezilis They are, but you need to resolve the pre-commit checks now. See the failed check above or run it locally as described in the docs: https://pypsa-eur.readthedocs.io/en/latest/contributing.html

lmezilis:
@lkstrp Ok, I think I completed that, please let me know if I missed anything.

config/test/config.electricity.yaml
fneum, line 1: Keep the changes to an absolute minimum here and try not to change any unrelated config settings. Only changes from `config.default.yaml` should be listed here.

scripts/add_electricity.py
```diff
@@ -174,6 +174,7 @@ def sanitize_carriers(n, config):
             add_missing_carriers(n, c.df.carrier)
 
     carrier_i = n.carriers.index
+    print(n.carriers.index)
```
fneum:
```suggestion
```
no print statements please

scripts/plot_summary.py
```diff
@@ -31,13 +30,17 @@
         "biogas",
         "onshore wind",
         "offshore wind",
-        "offshore wind (AC)",
+        "offshore wind",
```
fneum: Why this change? Now offshore wind is duplicated.

scripts/plot_summary.py
```diff
@@ -31,13 +30,17 @@
...
+        "offshore floating wind",
+        "wave farshore",
+        "wave nearshore",
+        "wave shallow",
         "solar PV",
         "solar thermal",
         "solar rooftop",
         "solar",
-        "building retrofitting",
+        "floating solarbuilding retrofitting",
```
fneum:
```suggestion
        "floating solar",
        "building retrofitting",
```

.gitignore
```diff
@@ -29,6 +29,7 @@ doc/_build
 /scripts/old
 /scripts/create_scenarios.py
 /config/create_scenarios.py
+# /envs/win-64.lock.yaml
```
fneum:
```suggestion
```

scripts/solve_network.py
```diff
@@ -1189,7 +1194,8 @@ def extra_functionality(
     """
     config = n.config
     constraints = config["solving"].get("constraints", {})
-    if constraints["BAU"] and n.generators.p_nom_extendable.any():
+
+    if constraints.get("BAU", False) and n.generators.p_nom_extendable.any():
```
fneum: Change not necessary. `config.default.yaml` allows assuming "BAU" setting exists; no fallback handling required here.

scripts/solve_network.py
```diff
@@ -728,6 +732,7 @@ def add_BAU_constraints(n: pypsa.Network, config: dict) -> None:
     lhs = p_nom.groupby(ext_carrier_i).sum()
     rhs = mincaps[lhs.indexes["carrier"]].rename_axis("carrier")
     n.model.add_constraints(lhs >= rhs, name="bau_mincaps")
+    logger.info("BAU is done.")
```
fneum:
```suggestion
```
not necessary

scripts/prepare_sector_network.py
```diff
@@ -507,6 +507,38 @@ def update_wind_solar_costs(
...
+                + connection_cost
+            )
+
+            logger.info(
+                f"Added connection cost of {connection_cost.min():0.0f}-{connection_cost.max():0.0f} Eur/MW/a to {tech}"
+            )
+
+            n.generators.loc[n.generators.carrier == tech, "capital_cost"] = (
+                capital_cost.rename(index=lambda node: node + " " + tech)
+            )
```
fneum: This code is very similar to the offshore wind handling above. Are you sure this can not be integrated rather than duplicated here? DRY coding principles.

scripts/plot_summary.py
fneum, line 1: For plotting, also colours and nice names must be added to `plotting.default.yaml`.

scripts/add_electricity.py
```diff
@@ -472,15 +472,81 @@ def attach_wind_and_solar(
                         + costs.at[car + "-station", "capital_cost"]
                         + connection_cost
                     )
-                else:
-                    capital_cost = (
-                        costs.at["offwind", "capital_cost"]
-                        + costs.at[car + "-station", "capital_cost"]
-                        + connection_cost
-                    )
+            if car == "offsolar":
```
fneum: The code in this block contains a lot of duplications per carrier. Are you sure this can not be integrated rather than duplicated here? DRY coding principles.

rules/build_electricity.smk
```diff
@@ -220,7 +220,7 @@ rule determine_availability_matrix_MD_UA:
         wdpa="data/WDPA.gpkg",
         wdpa_marine="data/WDPA_WDOECM_marine.gpkg",
         gebco=lambda w: (
-            "data/bundle/gebco/GEBCO_2014_2D.nc"
+            "data/bundle/gebco/GEBCO_2023_2D.nc"
```
fneum: This change in dataset still needs to be properly handled. See https://pypsa-eur.readthedocs.io/en/latest/data_sources.html.

In particulary, `data/versions.py` must be adapted:

https://github.com/PyPSA/pypsa-eur/blob/73fb005525ae4a372df0325c185c7e405603a542/data/versions.csv#L60-L61

Additionally, the retrieval rule (download) must be updated correspondingly.

Could you also explain why the update is necessary? Is it because the resolution was upgraded from 30 to 15 arc-seconds? Does it have any impact?

Could we also directly update to the 2025 version while we are at it? https://www.gebco.net/data-products-gridded-bathymetry-data/gebco2025-grid

Note, that the archive version should have a spatially reduced scope to Europe. It is unnecessary to download the full 4 GB of global bathymetry for a European model. For the required total bounds, see the current GEBCO archived data.

fneum (changes requested):
Looks mostly good. Some more cleaning up and dataset handling required before this can be merged. This will also require an update of [`pixi.toml`](https://github.com/PyPSA/pypsa-eur/blob/master/pixi.toml) as this feature depends on a new `atlite` version: 

https://github.com/PyPSA/atlite/pull/475

config/test/config.electricity.yaml
lmezilis: Yes of course. Most of the changes were made for the convenience of the studies at that time but I will try my best to correct everything.

rules/build_electricity.smk
lmezilis: From the start we were using the 2023 dataset but indeed there is no need for it, the influence is completely insignificant. There is no need to use it for the purpose of PyPSA. 

We thought that by using wave devices this could make a difference. But the GEBCO cells are already so small that they are not affected by the datasets we were using, let alone ERA5. 

I fetched all the upstream commits to be updated, and I think in build_electricity.smk now there is no reason to change anything.

scripts/add_electricity.py
lmezilis: I will make this as tidy as possible

scripts/plot_summary.py
lmezilis: Ok thank you!

scripts/prepare_sector_network.py
lmezilis: I will integrate it.

scripts/add_electricity.py
```diff
@@ -466,21 +466,22 @@ def attach_wind_and_solar(
 
                 # Take 'offwind-float' capital cost for 'float', and 'offwind' capital cost for the rest ('ac' and 'dc')
                 midcar = car.split("-", 2)[1]
-                if midcar == "float":
+                if supcar == "offwind" and midcar != "float":
```
lmezilis, line 531: I decided to make it the other way around since now the "offwind-ac" and "offwind-dc" are more unique. Let me know if I overlook anything.

scripts/plot_summary.py
lmezilis: Yes you are right, we were using mostly one type offwind and at that time we did not want the "(AC)" to apear in the summaries. I have reverted this.

lmezilis:
@fneum Hi Fabian, I have tried to resolve the integration (ubuntu or macos) but I cannot tell what is wrong with it. I tried to update the merge just in case that is the problem but please take a look if you have time.

I decided to keep very basic changes on the `config.electricity.yaml` file, as it seems that it is only used for testing. Do you want me to pass all `renewable `changes to the `config.default.yaml`? I can do that as the last step of this merge. Thank you.

PR #1886, opened 2025-11-06
Update dependency management to `pixi`

### LINKED ISSUE #1845: Use conda explicit file specification format to pin dependencies
The current implementation uses conda lock files to pin dependencies. Snakemake doesn't "see" these, though, so if you're using the `conda` directive in a rule (and the `conda` software deployment method), these lockfiles are meaningless.

Snakemake _can_ see pinned conda dependencies if stored in the explicit file specification format, as described in [their documentation](https://snakemake.readthedocs.io/en/stable/snakefiles/deployment.html#freezing-environments-to-exactly-pinned-packages).

This would require updating the git workflow `update-pinned-env.yaml` to call `snakedeploy update-conda-envs envs/environment.yaml` on a matrix of different operating systems (it will always create the pin file for the operating system it is run from.

The resulting files will then be picked up automatically by snakemake when using the `conda` software deployment method.

brynpickering:
Closes #1845 

The main benefit is simplifying the dependency lockfile, which comes as standard with `pixi`.
The added benefit is being able to bring together default/docs/test dependencies. For forks, they can easily extend dependencies in their own `feature` list, so as to clearly differentiate between upstream and additional project deps.

I've kept the process of updating the lockfile in a GitHub workflow. So long as the user doesn't change the dependencies in `pixi.toml`, their working environment will be installed from the lock file and will therefore get env reproducibility automatically.

## Why not `uv`?

These projects leverage a lot of binaries that can be a pain / impossible to install from PyPI.
Pixi will use `uv` under the hood whenever a PyPI dependency is added `pixi add --pypi ...` and most of the rest of the functionality is the equivalent (tasks, dependency groups, lightning fast dependency resolution, etc.).

## What about conda installs?

I've left conda env files in `envs/` so that those who cannot use `pixi` on their machines for some reason can still install a working environment.

- Remove `doc/requirements.txt` in favour of pixi for package management
- Keep `envs/*` but generate files automatically
- Use conda explicit file specification for platform-specific lock files in `envs/*`
- Move any PyPI deps that are available on a non-default Anaconda channel over to being a conda dep (i.e., all of them).
- Move makefile rules to being pixi tasks
- Remove all reference to the `conda` directive in snakemake rules. This didn't add value in practice anyway since there is only one working environment and there's no use of `conda` as the snakemake software deployment method
- Update RTD docs build to use pixi
- Update dockerfile to use pixi

## Checklist

### DIFF
--- a/.pixi/config.toml
+++ b/.pixi/config.toml
@@ -3,3 +3,4 @@
 # SPDX-License-Identifier: CC0-1.0
 
 pinning-strategy = "latest-up"
+run-post-link-scripts = "false" # set to "insecure" to allow running post-link scripts
--- a/pixi.toml
+++ b/pixi.toml
@@ -96,7 +96,6 @@ yaml = ">=0.2.5"
 [feature.doc.tasks.build-docs]
 args = ["dir", {"arg" = "output", "default" = "html"}]
 cmd = "dot -c && sphinx-build -T -b {{ output }} doc {{ dir }}/{{ output }} "
-depends-on = ["update-dags"]
 
 [feature.doc.dependencies]
 atlite = ">=0.2.9"


.readthedocs.yml
```diff
@@ -7,13 +7,18 @@ version: 2
...
+  os: ubuntu-24.04
   tools:
-    python: "3.11"
-  apt_packages:
-  - graphviz
-
-python:
-  install:
-  - requirements: doc/requirements.txt
+    python: "3.10"
```
euronion: Intentional downgrade?

pixi.lock
euronion, line 1: When should the `.lock` file be recreated?

.github/pull_request_template.md
```diff
@@ -7,7 +7,7 @@ Closes # (if applicable).
 
 - [ ] I tested my contribution locally and it works as intended.
 - [ ] Code and workflow changes are sufficiently documented.
-- [ ] Changed dependencies are added to `envs/environment.yaml`.
+- [ ] Changed dependencies are added to `pixi.toml`.
```
euronion: Should the file be manually edited or similar to `uv` using `uv add` (what's the `pixi` equivalent?)

euronion:
I like it!
* How to use this e.g. on a cluster environment in combination with `snakemake`?
* Can this be used in an enterprise environment on Windows, where e.g. dependency installation is difficult? Is there a fallback of using `conda` for existing users that have fought with their IT department to get `conda` installed, or do they need to switch to `pixi`?
* I believe the instructions for users should be a bit more consistent: The default should either be `pixi run snakemake ...` or `pixi shell` and then only `snakemake ...`, not a mix.
* Maybe for the contributing section, add info that if you fork it, you can add additional dependencies as features + how to install those.

lkstrp:
I need to test this because I've never used Pixi before, but what you describe sounds great!

> Can this be used in an enterprise environment on Windows, where e.g. dependency installation is difficult? Is there a fallback of using conda for existing users that have fought with their IT department to get conda installed, or do they need to switch to pixi?

But I think this is currently a deal breaker. As far as I understand, this is fully breaking any existing conda/ micromamba setup (so every setup). And not just breaking, people need to come up with yet another tool they often never used. If `pixi` works as expected and brings those benefits, I am happy to use it as default. But can we not just keep support for multiple package managers? Which means: keep conda, still mention it in the docs as one option, and keep the lock file management for conda as well?

And is this even closing #1845? Can snakemake see pixi locked environments? It's not [mentioned](https://snakemake.readthedocs.io/en/stable/snakefiles/deployment.html#freezing-environments-to-exactly-pinned-packages)

.github/pull_request_template.md
brynpickering: Using `pixi add`

.readthedocs.yml
brynpickering: Ah, not intentional. When I imported the docs reqs it set the python upper pin to "<3.11" which I assumed was dictated by the dependencies themselves. I'll see what happens when I unpin and let it run free.

pixi.lock
brynpickering: I created a github action to update it at the same cadence as the previous lock files (periodically & when `pixi.toml` is edited)

brynpickering:
Fair points @euronion @lkstrp ! As you can see in one of the github actions, we can export the dependencies to a conda-compatible version. Even better would be to export to [conda explicit spec files](https://pixi.sh/dev/reference/cli/pixi/workspace/export/conda-explicit-spec/) since then you get the equivalent of lock files that are conda/micromamba installable. They can also be used to solve #1845 directly (I have said it closes it as it feels like the best option available under the circumstances - will comment directly in that issue on why). 

This would work as we have all our dependencies coming from conda now. If any were added coming from PyPI exclusively then we'd have a few options:

1. revert to the existing method of using `conda-lock` and then export [a basic environment file from pixi](https://pixi.sh/dev/reference/cli/pixi/workspace/export/conda-environment/#global-options) and apply conda-lock to it. 
2. Fix dependencies more explicitly within pixi rather than going for `latest-up` so that an export of the environment has fixed dependencies by default, making them pseudo lock-files (they wouldn't lock dependencies of dependencies). 
3. Use [a third-part dependency](https://github.com/basnijholt/pixi-to-conda-lock) to go straight from a pixi.lock file to conda lock files, ignoring an `environment.yaml` file completely.

My suggestion would be to have a github action to export conda explicit files each time the pixi.lock is updated. These can then be used to create a working environment directly without installing `pixi` and could be the basis for rule-based containerisation, since snakemake is able to work with these `explicit` file specs.

brynpickering:
`pixi` takes about half a second to create spec files for all environments and all platforms, so seems like an easy way to provide the `conda` fallback option since they can be used directly to create an environment:

```fish
## this would be run as part of .github/workflows/update-lockfile.yaml
pixi workspace export conda-explicit-spec -e default envs --frozen 
for f in envs/*_conda_spec*; do mv "$f" "${f/_conda_spec/.pin}"; done  # for snakemake rule containerisation compatibility, the `txt` file would need to be updated to change `_conda_spec.txt` to `.pin.txt`.
##
conda create -n pypsa-eur --file envs/default_osx-arm64.pin.txt 
```

.github/workflows/test.yaml
```diff
@@ -37,29 +34,14 @@ jobs:
...
-          echo "Base env ${{ env.BASE_ENV }} not changed. Using locked envs."
-        fi
+    - name: Setup Pixi
+      uses: prefix-dev/setup-pixi@v0.9.1
+      with:
+        pixi-version: v0.44.0
+        frozen: true # use the lockfile as-is
+        cache: true
+        # Do not cache in branches
+        cache-write: ${{ github.event_name == 'push' && github.ref_name == 'main' }}
```
lkstrp, line 43: Previously if `environments.yaml` was changed, the CI installed from there instead of from the pinned environments to install any added dependencies. What happens now if you would update `pixi.toml`?

Otherwise the CI is so much cleaner, I love it.

.github/workflows/validate.yaml
```diff
@@ -17,12 +17,24 @@ jobs:
...
+    - name: Setup Pixi
+      uses: prefix-dev/setup-pixi@v0.9.1
+      with:
+        pixi-version: v0.44.0
+        frozen: true # use the lockfile as-is
+        cache: false
+
+    - name: Create environment file
+      run: |
+        pixi workspace export conda-environment environment.yaml
```
lkstrp: Just ignore `validate.yaml` it's not running currently anyways and I will bring an update soon.

doc/installation.rst
```diff
@@ -26,38 +26,19 @@ First of all, clone the `PyPSA-Eur repository <https://github.com/PyPSA/pypsa-eu
...
-    $ conda update conda
+    $ pixi install --frozen
 
-    $ conda env create -f envs/linux-64.lock.yaml # select the appropriate file for your platform
+This will install the dependencies for you to begin working on your project.
+`--frozen` is important to ensure you are getting the exact versions of every dependency, required for reproducibility.
 
-    $ conda activate pypsa-eur
+Then, prepend all your calls with `pixi run` (e.g. `pixi run snakemake -call -n`).
+Alternatively, you can call `pixi shell` in your terminal to activate your working environment, then continue as normal.
```
lkstrp: As discussed, keep docs for `conda` but add `pixi` as the preferred alternative. I'm not sure if we even need to mention `pixi install --frozen`. The user doesn't have to worry about dependencies at all, that's the charm isn't it?
1. Clone pypsa-eur 
2. run pixi run snakemake ...
3. Done

euronion:
> The pypi exclusive bridge we can cross when it's needed (and can't be avoided). We had via pip:
> 
> ```
>   - gurobipy
>   - highspy
>   - tsam>=2.3.1
>   - entsoe-py
> ```
> 
> which isn't recommended in the first place and I don't recall why they even have been there. Probably at some point not available via conda forge.

IIRC we had

* a patched version of `tsam` at some point and then problems with the `conda-forge` version
* `gurobipy` and `highspy` because it was not always available for MacOS on `conda-forge` (but was on PyPI)
* `entsoe-py` was according to `conda-forge` not updated as often as on PyPi, so I guess @FabianHofmann added it because there was a more up-to-date version available through `pypi` back then.

brynpickering:
> gurobipy and highspy because it was not always available for MacOS on conda-forge (but was on PyPI)

Yeah they probably got added around the time the changeover to arm64 was happening, which did cause quite some issues at the time!

doc/installation.rst
brynpickering: yeah, ideally the first install is against the current lock file. By default, I think the lock file is refreshed on first run, `--frozen` forces the lockfile to be used as-is. so `pixi shell --frozen`, `pixi install --frozen` or `pixi run --frozen` would all do the job I believe. There's also an environment variable that can be set to always set the `frozen` flag, but I'm not sure how that could be embedded into the repository by default...

coroa:
I am not sure about the need for `--frozen`. My understanding is that as long as the lockfile still satisfies the constraints in the pixi.toml manifest, then the lockfile is not updated. Ie you only need --frozen if you manually made a change to the dependencies in `pixi.toml` and you don't want to realize those. To update packages to new versions, you either have to delete the lockfile or use `pixi update`.

.github/workflows/test.yaml
brynpickering: We can just set `frozen` to False to allow it to do that. I'll update.

brynpickering:
@lkstrp I've added the explicit files in to allow users to run a "legacy" conda install. It feels dangerous to try to re-create the same lockfiles as the approach would be to (a) dump the pixi toml deps to a simple environment file then (b) create the lock files from that using `conda-lock`. This means resolving the dependencies using `conda-lock` separately to `pixi`. I'm sure they would resolve to the same exact dependency versions, but I can't be _certain_ about that. The explicit pin files act just like the conda lock files from the perspective of the user (they can install an environment with them using `conda`).

What do you prefer, use the explicit pin files or revert to the conda lock files and accept potential lock de-sync between pixi and conda-lock?

lkstrp:
Hmm, unsynced pins don't sound like a good idea and we want to use the Pixi based explicit pin file export. 
My main concern is that we don't wanna immediately break
``` bash
mm env create -f envs/osx-arm64.lock.yaml
mm env create -f envs/environment.yaml
```
for everyone with this PR. Without giving people time to adjust to either pixi or conda explicit lock files. 

You can't even create the unpinned `conda` environment right now. Is there a way to export the `environment.yaml` file from `pixi` @brynpickering? So we don't have to maintain  `environment.yaml` as well as `pixi.toml`? There should still be at least a way to get an unpinned `conda` environment.

For the pinned `conda` environments, you can use the same command, but a different file, and you would need to pass an environment name. We cannot do something hacky/ temporary like copying/ creating a symlink of `default_osx-arm64.pin.txt` to `osx-arm64.lock.yaml`, because the suffix doesn't match, even if the command is the same.

If nobody comes up with another idea, we can either:
-  Break right now, remove old lock files and people need to use pixi or the new explicit lock files right now
- Or not revert to the conda lock files, but keep creating an unsynced second version via `conda-lock` for a couple of weeks, before we remove them

I think we should go with the latter, but maybe that's my soft PyPSA Framework view. Any other opinions? @fneum 

**EDIT:**
Let's keep things simple and kick `conda-lock` and the old pinned versions. What we still need is a functional `environment.yaml` next to the new explicit pins. That's currently still missing.

fneum:
> Let's keep things simple and kick conda-lock and the old pinned versions. What we still need is a functional environment.yaml next to the new explicit pins. That's currently still missing.

I agree with this.

.github/workflows/validate.yaml
brynpickering: reverted

.github/workflows/test.yaml
brynpickering: updated

doc/installation.rst
brynpickering: Updated following @coroa's comment (that `--frozen` shouldn't be needed so long as the pixi.toml hasn't changed)

brynpickering:
@lkstrp I've re-introduced the environment.yaml, updated the docs to reference it, and added it to be updated when the lockfile is updated.

The one thing that might need thought is whether to add `pixi run` to all terminal commands in the docs. It could be that we recommend users run `pixi shell` (or set it to happen automatically within their IDE, e.g. [VSCode](https://pixi.sh/dev/integration/editor/vscode/) or [Zed](https://pixi.sh/dev/integration/editor/zed/)), then they can use all the terminal commands without needing to prepend with `pixi run`.

lkstrp:
> It could be that we recommend users run pixi shell (or set it to happen automatically within their IDE, e.g. [VSCode](https://pixi.sh/dev/integration/editor/vscode/) or [Zed](https://pixi.sh/dev/integration/editor/zed/)), then they can use all the terminal commands without needing to prepend with pixi run.

Great! Can we add the same links to the docs? `pixi shell` is great, so no prefix in the docs. My point is just that it should be clear to the user that we are taking the burden of choosing an environment away from them. They just need to use `pixi shell` or `pixi run`, and life is good.

fneum:
Wow, just tested locally and it works really well and fast! I approve!

I have only one question following up on the `pip` dependencies: As @euronion mentioned, it can sometimes happen that new package versions (and their fixes) are available faster via `pip`. That's for sure one use case that will likely reappear at some point. The other use case would be for installing a development version (e.g. PyPSA from some branch). Do I understand it correctly that this can also be done with `pixi add`?

https://pixi.sh/v0.48.2/reference/cli/pixi/add/#pixi-add

```sh
pixi add --pypi gurobipy
pixi add --git https://github.com/pypsa/pypsa --branch feature
```

brynpickering:
@fneum 

> it can sometimes happen that new package versions (and their fixes) are available faster via pip

Yeah, although it's easy enough to just become a conda-forge feedstock maintainer if it's a package that you want to expedite from PyPI to conda-forge ;)

You can install PyPI dependencies as you suggest, yes, particularly in spin-off / dev projects. The main reason* to try and keep conda-only is to be able to generate the conda explicit lock files (they can't handle PyPI dependencies). However, if we expect to transition to 100% pixi eventually then we can add PyPI deps to the core dep list once/if we remove support for conda direct install (or revert to the method of creating a standard conda env and creating a lock file from there using `conda-lock`).

(*) the other reason is that PyPI dependencies can sometimes trip up reproducibility as the conda lock file doesn't necessarily catch all dependencies of dependencies of PyPI deps. You get safer reproducibility if sticking with one package index.

brynpickering:
@lkstrp is the dockerfile still in use? I have fixed it up to use pixi (and to not build with the full repo when it only ever copies across the env setup) but not sure it's even necessary...

brynpickering:
@lkstrp I _think_ I've resolved all comments now (and tests should pass now that the pixi version in the CI is updated; was very outdated). Let me know if anything is missing.

One thing to note is that I've made the DAG-updater task (that @euronion introduced recently) a mandatory step before building the docs. This might not be desired _every time_ the docs are built locally. We could simply add the running of that task to the RTD build step and then the new DAGs would live only in the docs themselves rather than being updated within the repo directly. Or I can revert to having the DAGs built in a periodic github action, as @euronion originally introduced.

coroa:
@brynpickering Can you fix up the main PR description above for the final status? ie. for instance it still says, that envs/* is removed. Thanks

euronion:
> One thing to note is that I've made the DAG-updater task (that @euronion introduced recently) a mandatory step before building the docs. This might not be desired _every time_ the docs are built locally. We could simply add the running of that task to the RTD build step and then the new DAGs would live only in the docs themselves rather than being updated within the repo directly. Or I can revert to having the DAGs built in a periodic github action, as @euronion originally introduced.

I would not tightly link these two tasks to each other:
Building the dag can take some time, making building the docs, e.g. locally difficult.
The DAG pictures only need to be updated before a release and only once, so that's more efficient IMHO.

brynpickering:
@coroa @euronion done and done

brynpickering:
@euronion one argument I would make for building the DAGs more frequently is that the `latest` docs may have outdated DAGs at times. This would be a reason to run the DAG build on any push to `master`, rather than only on tags.

euronion:
> @euronion one argument I would make for building the DAGs more frequently is that the `latest` docs may have outdated DAGs at times. This would be a reason to run the DAG build on any push to `master`, rather than only on tags.

I don't mind that either, but I wouldn't link it in the `pixi` task. Rather add it to the CI with event trigger on PR to master or merge on master. @lkstrp did the implementation in the end, I'll let him judge.

lkstrp:
> I don't mind that either, but I wouldn't link it in the pixi task.

Same here

> We could simply add the running of that task to the RTD build step and then the new DAGs would live only in the docs themselves rather than being updated within the repo directly.

This is great, but you would still need the dag in the repo so that the local build can find it. Which means you end up with two processes to have an updated dag which is kind of overkill.

Also, the quality of the docs isn't very high right now. I would say we leave it as it is, merge this, and may introduce a dag build per commit for the docs in future, when we do a general overhaul (e.g. moving to markdown).

So I am happy to have this merged, if everyone agrees. Thanks a lot @brynpickering !

euronion:
> Do I understand it correctly that this can also be done with `pixi add`?
> 
> https://pixi.sh/v0.48.2/reference/cli/pixi/add/#pixi-add
> 
> ```shell
> pixi add --pypi gurobipy
> pixi add --git https://github.com/pypsa/pypsa --branch feature
> ```

Note for prosperity:
Adding a `git` source doesn't work as simply as this. To add the `snakemake-storage-plugin-cached-http`, I needed to call:
```bash
pixi add --pypi "snakemake-storage-plugin-cached-http @ git+https://github.com/PyPSA/snakemake-storage-plugin-cached-http.git"
```

PR #1893, opened 2025-11-10
Refactor: update heat source handling, introduce PyPS-onic pre-heating for geothermal and PTES

amos-schledorn:
Supersedes #1765 .

Following a discussion with @fneum, this feature is being rehauled and its description updated. For reference, this is the "old" description of its previous architecture:

Supersedes #1765 .

This feature implements some refactoring of how **Pit Thermal Energy Storage (PTES)** and **heat sources** are handled.
PTES boosting is now possible without custom constraints. Instead, energy balancing is handled more PyPS-onicly (yes, that's a word now!) through a few extra busses and links.

Geothermal boosting is introduced similarly with a preheater to better approximate real-world behaviour.
This results in an overall cleaner/more generic integration of heat sources.

## 1. New `HeatSource` Enum Class (`scripts/definitions/heat_source.py`)

A new enumeration class has been added to categorize heat sources:

| Heat Source | Properties |
|-------------|------------|
| **AIR, GROUND, SEA_WATER** | Inexhaustible sources - no resource bus needed |
| **GEOTHERMAL, RIVER_WATER** | Limited sources requiring a generator and resource bus |
| **PTES** | Special case - storage discharge treated as heat source |

Key properties for each source:
- `requires_bus`: Whether a resource bus is needed (True for limited sources)
- `requires_generator`: Whether a generator component is needed (geothermal, river_water)
- `has_constant_temperature`: Whether temperature is time-invariant (geothermal)

The class also provides methods for:
- Getting capital costs and lifetimes
- Determining heat pump bus2 connections
- Calculating efficiency2 for heat pumps (accounts for heat drawn from source)
- Generating carrier and bus names

## 2. `build_ptes_operations` Rule

**New configuration parameters:**
```yaml
sector:
  district_heating:
    ptes:
      enable: true
      temperature_dependent_capacity: false  # e_max_pu varies with ΔT
      charge_boosting_required: false  # NOT IMPLEMENTED
      discharge_resistive_boosting: false  # use resistive heaters vs heat pumps
      top_temperature: 90  # or "forward" for dynamic
      bottom_temperature: 10  # or "return" for dynamic
      design_top_temperature: 90
      design_bottom_temperature: 35
```

**Outputs:**
1. `temp_ptes_top_profiles`: PTES top layer temperature profile
2. `ptes_e_max_pu_profiles`: Normalized storage capacity (accounts for temperature-dependent capacity)
3. `ptes_boost_per_discharge_profiles`: Ratio of boost energy to discharge energy when T_forward > T_top. Used iff `sector:district_heating:ptes:discharge_resistive_boosting: true`

## 3. `build_heat_source_utilisation_profiles` Rule

Renamed from `build_heat_source_direct_utilisation_profiles`. Calculates when heat sources can be used and in what mode:

**Outputs:**
1. `heat_source_direct_utilisation_profiles`: Binary (1.0 when T_source ≥ T_forward for direct supply) - this already exists
2. `heat_source_preheater_utilisation_profiles`: Efficiency when T_return < T_source < T_forward (preheating mode) - this is new

### a) For exhaustible heat sources (currently geothermal, river-water and PTES with HP boosting but more to come in #1944)

```
Generator (only geothermal / river_water)
    │
    ▼
┌──────────────────┐
│   Resource Bus   │
└────────┬─────────┘
         │
         │  Utilisation Link
         │
    ┌────┴────────────────────────────┐
    │                                 │
    │ eff = 1 if T_source ≥ T_forward │ eff2 = 1 - eff
    │        else 0                   │
    ▼                                 ▼
urban                          Preheater-Input Bus
central                              │
heat                                 │  Preheater Link
                                     │
                ┌────────────────────┴────────────────────────┐
                │                                             │
                │ eff = T_source - T_return                            │ eff2 = 1 - eff
                │       ─────────────────────────             │
                │       (T_source - T_return) + ΔT_cooling    │
                ▼                                             ▼
            urban                                      Heat-Pump-Input Bus
            central                                          │
            heat                                             │
                                                             │  Heat Pump Link
                                                             │
                                        ┌────────────────────┴───┐
                                        ▲                        │
                                        │ eff = 1/COP            │ eff2 = 1 - 1/COP
                                        │                        ▼
                                Electricity                  urban
                                   Bus                       central
                                                             heat
```

For PTES, the generator is replaced with the `urban central water pits discharger` link.

### b) For Inexhaustible Sources (air / ground / sea_water)

**No buses created** — these sources don't need resource tracking, thus
- `heat_source.requires_bus` returns `False`
- `heat_source.get_heat_pump_input_bus()` returns `""`, which is also `bus2` for heat pumps.

**Links created:**

1. **Heat Pump Link**: Electricity → urban central heat
   - bus0: urban central heat (inverted link configuration)
   - bus1: Electricity Bus (eff = 1/COP)
   - bus2: not used (empty string)
   - No resource bus needed since the source is inexhaustible

```
                          Heat Pump Link
                                │
               ┌────────────────┴───┐
               ▲                    │
               │ eff = 1/COP        │ (no eff2 / bus2)
               │                    ▼
       Electricity              urban
          Bus                   central
                                heat
```

### c) PTES with Resistive Boosting

When `ptes:discharge_resistive_boosting: true`, PTES uses resistive heaters instead of heat pumps for the temperature lift. The structure bypasses the cascading utilisation/preheater/heat-pump chain.

**Key calculation (in `build_ptes_operations`):**
- **α = boost_per_discharge = (T_forward - T_top) / (T_top - T_bottom)**
- This is the ratio of boost energy needed relative to PTES discharge energy
- The `water pits discharger` link still points to the PTES resource bus `urban central ptes heat`
- Preheater and direct utilisation busses and links are created but `heat_source.requires_heat_pump()` returns `False`.
- Since no heat pump is created, all PTES heat is either routed through direct utilisation or resistive boosting, not through preheating.

```
                  ┌──────────────────┐
                  │   PTES Store     │
                  │   (water pits)   │
                  └────────┬─────────┘
                           │
                           │  Discharger Link                            
                          ▼
                  ┌─────────-────────┐
                  │   Resource Bus   │----> Utilisation Link -----> Urban central heat
                  │   (ptes heat)    │
                  └────────┬─────────┘
                         │
                         │  Resistive Booster Link
                         │
                        ┌────────────┴────────────────────┐
                        ▲                                 │
                        │ eff = α / (α + 1)               │ eff2 = 1 / (α + 1)
                        │                                 ▼
                 Resistive Heat                       urban
                      Bus  ──────────────┐            central
                        │                │            heat
                        │                │              ▲
                        │                │              │
           Resistive Heater        Stand-alone Link     │
           eff = 1                 eff = 1              │
                ▲                        │              │
                │                        └──────────────┘
         Electricity
            Bus
```

With resistive boosting, the ratio α determines how much of the output comes from PTES discharge vs electric heating:
- **eff2 = 1/(α+1)**: use 1 unit of PTES heat to produce α+1 units of district heating
- **eff = α/(α+1) = 1 - eff2**: energy conservation

## 5. Further onfiguration Changes

- Renamed `heat_pump_sources` → `heat_sources`
- Renamed `tes` → `ttes` (tank thermal energy storage)
- Removed `limited_heat_sources`, `direct_utilisation_heat_sources`, `temperature_limited_stores`

## Open issues
- [x] integration into brownfield setups
- [x] decide upon default behaviour (disable PTES/enable PTES with/without boosting; temperatures, ...)
- [x] add bus structure sketches to docs
- [x] calculation of pre-heating ratios needs reviewing

## Checklist

### Config

```run:
  name: 
  - ptes_90-10_dynamic_capacity_hpBoosting
  - ptes_fwd-ret_dynamic_capacity_hpBoosting
  - ptes_90-35_static_capacity_rhBoosting
  - ptes_90-10_dynamic_capacity_rhBoosting_geothermal
  - ptes_200-10_dynamic_capacity_hpBoosting
  scenarios:
    enable: true
    file: dev/scenarios.yaml
    shared_resources:
      policy: true

foresight: overnight

scenario:
  clusters:
  - 8
  planning_horizons:
  - 2040

clustering:
  temporal:
    resolution_sector: 25h

countries: ['DE', 'DK']

plotting:
  energy_threshold: 0
  enable_heat_source_maps: true
  interactive_bus_balance: 
    bus_name_pattern: "*urban central*"

sector:
  district_heating:
    supply_temperature_approximation:
      max_forward_temperature_baseyear:
        DE: 110
        DK: 80
      min_forward_temperature_baseyear:
        DE: 75
        DK: 55
      return_temperature_baseyear:
        DE: 60
        DK: 50
      relative_annual_temperature_reduction: 0.0
  ttes: false

##### scenarios.yaml #####


ptes_90-10_dynamic_capacity_hpBoosting:
  sector:
    heat_sources:
      urban central:
        - air
        - ptes
    district_heating:
      ptes:
        top_temperature: 90
        bottom_temperature: 10
        temperature_dependent_capacity: true
        discharge_resistive_boosting: false

ptes_fwd-ret_dynamic_capacity_hpBoosting:
  sector:
    heat_sources:
      urban central:
        - air
        - ptes
    district_heating:
      ptes:
        top_temperature: forward
        bottom_temperature: return
        temperature_dependent_capacity: true
        discharge_resistive_boosting: false

ptes_90-35_static_capacity_rhBoosting:
  sector:
    heat_sources:
      urban central:
        - air
        - ptes
    district_heating:
      ptes:
        top_temperature: 90
        bottom_temperature: 35
        temperature_dependent_capacity: false
        discharge_resistive_boosting: true

ptes_90-10_dynamic_capacity_rhBoosting_geothermal:
  sector:
    heat_sources:
      urban central:
        - air
        - ptes
        - geothermal
    district_heating:
      ptes:
        top_temperature: 90
        bottom_temperature: 10
        temperature_dependent_capacity: true
        discharge_resistive_boosting: true


ptes_200-10_dynamic_capacity_hpBoosting:
  sector:
    heat_sources:
      urban central:
        - air
        - ptes
    district_heating:
      ptes:
        top_temperature: 200
        bottom_temperature: 10
        temperature_dependent_capacity: true
        discharge_resistive_boosting: false
```
#### Parameters
(look reasonable)
image

##### `DE0 1 urban central ptes heat` bus
PTES discharge: 369 MWh

image

#####  `DE0 1 urban central ptes medium-temperature heat` bus
Discharge is fully absorbed by pre-heater, since no direct utilisation allowed

image

#####  `DE0 1 urban central ptes return-temperature heat` bus
231 MWh is used as heat pump input
 
image

#####  `DE0 1 urban central heat` bus
Remaining 138 MWh are used by preheater to heat return flow
Heat pump output is 282 MWh > 231 MWh due to  electricity input

image

#### Entire year, `ptes_90-10_dynamic_capacity_hpBoosting`, `DE0 1`
Discharging over the entire year follows parameters shown above, with direct utilisation mostly in the summer.
There are some periods where both direct utilisation and pre-heating are happening due to the `25h` resolution.

image

#### Entire year, `ptes_90-10_dynamic_capacity_hpBoosting`, `DK1 0`
Direct utilisation is possible throughout the entire year due to lower forward temperatures.

image

#### Entire year, `ptes_90-10_dynamic_capacity_rhBoosting_geothermal`, `DE0 0`
Here, the pre-heating ratio for geothermal seems a little high but matches the logic, as `(T_source - T_return) / (T_source - T_return + 6K) = 0.45`.

image

## EDIT
I've also updated `build_geothermal_heat_source_potentials` by introducing scaling of Manz et al.'s data.
Manz et al. assume a difference of 15K. This rule now scales the potentials based our actual temperature delta:

a) If source_temperature > forward_temperature (direct utilisation):
   scale_factor = (source_temperature - return_temperature) / 15 K

b) If forward_temperature >= source_temperature > return_temperature (preheating):
   scale_factor = (source_temperature - return_temperature + heat_source_cooling) / 15 K

c) If source_temperature <= return_temperature (heat pump only):
   scale_factor = heat_source_cooling / 15 K

## EDIT 2
The example results are using a buggy scaling of `p_max_pu` for geothermal heating. I've fixed that bug resulting in the model not building goethermal heating anymore. I've checked that it can (by lowering CAPEX) and pre-heating ratios are still valid. 
(just a note that when executing these configs, geothermal has to be forced in to get comparable results, so I'm leaving the "faulty" results here)

## EDIT 3
Following a suggestion of @fneum's, I've compared a scenario without PTES & geothermal heating, i.e. those technologies that should be affected by the feature, to the master:
```
only_unaffected_technologies:
  sector:
    district_heating:
      ptes:
        enable: false
    heat_sources:
      urban central:
        - air
        - river_water
        - sea_water
    ttes: true
  solving:
    options:
      noisy_costs: false
```
On the master, I have removed the section in `prepare_sector_network` that builds PTES. 
-> The objective value is constant across both feature and master to the 5th digit:
```feature=1.16952e+11 master=1.16952e+11  (Δ=57755.6, +0.000%)```

I've also compared runs with default settings in the feature and master (still DE/DK-25h) with similar results (different colours though).

District heating balance in master, default settings:
image

and feature, default settings:
image

Energy balance in master:
image

and feature:
image

### CODE REVIEW

**Comment by amos-schledorn** [comment]
Note: [36110d5](https://github.com/PyPSA/pypsa-eur/pull/1893/commits/36110d5ffb8abc3ead389348d3e8fd157eb64811) fixed a bug that caused the return flow temperature to be used in the COP approximation of all heat sources, including those pre-heating the return flow.
-> The above results assume incorrect COPs for heat sources above the return temperature boosted via heat pumps.

File: scripts/add_existing_baseyear.py [hunk 1]
```diff
@@ -580,40 +581,55 @@ def add_heating_capacities_installed_before_baseyear(
...
+                        bus1=nodes_elec,
+                        carrier=f"{heat_system} {heat_source} heat pump",
+                        efficiency=1 / efficiency.clip(lower=0.001),
+                        capital_cost=costs.at[costs_name, "capital_cost"],
+                        p_nom=existing_capacities.loc[
+                            nodes, (heat_system.value, f"{heat_source} heat pump")
+                        ]
+                        * ratio,
+                        p_max_pu=0,
+                        p_min_pu=-1 * efficiency / efficiency.clip(lower=0.001),
```
lindnemi, line 669: this line could be simplified to `[-1 if eff > 0 else 0 for eff in efficiency]`

doc/configtables/sector.csv
```diff
@@ -20,12 +20,14 @@ district_heating,--,,
...
--- -- -- enable,--,"{true, false}",Add option to enable supplemental heating of pit storage in district heating
--- -- -- booster_heat_pump: true,--,"{true, false}",Add option to enable a booster heat pump for supplemental heating of pit storage in district heating
--- -- max_top_temperature,C,float,The maximum top temperature of the pit storage according to DEA technology catalogue (2018)
--- -- min_bottom_temperature,C,float,The minimum bottom temperature of the pit storage according to DEA technology catalogue (2018)
+-- -- enable,--,"{true, false}",Enable investments in pit thermal energy storage (PTES) in district heating
+-- -- temperature_dependent_capacity,--,"{true, false}",Enable dynamic temperature-dependent energy capacity of PTES based on actual top/bottom temperatures
+-- -- charge_boosting_required,--,"{true, false}",Require boosting when charging PTES (placeholder for future feature)
+-- -- discharge_resistive_boosting,--,"{true, false}",Use resistive heaters instead of heat pumps for boosting PTES discharge when needed
+-- -- top_temperature,C,float,Operating top temperature of the PTES
+-- -- bottom_temperature,C,float,Operating bottom temperature of the PTES
```
cpschau: PR decription indicates that `forward` and `return` are also possible values, so not strictly a float

scripts/add_brownfield.py
```diff
@@ -358,7 +358,7 @@ def update_dynamic_ptes_capacity(
 
     update_heat_pump_efficiency(n, n_p, year)
 
-    if snakemake.params.tes and snakemake.params.dynamic_ptes_capacity:
+    if snakemake.params.ttes and snakemake.params.dynamic_ptes_capacity:
```
cpschau: Shouldn't this be `snakemake.params.ptes`

scripts/build_cop_profiles/run.py
```diff
@@ -106,6 +135,138 @@ def get_cop(
...
+    source_temperature: float | xr.DataArray,
+    central_heating_return_temperature: xr.DataArray,
+) -> float | xr.DataArray:
+    """
+    Determine the effective source inlet temperature for the heat pump.
+
+    For heat sources with preheating capability (e.g., PTES), when the source
+    temperature exceeds the return temperature, a preheater can be used to
+    preheat the return flow. In this case, the heat pump sees the return
+    temperature as its effective source inlet (since it lifts from there).
```
cpschau, line 185: Maybe add assumption of perfect heat exchangers

scripts/definitions/heat_source.py
```diff
@@ -0,0 +1,425 @@
...
+        Check if the heat source uses preheating when below forward temperature.
+
+        Preheating allows intermediate-temperature sources to warm the return
+        flow before a heat pump provides the final temperature lift, improving
+        overall system efficiency.
+
+        Returns
+        -------
+        bool
+            True for PTES, False otherwise.
```
cpschau: Not only true for PTES

scripts/definitions/heat_source.py
```diff
@@ -0,0 +1,425 @@
...
+        """
+        if self in [HeatSource.PTES, HeatSource.GEOTHERMAL]:
+            return True
+        else:
+            return False
```
cpschau: Couldn't we just set this attribute based on source and return temperature? Would make addition of new heat sources easier.

doc/configtables/sector.csv
```diff
@@ -20,12 +20,14 @@ district_heating,--,,
 -- -- rolling_window_ambient_temperature, h, int, Rolling window size for averaging ambient temperature when approximating supply temperature
 -- -- relative_annual_temperature_reduction,, float, Relative annual reduction of district heating forward and return temperature - defaults to 0.01 (1%)
 -- ptes,,,
--- -- dynamic_capacity,--,"{true, false}",Add option for dynamic temperature-dependent energy capacity of pit storage in district heating
--- -- supplemental_heating,,,
--- -- -- enable,--,"{true, false}",Add option to enable supplemental heating of pit storage in district heating
--- -- -- booster_heat_pump: true,--,"{true, false}",Add option to enable a booster heat pump for supplemental heating of pit storage in district heating
--- -- max_top_temperature,C,float,The maximum top temperature of the pit storage according to DEA technology catalogue (2018)
--- -- min_bottom_temperature,C,float,The minimum bottom temperature of the pit storage according to DEA technology catalogue (2018)
+-- -- enable,--,"{true, false}",Enable investments in pit thermal energy storage (PTES) in district heating
```
cpschau: Table entry for tes must be changed to ttes. Also, it feels a bit unpractical that there are so many rows between `district_heating` and  `ttes` & other heating technologies, although they can be also associated to district heating systems (here and in the config itself).

config/config.default.yaml
```diff
@@ -687,7 +683,7 @@ sector:
     annualise_cost: true
     tax_weighting: false
     construction_index: true
-  tes: true
+  ttes: true
```
cpschau, line 815: Feels weird that `ttes` and other heating keys follow so many rows further down.

cpschau (changes requested):
Great work @amos-schledorn! I think PR description and the code complemented each other very well, which made it almost a joy to review. There are only very few and minor things to change & discuss before this is ready to merge!

doc/configtables/sector.csv
amos-schledorn: I agree. I didn't change that because `ttes` is also enabled for individual heating, so the setting shouldn't be under `district_heating`. I'd suggest tackling this in a separate feature on unifying config settings across TES technologies.

scripts/definitions/heat_source.py
amos-schledorn: I considered that but couldn't come up with a good solution for time-varying temperatures (sometimes above, sometimes below return temperature) and for users experimenting with different temperatures (e.g. excess temperatures around the return temperature). The structure should, however, handle unnecessary pre-heaters by setting efficiencies their to zero. 
(In fact, pre-heaters are always build for all heat sources but might not be used - I'm assuming solvers will handle that in pre-solve).
Happy to hear suggestions!

config/config.default.yaml
amos-schledorn: Same point as above.

scripts/definitions/heat_source.py
cpschau: Just to document our oral exchange: Minor change could enable dynamic allocation of source temperature to source (source <= T_ret) or sink inlet (source > T_ret) and make `requires_preheater` attribute obsolete.

doc/configtables/sector.csv
cpschau: Agree about tackling reordering in a separate PR, but lets change the config table entry from `tes`to `ttes`, to align with the renaming of the config entry.

cpschau (changes requested):
Minor things remaining:

- config table key should mirror config keys
- dynamic COP source/sink inlet adjustments for boostable sources

scripts/definitions/heat_source.py
amos-schledorn: Done!

amos-schledorn:
Thanks for the review, @cpschau. I have
- renamed `return_temperature_bus` and `medium_temperature_bus` to `heat_pump_input_bus` and `preheater_input_bus`
- removed the manual `requires_preheater` attribute: the code now handles this dynamically by routing all exhaustible heat sources 
- tested the updated workflow (not documenting results again) by comparing model dispatch to manually computed boosting and preheating requirements

scripts/build_cop_profiles/run.py
```diff
@@ -106,6 +135,135 @@ def get_cop(
...
+        except KeyError:
+            raise ValueError(
+                f"Constant temperature for heat source {heat_source_name} not specified in parameters."
+            )
+
+    else:
+        if f"temp_{heat_source_name}" not in snakemake_input.keys():
+            raise ValueError(
+                f"Missing input temperature for heat source {heat_source_name}."
+            )
```
cpschau: I am not sure, but wouldn't snakemake already raise an error in `input_heat_sources` (https://github.com/PyPSA/pypsa-eur/blob/50261f7d196ae480bb20164cc8f5727d0f0be7b9/rules/build_sector.smk#L574-L576)?

scripts/prepare_sector_network.py
```diff
@@ -3115,98 +3129,91 @@ def add_heat(
...
-                        .reindex(index=n.snapshots)
-                    )
-                else:
-                    e_max_pu = 1
+            n.add(
+                "Link",
+                nodes,
+                suffix=f" {heat_system} water pits discharger",
+                bus0=nodes + f" {heat_system} water pits",
+                bus1=HeatSource.PTES.resource_bus(nodes, heat_system),
```
cpschau, line 3148: I assume this is the "urban central ptes heat" bus, that is added just before in L 3166? Some  comments would be helpful to understand what is going on. A compromise has to be found between keeping `prepare_sector_network` lean but also understandable.

scripts/prepare_sector_network.py
```diff
@@ -3115,98 +3129,91 @@ def add_heat(
...
+                bus=nodes + f" {heat_system} water pits",
+                e_cyclic=True,
+                e_nom_extendable=True,
+                e_max_pu=e_max_pu,
+                carrier=f"{heat_system} water pits",
+                standing_loss=costs.at["central water pit storage", "standing_losses"]
+                / 100,  # convert %/hour into unit/hour
+                capital_cost=costs.at["central water pit storage", "capital_cost"],
+                lifetime=costs.at["central water pit storage", "lifetime"],
+                e_nom_min=100000,
```
cpschau: PTES fandom must not be lived by manipulating the upstream model parameters 😆

scripts/build_geothermal_heat_potential.py
```diff
@@ -2,44 +2,76 @@
 #
 # SPDX-License-Identifier: MIT
 """
-Build heat source potentials for a given heat source.
+Build geothermal heat source potentials for district heating networks.
 
-This script maps and aggregates geothermal heat source potentials `onshore_regions`. Input data is provided on LAU-level and is aggregated to the onshore regions.
-It scales the heat source utilisation potentials to technical potentials by dividing the utilisation potentials by the full load hours of the heat source, also taking into account the energy unit set for the respective source in the config.
+This script maps and aggregates geothermal heat source potentials from LAU-level
```
cpschau, line 7: I think the justification of potential scaling could be explained a little bit better. It is not always clear which temperature difference you refer to (forward, return, heat source, cooling).

scripts/build_geothermal_heat_potential.py
```diff
@@ -58,6 +90,99 @@
...
+    xr.DataArray
+        Scaled heat source power [MW]. Dims: (time, name).
+    """
+    # Ensure alignment of regions
+    regions = heat_source_power.index
+    forward_temp = forward_temperature.sel(name=regions)
+    return_temp = return_temperature.sel(name=regions)
+
+    # Broadcast return_temperature to match forward_temperature dimensions
+    return_temp_broadcast = return_temp.broadcast_like(forward_temp)
```
cpschau, line 144: Do forward and return T ever arrive with different temporal indices?

scripts/build_cop_profiles/run.py
```diff
@@ -106,6 +135,135 @@ def get_cop(
...
+    """
+    Determine the effective sink inlet temperature for the heat pump.
+
+    For heat sources with preheating capability (e.g., PTES), when the source
+    temperature exceeds the return temperature, a preheater raises the return
+    flow to forward temperature. The heat pump then lifts from return to forward
+    temperature. When preheating is not used (source <= return), the heat pump
+    receives water at return temperature and heats it to forward temperature.
+
+    When source temperature > return temperature, preheater is used: preheater raises return flow, heat pump inlet is at forward temp.
```
cpschau: This is wrong. The sink inlet temperature should never be at forward T level but either at source T or return T level. The sink outlet T should be equal to the forward level after liftinf the temeprature.

Instead it should be:
```
 return xr.where(
        source_temperature > central_heating_return_temperature,
        source_temperature,
        central_heating_return_temperature,
    )
```

cpschau (changes requested):
Logic for sink inlet T must be changed, and e_nom_min for PTES must be removed.

scripts/build_cop_profiles/run.py
amos-schledorn: Good catch! I think it's actually supposed to be
```
return xr.where(
        source_temperature > central_heating_return_temperature,
        central_heating_return_temperature,
        source_temperature,
)
```
(which is what I implemented)

scripts/prepare_sector_network.py
amos-schledorn: But isn't this way more readable than a string? You can just hit f12/hover to get a full description or highlight where else it's being used.

cpschau: Sure, but maybe a comment above would be helpful to understand that the bus components initialized are these exact resource buses.

scripts/build_cop_profiles/run.py
cpschau: I think it is `xr.where(condition, value if true, value if false)`, so it should be flipped like I suggested before, if we are talking about the sink side.
Screenshot from 2026-02-05 09-35-13

scripts/build_geothermal_heat_potential.py
amos-schledorn: `return_temp` should be constant and not time-indexed, thus the broadcast.

scripts/prepare_sector_network.py
amos-schledorn: Gotcha. Didn't see that the bus creation uses an explicit string. It now does and I've also added examples to the `HeatSource.resource_bus` method.

scripts/build_cop_profiles/run.py
amos-schledorn: You're right of course, brain-farted a little - fixed.

fneum:
1. Is there any way to split this PR into multiple chunks? This would massively simplify the review. I started to review, but I found it hard to structure my approach. There seem to be some elements that are orthogonal and some that build up on each other.

2. Are the comparisons at the top of the PR up to date? Last edit seems to be from December.

3. If I understand correctly, there are changes to the topology of buses/links. How many more buses and links are added (as a function of regions) and what was the computational impact?

amos-schledorn:
> 1. Is there any way to split this PR into multiple chunks? This would massively simplify the review. I started to review, but I found it hard to structure my approach. There seem to be some elements that are orthogonal and some that build up on each other.
> 
>     2. Are the comparisons at the top of the PR up to date? Last edit seems to be from December.
> 
>     3. If I understand correctly, there are changes to the topology of buses/links. How many more buses and links are added (as a function of regions) and what was the computational impact?

Thanks for looking into that - I know this is a lot of code!
1. Not really - with the exception of adjusting geothermal heat source cooling. Lmk if you'd like this moved.
2. Updated the description - sorry! Numerical results have not been updated.
3. Fair point. I didn't record the computational impact. Per exhaustible heat source (by default PTES and geothermal, in #1944 also PtX excess heat), 2 additional busses and links should be created per region. For heat-sources with temperatures always above the forward temperature, all efficiencies of all links but direct utilisation are zero, which I'd expect strong solvers like Gurobi to handle well in presolve. For heat sources always below return temperature, preheating and direct utilisation efficiencies are zero (i.e. river-water heat, not enabled by default).
TLDR: 4 busses + 4 links per region.

Review giude:
- start with PR intro, pull the branch, review config changes
- review `prepare_sector_network`, inspect relevant `heat_source` attributes and methods assuming preheater, utilisation, COP profiles are correct
- review those profiles, i.e. changes in `build_heat_source_utilisation`, `build_geothermal_heat_potential`, `build_cop_profiles`
- the rest are minor supporting changes

PR #1901, opened 2025-11-14
Fix clean-tests section in pixi.toml

euronion:
Follow up to #1886 . `clean-tests` task was not working because of surplus `||`. And I think they are meant to be `&&` instead.

## Checklist

### DIFF
--- a/pixi.toml
+++ b/pixi.toml
@@ -135,12 +135,12 @@ integration-tests = """
     echo "All tests completed successfully."
 """
 clean-tests = """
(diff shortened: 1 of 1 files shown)


lkstrp (commented):
I think it should rather be `;` instead of `||` to always run all clean up commands, no matter if all succeed

euronion:
> I think it should rather be `;` instead of `||` to always run all clean up commands, no matter if all succeed

Works for me as well, as long as we fix the surplus `||`

coroa:
Agree for the `;`. `||` means: only run the next step, when the previous one failed.

brynpickering:
My bad, should have used `;` from the start...

PR #1838, opened 2025-09-22
Streamline workflow

FabianHofmann:
Closes #1916. (supersedes)

# Streamlined and restructured workflow

This PR refactors the workflow in order to streamline and unify optimization approaches and configuration settings. here are some highlights of the PR, otherwise no one will have the motivation to look at this large change. I suggest before going into the code, read this here first, even though it is long...

## Highlights
- No more wildcards excepts for {horizon}; all moved explicit to config entries
- No more cryptic filenames like elec_s_37_lv1.25_3H_2030.nc
- Processing of network follows the logic of `base.nc` -> `clustered.nc` -> `composed_{horizon}.nc` -> `solved_{horizon}.nc`
- Results are unified and made independent of horizons (`csvs/nodal_costs.csv`,  `csvs/capacities.csv` ...) 
- `compose` rule is the entry point for green-field, brown-field, perfect foresight model for both sector-coupled and electricity-only.
- Electricity-only models now run with myopic an perfect foresight 
- Make and plot summary now works for **all combinations of models (elec/sector) and foresights**

and now a more descriptive summary

## Overview
- The pr restructures the Snakemake workflow of PyPSA-Eur into a leaner pipeline (`base → simplified → clustered → composed → solved`) and standardises filenames so scenario information now lives in configuration instead of wildcards.
- All rules that previously emitted `*_s_{clusters}` or `..._{planning_horizons}` artefacts have been renamed; the new naming scheme makes horizons explicit via `{horizon}` and removes cluster suffixes.
- Post-processing rules were refactored to read the new solved network filenames and now operate uniformly across all foresight modes, so the map and summary targets remain identical for overnight, myopic, and perfect runs for both elec-only and sector-coupled.

## Workflow Changes
1. **Stage progression** – The workflow now moves strictly through `networks/base.nc` → `networks/simplified.nc` → `networks/clustered.nc` → `networks/composed_{horizon}.nc` → `RESULTS/networks/solved_{horizon}.nc`. Old intermediate targets such as `networks/base_s_{clusters}_{opts}_{sector_opts}_{planning_horizons}.nc` no longer exist.
2. **Unified composition** – `rules/compose.smk` encapsulates what used to be multiple `prepare_*`, `add_*`, and brownfield rules; it assembles everything needed for a given `horizon` and handles the myopic/perfect brownfield inputs.
[...]

PR #1902, opened 2025-11-14
Make model instance conveniently inspectable (separate problem prep and solving)

Irieo:

This PR refactors solve_network.py (and solve_operations_network.py) with the aim to make an instance of a optimization problem to be solved in the workflow easily inspectable.

### Why this PR

The reason is that `solve_network` function included all steps from model creation, unpacking extra_functionality via kwargs, and solving models based on if-conditions for different optimization models:

https://github.com/PyPSA/pypsa-eur/blob/822a92729e6973aa3aff741d6c94f1da2c75e8b2/scripts/solve_network.py#L1276-L1398

It was therefore inconvenient to investigate model instance (_are all required constraints attached? what is the RHS of a specific constraint? etc_) I spoke recently with @fneum @bobbyxng @cpschau who all confirmed it is pretty annoying and we should have refactored it long ago. I also ping @lindnemi with whom we discussed this problem long ago.

Similar impression I got from a series of PyPSA(-Eur) workshops we did this year, where folks from industry told me it is not straightforward to investigate optimization problem at hand with PyPSA-Eur.

And finally, I'm on making PyPSA-Eur workflow compatible with https://github.com/PyPSA/PyPSA/pull/1345 and https://github.com/PyPSA/PyPSA/pull/1154 and this PR is a predecessor.

### Changes

- I create a dedicated `prepare_optimization_problem()` function
- and a dedicated `collect_kwargs()` that separates model creation args and solver args
- and refactor main block to explicit steps incl. prepare_network, prepare_optimization_problem() and solve problem depending on optimization mode
- I keep all model- and solve-related functionality/options from original code

This makes code modular. One can mock_snakemake into a step after model prep and inspect optimization problem 
instance before solving it. I also move rolling horizon logic fully to a correct script (solve_operations_network.py).

### Tests

I tested the workflow for three modes

- [x] snakemake -call results/test-sector-overnight/networks/base_s_5_elec_.nc --configfile config/test/config.overnight.yaml
- [x] snakemake -call results/test-rolling-horizon/networks/base_s_5_elec__op.nc --configfile config/examples/config.rolling_horizon.yaml
- [x] snakemake -call results/test-iterative/networks/base_s_5_elec_.nc --configfile config/examples/config.iterative.yaml

I add `config.iterative.yaml` and `config.rolling_horizon.yaml` into collection of example configs within PR as well. 

### Other

I am aware of major refactoring effort in the draft PR https://github.com/PyPSA/pypsa-eur/pull/1838 and I believe these are complementary. I would prefer we merge this now so I can checkout to work on stochastic-compatible workflow; and this PR can be merged into #1838 during rebase to master.

### Checklist


FabianHofmann:
Hey @Irieo, this makes sense! happy to review (hope this week). I am just wondering about the bigger picture about how operational optimization could be generalized in to `solve_network.py` and if we shouldn't keep that flexibility. with outsourcing rolling horizon the workflow would not allows us to directly run the rh for a brownfield network?

Irieo:
> Hey @Irieo, this makes sense! happy to review (hope this week). I am just wondering about the bigger picture about how operational optimization could be generalized in to `solve_network.py` and if we shouldn't keep that flexibility. with outsourcing rolling horizon the workflow would not allows us to directly run the rh for a brownfield network?

I agree, operational optimization could be generalised within solve_network (while keeping advantages of this PR for making folks able to inspect model instance before RH run). I refactor here solve_operations_network.py mostly to stick to current structure w/o touching bigger picture. Can volunteer to merge this PR later to #1838 when it will be rebased; as well as merge RH logic into solve_network.py.

upd: ah @FabianHofmann thanks I got what you meant. Indeed, with the implementation we had here, brownfield + RH did run through w/o break but it was single-pass optimization. Now I enabled it. So `config/examples/config.myopic-rolling.yaml`  this would run brownfield fleet and RH as intended.

scripts/solve_network.py
```diff
@@ -1273,129 +1273,125 @@ def check_objective_value(n: pypsa.Network, solving: dict) -> None:
...
-        kwargs["min_iterations"] = cf_solving["min_iterations"]
-        kwargs["max_iterations"] = cf_solving["max_iterations"]
-        if cf_solving["post_discretization"].pop("enable"):
-            logger.info("Add post-discretization parameters.")
-            kwargs.update(cf_solving["post_discretization"])
-        status, condition = n.optimize.optimize_transmission_expansion_iteratively(
-            **kwargs
-        )
+    # Prepare solver kwargs
+    model_kwargs, solve_kwargs = collect_kwargs(config, solving, planning_horizons)
```
fneum: It would be even better for modularity and debugging if this function were not called inside `prepare_optimization_problem` but called outside with `model_kwargs` and `solve_kwargs` passed as arguments.

```py
 model_kwargs, solve_kwargs = collect_kwargs(config, solving, planning_horizons)
 prepare_optimization_problems(..., model_kwargs, solve_kwargs)
 ```

File: scripts/solve_network.py [hunk 2]
```diff
@@ -1273,129 +1273,125 @@ def check_objective_value(n: pypsa.Network, solving: dict) -> None:
...
 
-    if not rolling_horizon:
-        if status != "ok":
-            logger.warning(
-                f"Solving status '{status}' with termination condition '{condition}'"
-            )
-        check_objective_value(n, solving)
+    # Apply any overrides
+    model_kwargs.update(model_kwargs_override)
```
fneum: Then, can move this also to `collect_args` function, so their output is a final list of settings.

scripts/solve_network.py
```diff
@@ -1273,129 +1273,125 @@ def check_objective_value(n: pypsa.Network, solving: dict) -> None:
...
 
-    rolling_horizon = cf_solving.pop("rolling_horizon", False)
-    skip_iterations = cf_solving.pop("skip_iterations", False)
-    if not n.lines.s_nom_extendable.any():
-        skip_iterations = True
-        logger.info("No expandable lines found. Skipping iterative solving.")
+    return model_kwargs, solve_kwargs
+
+
+def prepare_optimization_problem(
```
fneum: To avoid ambiguity with `prepare_network()` and follow PyPSA naming more closely, I would call this `create_model()` or `create_optimization_model()`.

scripts/solve_network.py
```diff
@@ -1430,24 +1429,97 @@ def solve_network(
...
+            )
+            all_kwargs = {**model_kwargs, **solve_kwargs}
+            all_kwargs["track_iterations"] = cf_solving["track_iterations"]
+            all_kwargs["min_iterations"] = cf_solving["min_iterations"]
+            all_kwargs["max_iterations"] = cf_solving["max_iterations"]
+            all_kwargs["log_fn"] = snakemake.log.solver
+
+            if cf_solving["post_discretization"].get("enable", False):
+                logger.info("Add post-discretization parameters.")
+                all_kwargs.update(cf_solving["post_discretization"])
```
fneum: Functionality-wise it looks fine, but if you already have a dedicated function that identifies the kwargs, wouldn't it make sense to also include this logic in the function? Same for rolling horizon above.

Ideally you don't need to touch the kwargs after `collect_kwargs`.

scripts/solve_network.py
```diff
@@ -1430,24 +1429,97 @@ def solve_network(
...
-        solve_network(
-            n,
-            config=snakemake.config,
-            params=snakemake.params,
-            solving=snakemake.params.solving,
-            planning_horizons=planning_horizons,
-            rule_name=snakemake.rule,
-            log_fn=snakemake.log.solver,
-        )
+        if rolling_horizon:
```
fneum: Previously there was a test that the rule name was `solve_operations_networks`. If you want to open this for `solve_network` rule (this would be new functionality), there should be a check that no capacities are extendable.

scripts/solve_operations_network.py
fneum, line 1: Fine for now, but in the long-term we should get rid of this script and integrate into `solve_network.py`. The only specific that happens here is `n.optimize.fix_optimal_capacities()`.

scripts/solve_network.py
Irieo: Alright. `create_optimization_model()` since that PyPSA has own `create_model()`

scripts/solve_network.py
Irieo: I agree, this also follows the main idea of the PR. done.

scripts/solve_network.py
Irieo: I suppose this could even be removed, since `collect_args` has the final list of settings already.

scripts/solve_network.py
Irieo: Yeah, this boils down to where do we want handle kwargs of special optimization modes, iterative and RH: in the caller within if-clause specifying the mode or within `collect_kwargs`. Since we agreed we want `collect_kwargs` to list the final optimization settings, it does make sense to include this logic there.

scripts/solve_network.py
Irieo: Hmm, I'd rather bring back the rule name if-clause. If we check that no capacities are extendable in solve_network, than we can get rid of `solve_operations_network()`, which we eventually do; however that should be done within a dedicated PR.

Above is related Q/A https://github.com/PyPSA/pypsa-eur/pull/1902#issuecomment-3541369420

scripts/solve_operations_network.py
Irieo: Exactly.

Irieo:
> Looks good and conceptually I agree. There is some small optimization potential in dealing with the kwargs.
> 
> I am against dumping example configs (potentially high-maintenance and not tested to ensure they are up to date, lots of boilerplate settings for test case).

Thanks for reviewing it. I agree with the suggestion on kwargs and implemented it now.

WRT example configs: I agree that dumping configs that require maintenance isn't a good practice. However, we currently have no example config (let alone anything under test/) for rolling horizon. The model offers rolling-horizon optimization for both solve_operational_network (dispatch only) and solve_network (inv -> freeze inv -> pass to operational), and none of this is tested. Right now, we would learn about things breaking with RH through a GitHub issue, which isn't great. More, anyone trying to test this functionality (like I am doing here) has to inspect the code to assemble a config, and may run into untested non-working parts. I initially added RH to the test suite, then moved to the examples after some feedback. I'm removing the configs now since they are unrelated to this PR, but I suggest adding a copy of RH config with settings matching the Belgian test case to the test suite. I'm happy to volunteer implementing a minimalistic good enough test.

I have tested this final version of PR with 
- overnight foresight (config.overnight.yaml)
- myopic foresight (config.myopic.yaml)
- iterative transmission expansion (config.iterative.yaml)
- rolling foresight for dispatch only (new config now removed)
- browfield with rolling horizon (related to [this comment above](https://github.com/PyPSA/pypsa-eur/pull/1902#issuecomment-3541369420), new config now removed)
- perfect foresight is out of the list https://github.com/PyPSA/pypsa-eur/blob/efb69ee0be3c4594a6d2e305333d50ede21604ad/scripts/prepare_perfect_foresight.py#L38-L40

Irieo:
🟢

PR #1905, opened 2025-11-17
ci: fix unit tests typo

lkstrp:
Small follow up on #1886 

And a quick cry for help to get our CI back, @euronion

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -64,7 +64,7 @@ jobs:
 
     - name: Run unit tests
       run: |
-        pixi run unit-test
+        pixi run unit-tests
 
     - name: Upload artifacts
       if: always()


coroa:
We could also priority board the integration of https://github.com/PyPSA/snakemake-storage-plugin-cached-http before #1675 . Want me to propose a PR for that?

lkstrp:
If you think it won't take too much time, let's do it. The data layer still needs some final care

coroa:
> If you think it won't take too much time, let's do it. The data layer still needs some final care

Integration is trivial, but with pixi i think we need now all on conda-forge, don't we? I'll have to get the snakemake-storage-plugin then onto conda-forge. Will tackle tomorrow, put it on pypi now.

euronion:
> > If you think it won't take too much time, let's do it. The data layer still needs some final care
> 
> Integration is trivial, but with pixi i think we need now all on conda-forge, don't we? I'll have to get the snakemake-storage-plugin then onto conda-forge. Will tackle tomorrow, put it on pypi now.

No, you don't. You can install from Github or local if you want. I've already merged the `pixi` changes into #1675 , you can have a look there

lkstrp:
> Integration is trivial, but with pixi i think we need now all on conda-forge, don't we? I'll have to get the snakemake-storage-plugin then onto conda-forge. Will tackle tomorrow, put it on pypi now.

I think @coroa is right. pypi/ git to `pixi` works fine, but the `pixi` export to `conda` explicit files will make problems. Ref to https://github.com/PyPSA/pypsa-eur/pull/1886#issuecomment-3501610583. Which means the cleanest solution is conda-forge

coroa:
Ok, i'm doing that now. First up, the dependency `tqdm-loggable`: https://github.com/conda-forge/staged-recipes/pull/31513 . I have also prepared everything for putting `snakemake-storage-plugin-cached-http` onto bioconda. I just need `tqdm-loggable` to land first.

coroa:
Ok, one down. Now the bioconda one: https://github.com/bioconda/bioconda-recipes/pull/60746

PR #1911, opened 2025-11-19
fill cost calues with default after merging custom costs

martacki:
Currently cost values are filled with default values before merging them with the (raw) custom cost data. If the custom cost data however includes new technologies but misses, for example, the discount rates, the prepared results such as `capital_cost` remain missing / `NaN`.

I suggest changing the order.

It's not necessary for the `custom_prepared` step, because this is either a result or given (so it's filled either way).

## Checklist

- [x] Changed dependencies are added to `pixi.toml` (using `pixi add <dependency-name>`).

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -32,6 +32,17 @@ jobs:
         shell: bash -l {0}
 
     steps:
+    - name: Free up disk space
+      run: |
+        echo "Initial disk space"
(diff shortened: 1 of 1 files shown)


martacki:
enabling auto merge @fneum @lkstrp because this is such a mini change, wouldn't do that otherwise.

Hope it's all right - thanks!

PR #1912, opened 2025-11-19
Config Validation with Pydantic

### LINKED ISSUE #1721: Pin documentation package requirements
It's generally good practice to pin the versions of packages used for documentation. This avoids the documentation site breaking as new pacakge versions are released.

https://github.com/PyPSA/pypsa-eur/blob/478b7abcd846b6483f8379773676c776791e85bf/doc/requirements.txt#L5-L10

lkstrp:
Supersedes #1547

Closes #1547
Closes #1734
Closes #1721

This PR revives and supersedes #1547. This time using pydantic which can export to JSON Schema instead of writing the JSON Schema directly. 

With the new approach we can do a couple of new things and keep anything config related at one place:
1. Validation 
    - Run any validation: Just simple type checks (`bool`, `list[str]`), Literals (e.g. for logging levels `Literal["DEBUG", "INFO", "WARNING", "ERROR", "CRITICAL"]`, as well as any custom validation you can come up with. This PR tries to not break anything, which means only types, and `Literals` from the previous doctable `description`/`values` column have been taken over. In future we wanna get as strict as possible.
2. Defaults
    - Defaults are now defined in the model and exported to `config.default.yaml` and `schema.json`. The config can be used locally as is, but must be kept in sync upstream with the validation config. Since the file is auto generated now, we also lose all comments.
3. Typed Config Object
    - We can now use the pydantic schema object instead of the previous dict config. This gives us a typed object. Which means that we can have auto completion and documentation in the IDE (not just in the config.yaml), type checks via mypy etc and attribute style config selection. Both dict-like and attribute object access is then possible: `snakemake.config["run"]` and `snakemake.config.run`.
        - This is currently not implemented. We need to get rid of any use of `dash-case` in the config and move to `snake_case`. In the export, currently both is supported via aliases, but python attributes do not allow for `dash-case`. We could also duck type this, but it's cleaner to just remove them.
4. Deprecation of config keys
    - Deprecate keys and warn user via [Deprecated Fields](https://docs.pydantic.dev/latest/concepts/fields/#deprecated-fields)
5. Docs
    - The schema is also used in the documentation, which contains additional information such as types, defaults, examples and descriptions. All doctables have been removed. Check new docs build [here](https://pypsa-eur--1912.org.readthedocs.build/en/1912/configuration.html). 
    - When the schema is uploaded to `schemastore.org` (what we can do after this PR is merged), most IDEs will automatically give you inline tooltips on config files. It is also possible to set it up manually.
image

**Other changes**
- Splits up workflow in seperate jobs for unit/ integration tests
- Removes old `requirements.txt` for docs, pins docs dependencies in pixi
- Fixes nested string pre-commit issues

## Checklist
- [ ] Add plotting
- [x] Docs

Post Merge
- [ ] Apply for schemastore with  `**/pypsa-eur/config/*.yaml` and `**/pypsa-eur*/config/*.yaml`
- [ ] Move from `dash-case` to `camel_case` for all config keys and add typed object
- [ ] Stricter validation

### DIFF
--- a/doc/release_notes.rst
+++ b/doc/release_notes.rst
@@ -15,6 +15,8 @@ Upcoming Release
   exported from the schema. Changes to the default config, therefore now require the
   schema to be updated. Find a detailed explanation in the contributors documentation
   (https://github.com/PyPSA/pypsa-eur/pull/1912).
+  
+* Fix bugs when using PyPSA-Eur as a Snakemake module by making sure that all file paths are defined relative to a rule's input or an output (https://github.com/PyPSA/pypsa-eur/pull/1967).
 
 * Fix compatibility of rules `build_gas_input_locations` and `build_gas_network` with pyogrio >=0.12.0 (https://github.com/PyPSA/pypsa-eur/pull/1955).
 
--- a/doc/data_sources.rst
+++ b/doc/data_sources.rst
@@ -92,9 +92,9 @@ It is also possible to use the helper script in ``scripts/create_zenodo_depositi
     * ``doc/configtables/data.csv`` for the documentation
     * ``data_sources.rst`` data inventory for PyPSA-Eur
 
-==============
+##############
 Data inventory
-==============
+##############
 
 .. csv-table::
    :header-rows: 1
--- a/rules/build_electricity.smk
+++ b/rules/build_electricity.smk
@@ -268,7 +268,7 @@ rule determine_availability_matrix:
         renewable=config_provider("renewable"),
     input:
         unpack(input_ua_md_availability_matrix),
-        corine=ancient(f"{rules.retrieve_corine.output['tif_file']}"),
+        corine=ancient(rules.retrieve_corine.output["tif_file"]),
         natura=lambda w: (
             f"{NATURA_DATASET['folder']}/natura.tiff"
             if config_provider("renewable", w.technology, "natura")(w)
--- a/rules/retrieve.smk
+++ b/rules/retrieve.smk
@@ -84,8 +84,7 @@ if (CORINE_DATASET := dataset_version("corine"))["source"] in ["archive"]:
             unpack_archive(input["zip_file"], output_folder)
             copy2(input["zip_file"], output["zip_file"])
             copy2(
-                f"{CORINE_DATASET['folder']}/corine/g250_clc06_V18_5.tif",
-                output["tif_file"],
+                f"{output_folder}/corine/g250_clc06_V18_5.tif", output["tif_file"]
             )
 
 elif (CORINE_DATASET := dataset_version("corine"))["source"] in ["primary"]:
@@ -615,8 +614,10 @@ if (CO2STOP_DATASET := dataset_version("co2stop"))["source"] in [
             traps_table3=f"{CO2STOP_DATASET['folder']}/CO2JRC_OpenFormats/CO2Stop_DataInterrogationSystem/Hydrocarbon_Traps1.csv",
             traps_map=f"{CO2STOP_DATASET['folder']}/CO2JRC_OpenFormats/CO2Stop_Polygons Data/DaughterUnits_March13.kml",
         run:
+            output_folder = Path(output["zip_file"]).parent
+            output_folder.mkdir(parents=True, exist_ok=True)
             copy2(input["zip_file"], output["zip_file"])
-            unpack_archive(output["zip_file"], CO2STOP_DATASET["folder"])
+            unpack_archive(output["zip_file"], output_folder)
 
 
 if (GEM_EUROPE_GAS_TRACKER_DATASET := dataset_version("gem_europe_gas_tracker"))[


lkstrp:
> This is a very minimal example right now, and not integrated into the Snakefile. I want to discuss two things first:
> 1. On 4: We could define the defaults right away in the model, which means it would contain everything related: description, validation, type, examples as well as defaults. That would be the cleanest approach. `config.default.yaml` can still be automatically generated to not break any existing workflows, but if you wanna change defaults you would need to touch the model directly. This is what I vote for and what is done in this minimal version. We could also keep the `config.default.yaml` and only use the model for validation, so keep defaults seperated from docs, validation, types etc. This needs to be decided now. @fneum @coroa @FabianHofmann 
> 2. On 2: In a first version we wanna just do basic validation. But if we wanna define more complex stuff, it is a bit a question where do we define it. The cleanest would be to keep that all in `lib/validation/config` and not in any `scripts`. Similar to the example I have right now. Nothing we need to decide right now and I just wanna hear some opinions

I went with option 1 (defining defaults in Pydantic), which means that `config.default.yaml` can be used locally as is, but must be kept in sync upstream with the validation config. Since this file is auto generated now, we also lose all comments.

.github/workflows/test.yaml
```diff
@@ -18,8 +18,33 @@ concurrency:
...
+
+    defaults:
+      run:
+        shell: bash -l {0}
+
+    steps:
+    - uses: actions/checkout@v6
+
+    - name: Skip - no source changes
+      run: echo "Skipping tests because no source code changes detected"
```
fneum: Doesn't that require some conditional?

config/config.default.yaml
```diff
@@ -79,109 +70,125 @@ countries:
...
   extendable_carriers:
-    Generator: [solar, solar-hsat, onwind, offwind-ac, offwind-dc, offwind-float, OCGT, CCGT]
-    StorageUnit: [] # battery, H2
-    Store: [battery, H2]
-    Link: [] # H2 pipeline
-
-  powerplants_filter: (DateOut >= 2024 or DateOut != DateOut) and not (Country == 'Germany' and Fueltype == 'Nuclear')
+    Generator:
+    - solar
+    - "solar-hsat"
```
fneum, line 132: Are these quotation marks necessary? Looks cleaner without them.

doc/configuration.rst
```diff
@@ -60,10 +60,21 @@ Top-level configuration
    :start-at: version:
    :end-before: # docs
```
fneum, line 65: These tricks to cut configuration file do not work anymore with autogenerated file.

doc/configuration.rst
```diff
@@ -60,10 +60,21 @@ Top-level configuration
...
 
-.. csv-table::
-   :header-rows: 1
-   :widths: 22,7,22,33
-   :file: configtables/toplevel.csv
+.. dropdown:: Details
+
+   .. jsonschema:: ../config/schema.json#/properties/version
+      :lift_description:
+
```
fneum: This is the core information of the configuration page. I would not hide it in a dropdown, rather the other way around: the default config yaml could be in a dropdown.

doc/configuration.rst
```diff
@@ -60,10 +60,21 @@ Top-level configuration
...
+   .. jsonschema:: ../config/schema.json#/properties/tutorial
+      :lift_description:
```
fneum: image

The keys are not displayed, so it's unclear what is being described.

scripts/lib/validation/config/__init__.py
```diff
@@ -0,0 +1,355 @@
...
+
+
+class ConfigSchema(BaseModel):
+    """
+    Combined configuration schema for PyPSA-EUR.
+    """
+
+    # TODO Change to extra='forbid' once schema covers all config options
+    # For soft-forks it is recommended to either extend the schema for full config
+    # coverage or allow extra fields with extra='allow'
```
fneum, line 95: I would tend to keep it as `extra='allow'`. Otherwise anyone changing a small configuration setting for themselves has to find this setting or know how to extend schema.

scripts/lib/validation/config/biomass.py
```diff
@@ -0,0 +1,103 @@
...
+        description="The commodities that are included as municipal solid waste.",
+    )
+
+    model_config = ConfigDict(populate_by_name=True)
+
+
+class BiomassConfig(BaseModel):
+    """Configuration for `biomass` settings."""
+
+    year: Literal[2010, 2020, 2030, 2040, 2050] = Field(
```
fneum: I think this one can be any year between 2010 and 2050.

fneum (changes requested):
Great, this is very rigorous! It will be super helpful to have some checking of configuration files for consistency! The only disadvantage I see is that developing new configuration settings now becomes a bit more complicated and less accessible. But I tend to agree that it is best to keep everything in one place.

- I am glad all the comments disappear! We want all documentation in one place.

- `config.default.yaml` and `schema.json` are autogenerated and should not be touched. This should be noted when writing the release note.

- Did you consider another location for the config validation code? Maybe just `./validation/config`, then `scripts` is for the rules? Not sure, but the current structure looks quite deeply nested in subfolders.

- Thinking ahead, will the display also work in `mkdocs` with a similar plugin?

- Sorry, I may have created more configuration settings by merging #1935.

- Could you perhaps write up a few general principles of working with `pydantic` with links to resources in the documentation?

- Could you maybe summarise in one sentence (just for reference), why this is better than the standard validation in `snakemake`? https://snakemake.readthedocs.io/en/stable/snakefiles/configuration.html#validation

lkstrp:
> The only disadvantage I see is that developing new configuration settings now becomes a bit more complicated and less accessible.

For development, not really. You can still do whatever you want locally in your `config.default.yaml`, and Snakemake will use only this. The worst case is that you need to comment out the validation call. Just before merging, you need to adjust the schema to keep both versions in sync. I will explain this all in the docs.

> Did you consider another location for the config validation code? Maybe just `./validation/config`, then `scripts` is for the rules? Not sure, but the current structure looks quite deeply nested in subfolders.

I thought about it, but ended up with the current path:
- `config` subdir is included since we wanna aim to add also validation on input data for any rule, which would sit next to the `config` validation. This is also the reason why everything sits in `scripts`.
- `lib` I added in the desperate try to plan for more structure in pypsa-eur. `_helpers.py` could go there, tons of boilerplate of rules, splitting up rules etc. 
- Putting config validation in an extra `./validation` project level dir, while still having rule validation in `scripts` feels odd. Also adding it to `./configs` isn't a good idea.
Which leads to `./scripts/lib/validation/config`, but with basically to placeholder dirs which should be further filled in future.j

> Thinking ahead, will the display also work in `mkdocs` with a similar plugin?

Haven't tried it yet, but a way will be found. This will not be a blocker to deprecate `doctables`.
 
> Could you maybe summarise in one sentence (just for reference), why this is better than the standard validation in `snakemake`? https://snakemake.readthedocs.io/en/stable/snakefiles/configuration.html#validation

Using Snakemake for validation doesn't change much. Those are basically two questions: How you write the schema and how you validate it:
- In #1547, I wrote the schema straight away in JSON Schema format and you could use snakemake to validate it. But Snakemake does not check for deprecated arguments, which JSON schemas in general support, and therefore I used already there a Python library to validate the schema.
- In here, the schema is written as a Pydantic model. This makes it much easier to write and maintain than having everything in JSON. It can still be exported to a schema representation, which is used to build docs, or get IDE documentation etc. And we could still use snakemake to validate on that JSON schema representation. But we can also use Pydantic for validation, which is done here, and allows us to have:
  - More complex validations in Pydantic, which can't be exported to the JSON schema representation.
  - Not only validate the Snakemake config (which is a simple Python dictionary), but also cast it to a typed Python object, which allows linters to know the type and gives you autocompletion on possible config keys, etc. See (3) above. This is not implemented here, but would be nice to have in the future.

config/config.default.yaml
lkstrp: Yes, since this uses `dash-case`. We could add custom handlers, but it is better to keep it as a reminder to move to `snake_case` everywhere.

fneum:
Thanks for the responses! That all sounds reasonable.

doc/configuration.rst
lkstrp: I've cleaned it up now, but still kept the detailed schema in a drop down menu. Otherwise, some configurations look just too bloated

scripts/lib/validation/config/__init__.py
lkstrp: To be discussed in future

coroa:
> Since [config.default.yaml] is auto generated now, we also lose all comments.

proposal for using ruamel.yaml to bring back comments (not all top-level sections do have a doc heading, i think), in https://github.com/lkstrp/pypsa-eur/pull/34

coroa:
>  Apply for schemastore with **/pypsa-eur/config/*.yaml and **/pypsa-eur*/config/*.yaml

I think i would prefer to add language-server comments to the yaml files linking relatively to the schema. ie. add to `config.default.yaml`:
```
# yaml-language-server: $schema=./schema.json

version: v2025.07.0
tutorial: false
```

This would be more self-contained and apply the right version automatically. This comment is interpreted as is by zed, in vscode one needs the redhat's [vscode yaml extension](https://github.com/redhat-developer/vscode-yaml).

For using schemastore we would probably have to come up with a versioning scheme?

coroa:
Currently links in the description are markdown formatted and therefore don't parse correctly in the docs. In principle this is temporary, since we are switching to mkdocs eventually. I added a few commits to lkstrp#34 that convert those links to sphinx and then add a `markdownDescription` field to the generated json schema. This renders fine correctly in sphinx and in the zed and vscode tooltips.

doc/configuration.rst
coroa: I agree with @fneum that we should put the generated jsonschema directive first and without the dropdown. I think i would even re-include the defaults there. I think i would only keep it in the dropdown for the data section where it does not add any content.

scripts/lib/validation/config/data.py
```diff
@@ -0,0 +1,217 @@
...
+        description="Hotmaps industrial sites data source configuration.",
+    )
+    enspreso_biomass: _DataSourceConfig = Field(
+        default_factory=lambda: _DataSourceConfig(source="archive", version="latest"),
+        description="Enspreso biomass data source configuration.",
+    )
+    osm: _DataSourceConfig = Field(
+        default_factory=lambda: _DataSourceConfig(source="archive", version="latest"),
+        description="OSM data source configuration.",
+    )
```
coroa, line 45: Why not use the _DataSourceConfig defaults implicitly?

```suggestion
    osm: _DataSourceConfig = Field(
        default_factory=_DataSourceConfig,
        description="OSM data source configuration.",
    )
```

lkstrp:
I merged https://github.com/lkstrp/pypsa-eur/pull/34, thanks @coroa ! The improved schema writer is great

I also changed and added the docs, find them here:
- [Configuration — PyPSA-Eur](https://pypsa-eur--1912.org.readthedocs.build/en/1912/configuration.html)
- [Config Validation — PyPSA-Eur](https://pypsa-eur--1912.org.readthedocs.build/en/1912/validation_dev.html)

The plotting schema I will leave for the future. I don't wanna bloat this more. If you both agree I think we are good to go @fneum  @coroa

coroa:
Thanks, Lukas.

I like the new docs. The only issue i spotted was that the offwind-x section does not have any yaml syntax (probably to do with the actual sections being called offwind-ac and dc).

We also had seen in the open-tyndp project the need for some cross-dependent validation tracked in https://github.com/open-energy-transition/open-tyndp/issues/43 . For example in
[open-tyndp/.../config.default.yaml#L146-149](https://github.com/open-energy-transition/open-tyndp/blob/09abe84ce56344f1d191acafc6416de28ce37880/config/config.default.yaml#L146-L149) 
a carrier must not be in `renewable_carriers` and `tyndp_renewable_carriers` at the same time. Any idea how something like this would be supported?

Happy to approve the PR!

lkstrp:
You would use [`@model_validator`](https://docs.pydantic.dev/latest/concepts/validators/#model-validators) and define it at the lowest common ancestor, of all keys you wanna compare. I added a section on custom validators to the docs as well, referring to the `model_validator` and `field_validator`

lkstrp:
Regarding deprecation of fields:

That's possible and already build in with Pydantic, see [Fields - Pydantic Validation](https://docs.pydantic.dev/latest/concepts/fields/#deprecated-as-a-boolean). These will be exported to `schema.json` as well, and will then show up in the documentation and IDE. However, to get the warnings in the logs, we would need to use the typed object returned by the validation. This is the plan for a future PR anyway (see PR description). A runtime deprecation warning would then be emitted when the key is accessed, which is convenient since you also see where the key is used.

But you don't get those warnings, if we don't use the typed config object (which is currently the case). We could add a custom handler for it as well, but I would not see it as a blocker for this PR to get merged and get things rolling first.

doc/release_notes.rst
```diff
@@ -9,6 +9,13 @@ Release Notes
 Upcoming Release
 ================
 
+* Important: PyPSA-Eur now uses a validation schema for configuration files. The schema 
+  also contains the default values for all known configuration options, which means 
+  `config/config.default.yaml` still exists and can be used, but will be automatically
+  exported from the schema. Changes to the default config, therefore now require the
+  schema to be updated. Find a detailed explanation in the contributors documentation
+  (https://github.com/PyPSA/pypsa-eur/pull/1912).
```
fneum, line 17: Maybe highlight that `config.default.yaml` should *not* be edited in commits.

doc/validation_dev.rst
```diff
@@ -0,0 +1,170 @@
...
+Regenerating Config Files
+-------------------------
+
+Snakemake will only read from the ``config/config.default.yaml``, which needs to be generated
+after making changes to the Pydantic model. To regenerate the default config and JSON
+schema:
+
+.. code-block:: console
+
+   $ pixi run generate-config
```
fneum, line 90: Maybe also give the command without `pixi` (if simple enough).

doc/validation_dev.rst
```diff
@@ -0,0 +1,170 @@
...
+
+       @model_validator
+       def check_file_path_required(self):
+           if self.file_enabled and not self.file_path:
+               raise ValueError("file_path is required when file_enabled is True.")
+           return self
+
+Again, find more information in the Pydantic documentation on
+[Field Validators](https://docs.pydantic.dev/latest/concepts/validators/#field-validators)
+and [Model Validators](https://docs.pydantic.dev/latest/concepts/validators/#model-validators).
```
fneum: markdown syntax in rst

config/config.default.yaml
```diff
@@ -1,15 +1,19 @@
-# SPDX-FileCopyrightText: Contributors to PyPSA-Eur <https://github.com/pypsa/pypsa-eur>
-#
-# SPDX-License-Identifier: CC0-1.0
+%YAML 1.1
+---
+# yaml-language-server: $schema=./schema.json
```
fneum, line 3: Won't REUSE complain if SPDX header disappears?

doc/validation_dev.rst
lkstrp: Not really 

`python -c 'from scripts.lib.validation.config import generate_config_defaults, generate_config_schema; generate_config_defaults(); generate_config_schema()'`

fneum:
Same zenodo problem as in #1958

PR #1916, opened 2025-11-21
feat:  refactors temporal aggregation as a separate rule

### LINKED ISSUE #194: [SUB] Refactor temporal aggregation as a separate rule

The temporal aggregation of time dependent data and snapshots in the model is currently included in `prepare_sector_network`. However, it makes sense to separate this into a separate rule.
Tasks:
- [ ] Refactor `prepare_sector_network` to separate temporal aggregation into its own subsequent rule
- [ ] Check if it is possible to change the input of `time_aggregation` from "networks/base_s_{clusters}_elec_{opts}.nc" to "networks/base_s_{clusters}_{opts}_{sector_opts}_{planning_horizons}.nc". This way, the segmentation option will be able to work again on time dependent data that is only added in `prepare_sector_network`. The issue is documented in https://github.com/open-energy-transition/open-tyndp/issues/343

tgilon:
Closes # (if applicable).

This PR introduces a new distinct rule to apply the time aggregation on the time-varying network data. To achieve this:

- It refactors the temporal aggregation of a sector-coupled model as a separate rule,
- It renames previous `temporal_aggregation` as `build_snapshot_weightings`, and
- It introduces a new rule `temporal_aggregation` after `prepare_sector_network` that effectively applies the aggregation.

The benefit of doing this is that it makes the prepared network more explicit at hourly resolution.

## Checklist

- [x] ~~Changed dependencies are added to `pixi.toml` (using `pixi add <dependency-name>`).~~
- [x] ~~Changes in configuration options are added in `config/config.default.yaml`.~~
- [x] ~~Changes in configuration options are documented in `doc/configtables/*.csv`.~~
- [x] ~~Sources of newly added data are documented in `doc/data_sources.rst`.~~


fneum:
This PR would not be compatible with the approach proposed in #1838. @FabianHofmann presented this today and to me it looks like #1838 would be preferable (without having reviewed the code in detail yet) as it elegantly unifies overnight, myopic, and perfect foresight code.

I would give that PR priority. One could always export the hourly model for reference as additional output. I will keep it open for now but close if #1838 gets merged.

Issue #1917, opened 2025-11-21
Allow incremental transmission expansion in the myopic optimisation


martacki:
Working implementation exists here (only tailored to a slightly different workflow). Savely ignore changes in the `NTC` script(s). Still, might be useful:
https://github.com/open-energy-transition/pypsa-wal/pull/41

## Describe the feature you'd like to see
If set to `vopt`, `copt` or alike, the myopic optimisation can already massively expand the grid during the first `planning_horizon`. As this is often not feasible, it would be better to allow for incremental increases in the `base_year`, and then allow for further increases in the following planning horizons.

Conversely, if initially set to `v1.05` or `c1.05`, the model might hit the upper limit already during the first few iterations and won't have further expansion options.

I am attaching a working implementation, which might need a bit of tailoring for `pypsa-eur` upstream.

I also encountered an issue with this line:
`lines_s_nom = n.lines.s_nom.where(n.lines.type == "", _lines_s_nom)`
where it was setting the `s_nom` to a horrific high value, while a line type was present. I didn't have proper time to debug this, so maybe it was also just an issue on my side.

PR #1918, opened 2025-11-21
increase transmission limit in each myopic run

martacki:
Closes # (if applicable).
#1917 

see the linked issue

## Checklist

- [x] Changed dependencies are added to `pixi.toml` (using `pixi add <dependency-name>`).


config/config.default.yaml
```diff
@@ -154,6 +154,13 @@ electricity:
     by_country: false
 
   transmission_limit: vopt
+  transmission_limit_myopic:
+    2025: v1.05
+    2030: v1.05
+    2035: v1.05
+    2040: v1.05
+    2045: v1.05
+    2050: v1.05
```
fneum: This will be useful! But I think the setting should be integrated in the existing setting:

```yaml
transmission_limit: vopt
```

or

```yaml
transmission_limit:
    2025: v1.05
    2030: v1.1
    2035: v1.15
    2040: v1.2
    2045: v1.25
    2050: v1.3
```

This follows how we do this elsewhere.

martacki: I guess, I was thinking the same but ultimately the change you're suggesting would create somewhat redundant code. The `base_year` automatically uses the `transmission_limit` set in the config; just the default setting of `vopt` is somewhat not optimal. It would be better to have `v1.05` or alike. If TYNDP expansion is active probably `v1.0` would be best.

But this is generally true, if not optimizing in 5 year increments but 10 year increments, I'd also use `v1.1` rather than `v1.05`. (Otherwise the model only has 5% increase limit for 10 years)... I guess there's no perfect way to do this?

Anyway if we want to overwrite it also for the `base_year`, the same fragment is needed in `add_baseyear.py`.

Pls let me know what you think - not only for this, but also maybe an idea how to handle the default setting depending on how the `planning_horizon` increments are set?

martacki: just checked again - main requirement would be either adding a `planning_horizon` to `prepare_network`, or adding another parameter `elec:` to the keys of `transmission_limit`.

Increasing the limit is not necessary as the new limit is now always calculated from `s_nom_opt` + `p_nom_opt` from the previous horizons, so keeping it at `v1.05` is better. We could ofc calculate it always from `p_nom`, not `p_nom_opt`, then your suggestion makes sense (better to resolve the issue from my previous comment, worse in case the previous horizon didn't optimise transmission much , in which case there might be a bit of excessive build-out later which I guess is rather unlikely)

fneum: > but also maybe an idea how to handle the default setting depending on how the planning_horizon increments are set?

I think this should be left to the user. I would not automatically infer this.

> just the default setting of vopt is somewhat not optimal. It would be better to have v1.05 or alike.

I would stick with `vopt` as default because this is the current state. Users can just adjust.

> just checked again - main requirement would be either adding a planning_horizon to prepare_network, or adding another parameter elec: to the keys of transmission_limit.

I think the solution will come with @FabianHofmann's #1838. So this PR should be merged afterwards and this is no concern anymore.

> Increasing the limit is not necessary as the new limit is now always calculated from s_nom_opt + p_nom_opt from the previous horizons, so keeping it at v1.05 is better. 

Ok, agreed. Then the limit is always relative to the volume in the previous horizon, not relative to the capacities in the base year. That is fine. The indexing by planning horizon makes sense nevertheless as build-out could be configured to increase.

```yaml
transmission_limit:
    2025: v1.05
    2030: v1.05
    2035: v1.05
    2040: v1.1
    2045: v1.1
    2050: v1.1
```

Does that sound good?

martacki:
Ok, after merging master, it should be ready.

PR #1919, opened 2025-11-23
hotfix: handle unavailability of WDPA where optional

fneum:
Becomes superfluous with #1675


euronion:
Dataset is back online.

PR #1920, opened 2025-11-24
ci: free up disk space

lkstrp:
See https://github.com/PyPSA/PyPSA/pull/1447


coroa:
Thanks, that is already helpful for us in the open-tyndp project.

PR #1935, opened 2025-12-04
Feature: Interactive balance maps

bobbyxng:
- Implementing interactive balance maps (equivalent of `plot_balance_map`) using PyPSA 1.0+ `n.explore()` https://docs.pypsa.org/latest/user-guide/plotting/explore/
- Tested and used in github.com/resilient-project/pypsa-nrw @Irieo @LukasFrankenQ 
- Renaming `plot_balance_map` to `plot_balance_map_static`, introducing `plot_balance_map_interactive`
- Refactoring existing code slightly and modified variable names for consistency across both rules
- Collector rule in `collect.smk` expands across both static and interactive maps. Introduction of sub-collectors `plot_balance_maps_static` and `plot_balance_maps_interactive`
- Snakemake `all` will output both static and interactive maps
- Using [underscore] "_" for carrier names, as spaces have run into breaking snakemake workflows in slurm environments (e.g. for co2 stored, urban central heat, etc.)
- Different scaling (however proportional within the html of course) as `n.explore()` can do auto-scaling based on maximum specified sizes. This however means, that proportionality is not guaranteed across different scenarios or runs, however makes inspection more user-friendly (I believe this is the right trade off).
- Introducing carrier-specific branch colors, modified standard colors and scaling in static map plotting
- Comparisons below

## Checklist
- [X] Changes in configuration options are added in `config/plotting.default.yaml`.

## Comparisons 
Static (left), interactive (right)

### AC
Screenshot From 2025-12-04 16-16-02

### co2 stored
Screenshot From 2025-12-04 16-16-28

### gas
Screenshot From 2025-12-04 16-16-54

### H2
Screenshot From 2025-12-04 16-18-33

### methanol (no regionalisation, single price)
Screenshot From 2025-12-04 16-18-56

### oil (no regionalisation, single price)
Screenshot From 2025-12-04 16-19-15

### Hover over price regions (co2 stored example):
image


config/plotting.default.yaml
fneum, line 1: Could you limit the breaking config changes in this file? In my opinion the renaming of settings is nice, but not worth breaking people's config files over it. I would also keep `balance_map` instead of `balance_map_interactive`.

config/plotting.default.yaml
```diff
@@ -142,107 +176,195 @@ plotting:
...
+      region_vmin:
+      region_vmax:
+      region_alpha: 0.8
+      conversion: 1_000_000
+      branch_color: darkseagreen
+      branch_width_max: 20
+      bus_size_max: 15000
+      arrow_size_factor: 2
+      map_style: road
+      tooltip: true
```
fneum, line 276: This section needs to be documented in `plotting.csv`.

scripts/plot_balance_map_interactive.py
```diff
@@ -0,0 +1,258 @@
...
+    region_cmap = settings.get("region_cmap") or "Reds"
+    region_alpha = settings.get("region_alpha") or 0.7
+    region_unit = settings.get("region_unit") or ""
+    branch_color = settings.get("branch_color") or "darkseagreen"
+    arrow_size_factor = settings.get("arrow_size_factor") or 1.5  # PyPSA default
+    bus_size_max = settings.get("bus_size_max") or 10000  # PyPSA default
+    branch_width_max = settings.get("branch_width_max") or 10  # PyPSA default
+    map_style = (str(settings.get("map_style")) or "road").lower()
+    map_style = VALID_MAP_STYLES.get(map_style, "road")
+    tooltip = settings.get("tooltip") or True
```
fneum: With our setting you can write code assuming that the config settings exist. Defaults are all handled in `config.default.yaml`.

fneum (changes requested):
Great job and extremely useful! My only concern is the number of breaking changes in `plotting.default.yaml` which could be reduced.

Also the trick with the `_` must be documented in `plotting.csv`.

Also don't forget the release note.

config/plotting.default.yaml
bobbyxng: yes I reverted to original key names prior to this PR, the interactive keys are named, accordingly

scripts/plot_balance_map_interactive.py
bobbyxng: good point, taken care of!

config/plotting.default.yaml
bobbyxng: done!

bobbyxng:
Thanks @fneum for the quick review and comments. Everything should be implemented now from my side, if I understood your points correctly :) 🟢

Issue #1930, opened 2025-12-01
Handle new datasets with new data layer


fneum:
Required follow-up to #1675.

```
rule retrieve_seawater_temperature
rule retrieve_hera_data_test_cutout
rule retrieve_hera_data
rule retrieve_bidding_zones
```

See https://github.com/open-energy-transition/pypsa-eur/pull/67/

PR #1956, opened 2026-01-06
Fix raw OSM build and release process

brynpickering:
Fixes issues found in a soft fork project.

Result of the two changes can be found in this zenodo record: https://zenodo.org/records/18161054

- Delete column filtering when preparing release. It was previously filtering out columns it would later want access to, after column renaming.
- Use invariant names for virtual buses. This means that when the OSM data updates, the virtual bus ID does not change (previously, the integer naming convention was based on the order of processing the raw data rows, which was not deterministic)

## Checklist

- [ ] For new data sources or versions, `these instructions <https://pypsa-eur.readthedocs.io/en/latest/data_sources.html>`_ have been followed.


scripts/build_osm_network.py
```diff
@@ -148,18 +148,23 @@ def _add_line_endings(buses, lines, add=0, name="line-end"):
     -------
         - pd.DataFrame: DataFrame containing the virtual bus endpoints with columns 'bus_id', 'voltage', 'geometry', and 'contains'.
     """
-    endpoints0 = lines[["voltage", "geometry"]].copy()
```
bobbyxng, line 151: propose to add some sort of sorting still, otherwise it could give the user a false sense of security, i.e., when the underlying data changes and bus 0 and bus 1 "swap". Sorting could e.g. be based on which end point is further North or West for example

scripts/build_osm_network.py
```diff
@@ -651,10 +656,9 @@ def _create_station_seeds(
     buses_to_rename = buses_to_rename.sort_values(
         by=["country", "lat", "lon"], ascending=[True, False, True]
     )
-    buses_to_rename["bus_id"] = buses_to_rename.groupby("country").cumcount() + 1
```
bobbyxng, line 654: generally very good idea, to use the endpoints for unique naming, I like it. However, I would suggest to ensure that this does not happen, e.g.:
"DE_way/25046847-2-110_0-380" (see test config in my main comment). This occurs because https://www.openstreetmap.org/way/25046847 has two circuits, first suffix "-2" denotes the circuit, "110" the voltage level, "_0" the endpont (consider sorting, see above), and "380" the voltage level of the virtual bus (but this should be 110 then, ideally, the script does not add another suffix, but uses the "110". Right now its contradicting/in the best case redundant. You would have, depending on the implementation of your choice, sth. like:
- "DE_way/25046847-2-110_0"
- "DE_way/25046847-2-110_1"
- "DE_way/25046847-1-380_0"
- "DE_way/25046847-1-380_1"

scripts/prepare_osm_network_release.py
```diff
@@ -90,7 +90,6 @@ def export_clean_csv(df, columns, output_file):
     Returns:
```
bobbyxng, line 94: In my test run, I still found a bug, starting from PyPSA v1. component index names are all consistently named "name", so right now, an error is thrown "bus_id"/"transformer_id"/... etc. is not found. This is not directly related to this PR, but please double-check :)

bobbyxng:
Thanks @brynpickering for the PR and suggestion, I like the idea very much to make the virtual buses more persistent/invariant, at the time of implementation, numbering was the easier way to go. Using endpoints is also a good idea, I made a few comments, see the individual conversations. Here are some summarised thoughts:
- Suggestion to implement some basic sorting (ensuring that bus0 stays 0 and bus1 stays 1 and does not "flip" at some point)
- Suggestion to clean up the names (needs to be addressed in other parts of the workflow/script), so namings like this "DE_way/25046847-2-110_0-380" do not happen (110 and 380 are "contradicting", ideally only one of the two suffixes is needed --> also unique).
- Open question: A virtual bus always connects two lines. Both endpoints go into the pool of virtual bus candidates. Before, there was a simple method to drop duplicates based on identical geometry and voltage (and to prioritise "real" substations over virtuals in the naming). I think this was fine before, as there was no "ambition" of having persistent virtual bus names. Now, if we do this correctly, we need to make sure, that we have an invariant/deterministic way of determining the "name-giving" endpoint ("entering" or "exiting" line). One simple idea could be to take the longer of the two lines (similar to taking the largest substation, if they are merged together).

brynpickering:
Thanks for the review @bobbyxng. I have been working with a fork that still uses pypsa <v1 hence the difference with the bug you've seen. PyPSA-Eur doesn't yet _require_ pypsa v1 (the config sets `>=0.35.0`) so I guess a fix needs to be cross-compatible.

scripts/build_osm_network.py
brynpickering: Are you suggesting that we _should_ have that resulting list of virtual buses? Or that it contains redundancies that we would rather avoid?

scripts/prepare_osm_network_release.py
brynpickering: fixed.

scripts/build_osm_network.py
brynpickering: ~fixed.~ At least, updated to something more robust by taking the first and last points along the lines rather than points on the bounding box described by the line. 

I don't think there is a neat way to sort these points in the way suggested. If the underlying data changes, i.e. the linestrings change, then it should be expected that the virtual buses could change. For them to _swap_ while the line remains the same would require the line to be redrawn in the other direction on OSM. This seems quite unlikely.

scripts/build_osm_network.py
brynpickering: When running with the config:

```yaml
countries: ['BE', 'DE', 'GB']

data:
  osm:
    source: build
    version: unknown
```

I don't get `DE_way/25046847-2-110_0-380`. In fact, I don't see how I could since `DE_way/25046847-2-110_0` doesn't make any reference to the 380kV circuit

brynpickering: EDIT: just had another look at this and now know what you mean. Will see if it can be handled gracefully

brynpickering: OK, that should be fixed.

bobbyxng: Sorry, I forgot to add my config, I tested it with lower voltages as well (to make sure the experimental distribution grid config also still works https://github.com/PyPSA/pypsa-eur/blob/master/config/examples/config.distribution-grid-experimental.yaml):

```
# SPDX-FileCopyrightText: Contributors to PyPSA-Eur <https://github.com/pypsa/pypsa-eur>
#
# SPDX-License-Identifier: CC0-1.0
run:
  name: "distribution-grid-experimental"

countries:
- AT
- BE
- CH
- DE

electricity:
  voltages: [63., 66., 90., 110., 132., 150., 220., 300., 330., 380., 400., 500., 750.]
  base_network: osm

transmission_projects:
  enable: false

data:
  osm:
    source: build
    version: unknown
```

brynpickering:
> Open question: A virtual bus always connects two lines.

Except at stubs, where the virtual bus is the end of one line and that is all. But for the other cases you mention, I see what you mean. 

> One simple idea could be to take the longer of the two lines (similar to taking the largest substation, if they are merged together)

Seems reasonable. As with other concerns around the underlying data changing, I think it's fine if this isn't strictly persistent if the OSM data changes and the longest line changes.

scripts/build_osm_network.py
bobbyxng: fair point!

scripts/build_osm_network.py
bobbyxng: looks very good @brynpickering, thank you!

bobbyxng:
@brynpickering: @lkstrp fixed some bugs in the upstream wrt. to workflow testing (https://github.com/PyPSA/pypsa-eur/pull/1992). Could you please merge it into your branch/PR? Only when the checks are successful, I can approve.

Thanks! 

Edit:// General comment: Missing push rights is due to this fork being managed by OET Github user group (which I am not part of). I don't know if there's a better way to handle this in the future, otherwise everything just takes longer and creates a lot of friction @fneum

fneum:
Normally, maintainers of the repositories for which PR is made gain push rights on the branch of the PR, but it might be blocked from OET-org side.

fneum:
Failing CI due to https://github.com/PyPSA/pypsa-eur/issues/1930, not this PR's fault.

brynpickering:
Thanks @bobbyxng ! Any chance you could publish a new zenodo version to the existing record, based on this update (and updating the OSM content more generally)?

bobbyxng:
> Thanks @bobbyxng ! Any chance you could publish a new zenodo version to the existing record, based on this update (and updating the OSM content more generally)?

Yes, just did not have the time yet :) I wanted to make another PR on improving the workflow (comparison of individual builds) and update the zenodo record in this process. Will prioritise this next! @brynpickering

PR #1960, opened 2026-01-09
feat: move archive from zenodo to s3 bucket

### LINKED ISSUE #1969: Add cutouts for 2024 & 2025 to Zenodo bundle
Would like to see: Weather Data Cutouts 2024.nc for PyPSA-Eur

Existing database has entries of 1996 to 2023.nc
https://zenodo.org/records/15349674

named like: [europe-2023-sarah3-era5.nc](https://zenodo.org/records/15349674/files/europe-2023-sarah3-era5.nc?download=1)

Would be useful for validation of Congestion Power Flow analysis, based on 2024 Cross-Border data from Entso-e
And based on PV and Day-Ahead electricity price data of 2024

These have to align with weather year used

Nice to have: data set of 2025 as well

lkstrp:
Closes #1969

Moves all archived data from zenodo to [Object Storage](https://www.hetzner.com/de/storage/object-storage/).

1. We could switch to [snakemake-storage-plugin-s3](https://github.com/snakemake/snakemake-storage-plugin-s3), but I don't see many reasons. This is a public store and auth is not needed, download speeds are already much faster than from zenodo and we don't need any extra dependencies.

2. Cutouts are currently served from a single versioned dictionary. To allow updating/ adding a single cutout, (without uploading all again (like we did on zenodo)), but still be able to access all from the newest version, I added a small MANIFEST lookup. That logic could also be moved to [`snakemake-storage-plugin-cached-http
`](https://github.com/PyPSA/snakemake-storage-plugin-cached-http), and we could use it for all dir based archives. The logic is simple: When you update file B and C for `v2`, but not file `A`, the MANIFEST simpliy points for files A to the upload version in `v1`, so all three files can be accessed under `v2` without reupload. It's similar to a symlink.

3. Zenodo allows for metadata, DOI's, long term storage which is append only (I can't change already released versions). I would argue that leaving Zenodo is good anyway, since we are missuings the system. DOI's and long term storage make sense for some precious research outputs/ results. What we are doing here is basically just caching input datasets which are sometimes slightly processed. That's not what Zenodo is for.
    - Metadata we could just add to the MANIFEST file.
    - Long term storage is up to us. As long as Hetzner is alive and we pay for the service, we can provide it. CERN can provide them a bit more long term I am sure. Long term storage is needed so people can rerun there workflows with pinned input data from the archive.
    - The bucket is currently not append only. Files are uploaded per version, but someone with access could come and just change old versions. So for reproducibility you would need to trust the maintainers. We could set it up though, it's more a question if this burden is necessary: [Object Lock: Legal Hold - Hetzner Docs](https://docs.hetzner.com/storage/object-storage/howto-protect-objects/protect-object-lock-legal-hold/)
    - DOI's are gone. Only way to get them back is mirror the bucket archive with zenodo. Which for some datasets where we need and wanna keep the Zenodo repo (@bobbyxng OSM ?), is easily possible.

4. A contributor will not have access to the bucket and will not be able to add data to the archive. There we have two options:
    - We keep it as is, the archive could hold data from any source and a contributor can add new ones. If needed we migrate the datasets at some point to the bucket (as done now).
    - A contributor can only add a dataset as a `primary` source. We then need to check it and add it to the archive. This also brings some security benefits that @amos-schledorn raised. With malicious intent, it could be possible (or at least is easier) to integrate defective data into the workflow either directly or by changing it afterwards. The review will most likely catch it, but there is no guarantee with very large diff PRs. If we manage the archive and use a Legal Hold/Append Only bucket, you could safely pin archived versions. Companies and organisations who care, would probably still not trust the process and serve only their own data. So again the question how much this is needed.

## Checklist

### DIFF
--- a/config/schema.json
+++ b/config/schema.json
@@ -1712,7 +1712,7 @@
             }
           }
         },
-        "cutouts": {
+        "cutout": {
           "description": "Configuration for a single data source.",
           "properties": {
             "source": {
@@ -12764,7 +12764,7 @@
             }
           }
         },
-        "cutouts": {
+        "cutout": {
           "description": "Configuration for a single data source.",
           "properties": {
             "source": {


data/versions.csv
```diff
@@ -1,101 +1,101 @@
...
 "country_hdd","build","unknown","['latest', 'supported']","","Latest dataset built using ERA5 runoff data. This takes a very long time to build."
-"natura","archive","2025-08-15","['latest', 'supported']","https://zenodo.org/records/16881818/files/natura.tiff",""
-"natura","archive","v0.8.2","['supported']","https://zenodo.org/records/16874772/files/natura.tiff",""
+"natura","archive","2025-08-15","['latest', 'supported']","https://fsn1.your-objectstorage.com/pypsa-wda/pypsa-eur/natura/2025-08-15/natura.tiff",""
+"natura","archive","v0.8.2","['supported']","https://fsn1.your-objectstorage.com/pypsa-wda/pypsa-eur/natura/v0.8.2/natura.tiff",""
 "natura","build","unknown","['latest', 'supported']","https://sdi.eea.europa.eu/datashare/s/tWpGXaWdWGYcqsL/download",""
-"bfs_road_vehicle_stock","archive","2024-03-13","['latest', 'supported']","https://raw.githubusercontent.com/PyPSA/pypsa-eur/5b5d308bf70f15dd0b107d8a19c121093dcdd5bd/data/gr-e-11.03.02.01.01-cc.csv",""
+"bfs_road_vehicle_stock","archive","2024-03-13","['latest', 'supported']","https://raw.githubusercontent.com/pypsa-wda/pypsa-eur/5b5d308bf70f15dd0b107d8a19c121093dcdd5bd/data/gr-e-11.03.02.01.01-cc.csv",""
```
fneum: intended?

fneum:
> We could switch to [snakemake-storage-plugin-s3](https://github.com/snakemake/snakemake-storage-plugin-s3), but I don't see many reasons.

Let's start without it and see how it's going.

> It's similar to a symlink.

Sounds elegant!

> Metadata we could just add to the MANIFEST file.

Yes, that's a good idea!

> DOI's are gone. 

For datasets that we merely mirror (or pre-processed a bit), this is totally fine. For primary datasets we produce (I mainly think of OSM grid dataset), we should still maintain a Zenodo version.

> So for reproducibility you would need to trust the maintainers. 

I think this is fine; what is important is that we keep old versions accessible to run older versions.

> We keep it as is, the archive could hold data from any source and a contributor can add new ones. If needed we migrate the datasets at some point to the bucket.

That sounds reasonable; contributors can request us placing their files in a bucket (if necessary) and we can request in review not using external archive URLs (if necessary).

We should adjust the [Data Sources](https://pypsa-eur.readthedocs.io/en/latest/data_sources.html) documentation.

Could you give us/me a private introduction how to manage the bucket?

fneum (approved):

coroa:
Thank you! Should we still keep zenodo as an alternative archive, maybe as a `zenodo` type alive?

Is there a way to retrieve hash from the object storage for checksumming? So that we update the cached-http plugin and can keep checksums? Debian mirrors add a `SHA512SUMS` file to each directory: https://cdimage.debian.org/debian-cd/current/amd64/iso-cd/SHA512SUMS

> Metadata we could just add to the MANIFEST file.

Ah yes, please.

data/versions.csv
lkstrp: No, thanks for the good catch

coroa:
Maybe a bit late, but the [quilt python library](https://docs.quilt.bio) provided quite some versioning and s3 bucket managing conventions. With some cataloguing services and command lines for managing. Not sure if that is needed at this point.

rules/retrieve.smk
```diff
@@ -308,10 +312,29 @@ rule retrieve_bidding_zones:
...
+            if filename not in manifest["files"]:
+                raise ValueError(
+                    f"Cutout '{filename}' not found in manifest for version {CUTOUT_DATASET['version']}. "
+                    f"Available: {list(manifest['files'].keys())}"
+                )
+            file_version = manifest["files"][filename]
+            return f"{CUTOUT_BASE_URL}/{file_version}/{filename}"
+        except urllib.error.HTTPError:
+            # No manifest, use URL directly
+            return f"{CUTOUT_DATASET['url']}/{filename}"
```
euronion: Can we do this on the object storage side? Create the symlink there? Having the logic here means we can no longer easily switch to a different provider, like a fallback to `zenodo` for the existing cutouts.

If you want to keep it here (I don't like having this code in the model in this location), can you add some more comments to keep it maintainable?

coroa: Why can't you put that into the dataset_version logic?

lkstrp: > Can we do this on the object storage side? Create the symlink there?

We can't. 

But the fallback still works. If there is no MANIFEST, the retrieval is just done as it was before:

```
        except urllib.error.HTTPError:
            # No manifest, use URL directly
            return f"{CUTOUT_DATASET['url']}/{filename}"
```

So you can just switch back to zenodo. But I agree, we should rather move that to the storage plugin (if @coroa agrees)

lkstrp: We could, but you need to create one entry per file then, and remove the directory logic, which is currently used in the workflow. If we add all cutouts from 1940, that would just bloat `versions.csv`. I rather thought keeping support for `directory` logic and just adding that there is no need to always reupload everything would be the cleanest. But, as I say in the comment above, I would move that to the storage plugin

coroa: Not sure, i understand how that would look like, but i am not against putting stuff into the plugin if we can make it somewhat generally useful.

coroa: Uh, @euronion , nice cross posting.

euronion (approved):

rules/retrieve.smk
euronion: @lkstrp Can't you just setup up a symlink in the object storage for the cutouts that are not updated?

Although, I would argue, that you should always update the cutouts, even if you do not expect changes.

euronion:
Can we have a different URL? The current URL for the object storage seems a bit fishy. Either through a redirect via pypsa.org , or just generally a nicer URL?
People know about Zenodo, but they likely don't know about `your-objectstorage.com`.

I also support @coroa suggestion of keeping the `zenodo` entries that we already have.

rules/retrieve.smk
euronion: uups

lkstrp:
Thanks for the quick review everyone!

> Thank you! Should we still keep zenodo as an alternative archive, maybe as a zenodo type alive?

> I also support @coroa suggestion of keeping the zenodo entries that we already have.

We can, you can just change the url back to zenodo in `versions.csv` and done. We could even also just add duplicated entries with `<version>-zenodo` tag. It would just never be the `latest` tag and just a backup. And then it works out of the box, no need to adjust anything.

> Is there a way to retrieve hash from the object storage for checksumming? So that we update the cached-http plugin and can keep checksums? Debian mirrors add a SHA512SUMS file to each directory: https://cdimage.debian.org/debian-cd/current/amd64/iso-cd/SHA512SUMS

There is (at least for files <5GB), but again, I don't see too much of a reason. We use checksums to check if we got the expected file, which could have been corrupted during transfer, or just the wrong file was served. In my understanding, corruption during transfer is not really an issue anymore, and we rather wanna check if the correct file was served. Since we get data and checksum from Zenodo, we only check that they did not mixed up files by some random bug and the file is not corrupted during transfer.

But on a bucket fully controlled by us, the checksum only checks corruption during transfer. If we mixed up the files by accident, you would just get another checksum. And you also don't check if Hetzner has some serious bugs because they also provide you the checksum.

Which simply means that if we want to check the identity of files (which I am in favour of), I would rather vote for hardcoding them for each archived version in `versions.csv`. Any reasons/discussions so far why we don't want to do this, @euronion? Or adding them also to the MANIFEST, but I think `versions.csv` is the better place

> Can we have a different URL? The current URL for the object storage seems a bit fishy. Either through a redirect via pypsa.org , or just generally a nicer URL?
People know about Zenodo, but they likely don't know about your-objectstorage.com.

I could move it to `assets.pypsa.org` or `data.pypsa.org`. Thought this is unnecessary overhead, but that no one knows `your-objectstorage.com` is a good point

rules/retrieve.smk
lkstrp: > Not sure, i understand how that would look like, but i am not against putting stuff into the plugin if we can make it somewhat generally useful.

I'll draft a PR

> @lkstrp Can't you just setup up a symlink in the object storage for the cutouts that are not updated?

That is not a file system, there are no symlinks or anything. It's a simple flat key value store. You can't even rename things (just copy and delete)

> Although, I would argue, that you should always update the cutouts, even if you do not expect changes.

I agree, that just comes with costs for us. If we upload all years for europe from 1940, that would be ~500GB per version. In best case we extend this to provide cutouts for more PyPSA Models, different cutouts (just DE, even if it's included within Eur and so on). That adds up quickly. Even if the extra TB is not too expensive, it's better to not reupload files which are expected to be nearly the same

euronion:
> > Thank you! Should we still keep zenodo as an alternative archive, maybe as a zenodo type alive?
> 
> > I also support @coroa suggestion of keeping the zenodo entries that we already have.
> 
> We can, you can just change the url back to zenodo in `versions.csv` and done. We could even also just add duplicated entries with `<version>-zenodo` tag. It would just never be the `latest` tag and just a backup. And then it works out of the box, no need to adjust anything.

Seems hacky, I'd have suggested `source="zenodo"`, but I don't mind the hacky way ;)

> Which simply means that if we want to check the identity of files (which I am in favour of), I would rather vote for hardcoding them for each archived version in `versions.csv`. Any reasons/discussions so far why we don't want to do this, @euronion? Or adding them also to the MANIFEST, but I think `versions.csv` is the better place
> 

We kept `data/versions.csv` small and simple for the beginning. Feel free to expand now.
Checksum was until now mostly for corruption during transfer. If we now use checksums for checking whether the data is the same / changed, aren't we duplicating some logic from `snakemake` now?

> > Can we have a different URL? The current URL for the object storage seems a bit fishy. Either through a redirect via pypsa.org , or just generally a nicer URL?
> > People know about Zenodo, but they likely don't know about your-objectstorage.com.
> 
> I could move it to `assets.pypsa.org` or `data.pypsa.org`. Thought this is unnecessary overhead, but that no one knows `your-objectstorage.com` is a good point

Yes, please. Both suggestions are nice, preference for `data.pypsa.org`.

rules/retrieve.smk
euronion: > Even if the extra TB is not too expensive, it's better to not reupload files which are expected to be nearly the same.

I strongly disagree on this. We can debate about not reuploading files that are identical, but files that are "nearly" the same are "different", i.e. should be uploaded again.

lkstrp: In general, yes. But I'm not sure about this case. What is the use case? The process I have in mind is as follows:

Let's say we start with:
`cutouts/v1`
- EUR_2019
- EUR_2013
- DE_2012

Then you want to add other years. There is no need to re-upload the 2013 cutout, so you would do:
`cutouts/v1.1`
- EUR_2010
- EUR_2011
- ...

Via the manifest/symlink, `EUR_2019`, `EUR_2013` and `DE_2012` would still be available with both versions, `v1.1` and `v1.0`.

If you add another attribute to the cutouts, you would need to reupload a new version for all of them and do the following:

`cutouts/v2`
- EUR_2010
- EUR_2011
- EUR_2013
- EUR_2019

The manifest wouldn't point to `DE_2012`, since there we didn't add the attribute. It's only available under `v1.1` and `v.1.0`. Or we reupload it as well. 

And this can be extended to any other area. This is also what has happened so far with Zenodo. There it's just a full copy of the same file (which is the same as using the MANIFEST).

"nearly" the same if you add another cutout, you don't have to recalculate all of them. You could if you want do, like with the `v2` above. Nearly the same just means, we can't be 100% sure, that ERA5 give us at two different retrieval points, the exact same data. And only to hedge against that, what you are arguing for, I don't think is worth it

fneum:
> In my understanding, corruption during transfer is not really an issue anymore, and we rather wanna check if the correct file was served.

The checksums were basically a tool to flag incomplete downloads.

amos-schledorn:
Thanks, @lkstrp! Would this setup (if not immediately, then in the future) allow for automatic security check-ups of the archived data? (searching for virusses/.exe files etc.) 
That would mitigate some human error of contributors reviewing and uploading data sets.

lkstrp:
Most of what we discussed is applied now. 
- `zenodo` is a now an alternative archive next to `archive`. Before I continue with docs and updating the cli tool. What is our goal here? Do we fully wanna mirror the s3 archive (`data.pypsa.org`) with zenodo? And keep them in sync all the time? It's not too hard to do that with the CLI tool, not sure though if this is necessary @euronion @coroa 
- I made a couple of changes to `versions.csv` and added a schema to enforce them. @euronion could you have a look again?
- Checksum and metadata will be in `MANIFEST.yaml` and the storage plugin can check it also from `data.pypsa.org`. But this is another PR over there
- Checksum within PyPSA-Eur (which I still think makes most sense), I might add in another PR. This should be optional anyways

fneum:
> Do we fully wanna mirror the s3 archive (data.pypsa.org) with zenodo?

I would say no, unless it's data we produced (like OSM).

test/test_data_versions_layer.py
```diff
@@ -0,0 +1,105 @@
...
+
+VALID_TAGS = {
+    "latest",
+    "supported",
+    "partially-supported",
+    "not-supported",
+    "deprecated",
+    "might-work",
+    "not-tested",
+    "broken link",
```
euronion: Tags with or without spaces? (`broken link` or `broken-link`) -> consistency

test/test_data_versions_layer.py
```diff
@@ -0,0 +1,105 @@
...
+from datetime import date
+from pathlib import Path
+
+import pandas as pd
+import pandera.pandas as pa
+from natsort import natsort_keygen
+from pandera.pandas import Check, Column
+
+VERSIONS_CSV = Path(__file__).parent.parent / "data" / "versions.csv"
+VALID_SOURCES = ["primary", "archive", "zenodo", "build"]  # Order matters
```
euronion: Order matters for what/why? (can you expand the comment?)

rules/common.smk
```diff
@@ -93,10 +93,8 @@ def load_data_versions(file_path):
         comment="#",
     )
 
-    # Turn 'tags' column from string representation of list to individual columns
-    data_versions["tags"] = data_versions["tags"].apply(
-        lambda x: json.loads(x.replace("'", '"'))
-    )
+    # Turn pipe-separated tags into individual columns
+    data_versions["tags"] = data_versions["tags"].str.split("|")
```
euronion: Just leaving my first reaction here: "ewwwkay". What about using " " (space) as a separator?

data/versions.csv
euronion, line 1: * I did prefer having the fields quoted with `" ..."` for the `csv` file. Was there a good reason to remove the quoting?
* Thank you for sorting by dataset and source

rules/retrieve.smk
```diff
@@ -835,7 +835,7 @@ if (TYDNP_DATASET := dataset_version("tyndp"))["source"] in ["primary", "archive
 
 
 
-if OSM_DATASET["source"] in ["archive"]:
+if OSM_DATASET["source"] in ["archive", "zenodo"]:
```
euronion: This adjustment is necessary for all datasets that are now on `zenodo` and in `archive`.

test/test_data_versions_layer.py
lkstrp: Order defines the sort priority, so it's: 
```
    """Sort by dataset (asc), version (desc, natural), source (as predefined)."""
```
predefined means the order here

rules/common.smk
lkstrp: Sold

config/config.default.yaml
```diff
@@ -1377,7 +1380,7 @@ data:
   mobility_profiles:
     source: archive
     version: latest
-  cutout:
+  cutouts:
```
euronion:
```suggestion
  cutout:
```

data/versions.csv
```diff
@@ -1,103 +1,126 @@
...
+wdpa,Jul2025,archive,latest supported,2025-12-02,"We are legally not allowed to redistribute this dataset, luckily the web archive is keeping copies of it.",https://web.archive.org/web/20250715071823if_/https://d1gam3xoknrgr2.cloudfront.net/current/WDPA_Jul2025_Public_shp.zip
+wdpa_marine,unknown,primary,latest supported,2025-12-02,"WDPA maritime changes its URL every month, the URL here is used as a template and {bYYYY} is replaced inside the retrieve.smk.",https://d1gam3xoknrgr2.cloudfront.net/current/WDPA_WDOECM_{bYYYY}_Public_marine_shp.zip
+wdpa_marine,Jul2025,archive,latest supported,2025-12-02,"We are legally not allowed to redistribute this dataset, luckily the web archive is keeping copies of it.",https://web.archive.org/web/20250715084308if_/https://d1gam3xoknrgr2.cloudfront.net/current/WDPA_WDOECM_Jul2025_Public_marine_shp.zip
+worldbank_urban_population,unknown,primary,latest might-work,2025-12-02,"This is the original World Bank API link, which is sometimes updated; it is not guaranteed to work with the current codebase and data changes without notice.",https://api.worldbank.org/v2/en/indicator/SP.URB.TOTL.IN.ZS?downloadformat=csv
```
euronion, line 125:
```suggestion
worldbank_urban_population,unknown,primary,latest supported,2025-12-02,"This is the original World Bank API link, which is sometimes updated; it is not guaranteed to work with the current codebase and data changes without notice.",https://api.worldbank.org/v2/en/indicator/SP.URB.TOTL.IN.ZS?downloadformat=csv
```

scripts/lib/validation/config/data.py
```diff
@@ -199,9 +203,9 @@ class DataConfig(BaseModel):
         default_factory=_DataSourceConfig,
         description="Mobility profiles data source configuration.",
     )
-    cutout: _DataSourceConfig = Field(
+    cutouts: _DataSourceConfig = Field(
         default_factory=_DataSourceConfig,
-        description="Cutout data source configuration.",
+        description="Cutouts data source configuration.",
```
euronion:
```suggestion
        description="Cutout data source configuration.",
```

scripts/lib/validation/config/data.py
```diff
@@ -199,9 +203,9 @@ class DataConfig(BaseModel):
         default_factory=_DataSourceConfig,
         description="Mobility profiles data source configuration.",
     )
-    cutout: _DataSourceConfig = Field(
+    cutouts: _DataSourceConfig = Field(
```
euronion:
```suggestion
    cutout: _DataSourceConfig = Field(
```

rules/retrieve.smk
```diff
@@ -334,15 +334,15 @@ rule retrieve_bidding_zones:
         "../scripts/retrieve_bidding_zones.py"
 
 
-if (CUTOUT_DATASET := dataset_version("cutout"))["source"] in [
+if (CUTOUT_DATASET := dataset_version("cutouts"))["source"] in [
```
euronion:
```suggestion
if (CUTOUT_DATASET := dataset_version("cutout"))["source"] in [
```

rules/common.smk
```diff
@@ -210,7 +208,7 @@ def solved_previous_horizon(w):
 
 def input_cutout(wildcards, cutout_names="default"):
 
-    cutouts_path = dataset_version("cutout")["folder"]
+    cutouts_path = dataset_version("cutouts")["folder"]
```
euronion:
```suggestion
    cutouts_path = dataset_version("cutout")["folder"]
```

data/versions.csv
```diff
@@ -1,103 +1,126 @@
...
+powerplants,0.7.0,archive,supported,2026-01-20,"Part of the `powerplantmatching` package and versioned on GitHub, i.e. no dedicated 'archive' entry.",https://data.pypsa.org/workflows/eur/powerplants/0.7.0
+scigrid_gas,1.1.2,primary,latest supported,2025-12-02,the primary is already from Zenodo published by original authors,https://zenodo.org/records/4767098/files/IGGIELGN.zip
+scigrid_gas,1.1.2,archive,latest supported,2026-01-20,the primary is already from Zenodo published by original authors,https://data.pypsa.org/workflows/eur/scigrid_gas/1.1.2
+scigrid_gas,1.1.1,archive,deprecated not-supported,2026-01-13,,https://data.pypsa.org/workflows/eur/scigrid_gas/1.1.1/IGGIELGN.zip
+scigrid_gas,1.1.0,archive,deprecated not-supported,2026-01-13,,https://data.pypsa.org/workflows/eur/scigrid_gas/1.1.0/IGGIELGN.zip
+seawater_temperature,v1.0,primary,latest supported,2026-01-21,Test cutout seawater temperature,https://zenodo.org/records/15828866/files/seawater_temperature.nc
```
euronion, line 112: Should be build, not primary. See the code changes we did in the not merged PR at https://github.com/open-energy-transition/pypsa-eur/pull/67 . Would be nice to get those changes in as well soon. (we don't have capa for them right now)

scripts/retrieve_seawater_temperature.py
```diff
@@ -68,7 +68,7 @@
     if snakemake.params.default_cutout == "be-03-2013-era5":
         logger.info("Retrieving test-cutout seawater temperature data.")
 
-        url = "https://zenodo.org/records/15828866/files/seawater_temperature.nc"
+        url = snakemake.params.test_data_url
```
euronion, line 71: Wrong, but right for now to get it merged

test/test_data_versions_layer.py
```diff
@@ -0,0 +1,123 @@
...
+is_sorted = Check(lambda df: df.equals(sort_versions(df)), error="Data must be sorted")
+archive_has_url = Check(
+    lambda df: df.loc[df["source"] == "archive", "url"].str.len().gt(0).all(),
+    error="Archive entries must have a URL",
+)
+one_latest_per_dataset_source = Check(
+    lambda df: df[df["tags"].str.contains("latest")]
+    .groupby(["dataset", "source"])
+    .size()
+    .le(1)
```
euronion: Should be exactly one?

test/test_data_versions_layer.py
```diff
@@ -0,0 +1,123 @@
...
+    ],
+    coerce=True,
+    strict=True,
+    ordered=True,
+    unique=["dataset", "version", "source"],
+)
+
+
+def load_versions() -> pd.DataFrame:
+    return pd.read_csv(VERSIONS_CSV, dtype=str, keep_default_na=False)
```
euronion, line 99: Let the records show that I was in favour of `quoting=ALL`

lkstrp: ⛔

lkstrp:
This is going to be merged now with a list of future todos:

- [ ] Couple datasets csv for docs with versions csv, export metadata in manifest and enforce documentation of datasets
- [ ] The documentation needs to list which cutouts are available from the archive
- [ ] Add new archive cli tool, update contributors documentation
- [ ] Update storage plugin to check checksums
- [ ] Change `seawater_temperature` from primary to build (#1930)
- [ ] Checksums within PyPSA-Eur (maybe)
- [ ] @euronion Comments

doc/data_sources.rst
```diff
@@ -18,88 +18,77 @@ data sources.
...
+Understanding ``versions.csv``
+*****************************
+
+The file ``data/versions.csv`` is the central registry for all data sources and their versions.
+Each row defines a specific version of a dataset with the following columns:
+
+* ``dataset``: The name of the dataset (e.g., ``worldbank_urban_population``).
+* ``version``: The version identifier, typically following the original data source's versioning (e.g., ``2025-08-14``).
+* ``source``: The source type - ``primary`` (original data source), ``archive`` (mirrored copy on ``data.pypsa.org``), or ``build`` (generated from other data).
+* ``tags``: Space-separated tags like ``latest``, ``supported`` or ``deprecated``.
```
euronion, line 47: Add all allowed options or ref to validation script

doc/data_sources.rst
```diff
@@ -18,88 +18,77 @@ data sources.
 
 .. _managing_data_versions:
 
-######################
-Managing Data Versions
-######################
+#################
+Data Versioning
+#################
```
euronion, line 23: Great, could move to the Contributing section (where the config validation stuff is residing now)

doc/data_sources.rst
```diff
@@ -18,88 +18,77 @@ data sources.
...
+*******************************
+
+If you notice that a data source has been updated and want to add the new version to PyPSA-Eur:
+
+1. Add a new row to ``data/versions.csv`` with the same ``dataset`` name, the new ``version``, ``source`` set to ``primary``, and the ``url`` pointing to the original data source.
+2. Set appropriate tags (typically ``latest supported``).
+3. Update the tags of the previous version (remove ``latest``, keep ``supported`` if still compatible).
+4. Create a pull request with your changes.
+5. Of course, any potential workflow adjustments should be considered and implemented as well.
+
```
euronion, line 70: 6. The project maintainers will create the corresponding archive version of the dataset on data.pypsa.org using the information that you have provided for the new entry in data/versions.csv

doc/data_sources.rst
```diff
@@ -18,88 +18,77 @@ data sources.
...
+1. Add a ``primary`` entry to ``data/versions.csv`` with a new unique dataset name, version, and URL pointing to the original data source.
+2. Implement a ``retrieve`` rule for your dataset in ``rules/retrieve.smk``.
+   Take inspiration from existing rules in the file.
+3. Add the new data source to:
+
+   * ``data`` section in the pydantic schema ``scripts/lib/validation/config/data.py``
+   * ``data_inventory.csv`` data inventory for PyPSA-Eur
+
+4. Create a pull request with your changes.
+
```
euronion, line 89: 5. The project maintainers will create the corresponding archive version of the dataset on data.pypsa.org using the information that you have provided for the new entry in data/versions.csv

doc/data_sources.rst
```diff
@@ -18,88 +18,77 @@ data sources.
...
+.. note::
+    Maintainers of the repository will create the corresponding ``archive`` entry after reviewing your contribution.
```
euronion, line 91: Delete

PR #1979, opened 2026-01-15
Fix atlite cutouts default value

bobbyxng:
- Bug fix: Revert to atlite default cutout value `europe-2013-sarah3-era5` instead of  `europe-1940-2024-era5`

## Checklist


fneum (approved):
Yes, that shouldn't have changed at all!

PR #1977, opened 2026-01-15
chore: fix formatting and improve consistency in data layer doc

tgilon:
Closes # (if applicable).

## Checklist

- [ ] I tested my contribution locally and it works as intended.
- [ ] Code and workflow changes are sufficiently documented.
- [ ] Changed dependencies are added to `pixi.toml` (using `pixi add <dependency-name>`).
- [ ] Changes in configuration options are added in `config/config.default.yaml`.
- [ ] Changes in configuration options are documented in `doc/configtables/*.csv`.
- [ ] For new data sources or versions, [these instructions](https://pypsa-eur.readthedocs.io/en/latest/data_sources.html) have been followed.
- [ ] A release note `doc/release_notes.rst` is added.

PR #1980, opened 2026-01-15
follow up on #1977

lkstrp:
@tgilon FYI: doctables don't exist anymore and description for config is done in the schema: https://github.com/PyPSA/pypsa-eur/blob/master/scripts/lib/validation/config/data.py


fneum:
Tak

tgilon:
Missed it, thank you for the heads-up  @lkstrp !

PR #1996, opened 2026-01-22
Add script path provider

brynpickering:
This creates a comparable path provider for `scripts` as is currently provided for other snakemake directives. 

The reason for adding it is so that if you use rule inheritance, you always get the correct script location, irrespective of the location of the snakefile that is undertaking the inheritance.

For instance, say you want to run a modified `build_electricity_demand` rule:

```python
use rule build_electricity_demand as test_rule with:
    output:
        resources("electricity_demand_test.csv"),
```

And this rule is defined in `rules/subdir/test.smk`

In the current approach, the script will be searched for at `"../scripts/build_electricity_demand.py"`. It is no longer there, but rather at `"../../scripts/build_electricity_demand.py"`. However, the `script`/`shell`/`run` directive of a rule cannot be modified when using rule inheritance, so you're currently stuck with the incorrect `"../scripts/build_electricity_demand.py"`. We have a real use-case in a soft-fork where this has caused a problem, forcing us to manually copy the whole rule into our snakefile.

The proposed solution sets absolute paths for the scripts which then work in rules inheriting those scripts. 

In theory, you could also update the script path provider to your own custom directory if you had `pypsa-eur` as a submodule but wanted to use your own scripts. I didn't design it for this edge case though. 

## Checklist

### DIFF
--- a/Snakefile
+++ b/Snakefile
@@ -16,6 +16,7 @@ from scripts._helpers import (
     get_scenarios,
     get_shadow,
     path_provider,
+    script_path_provider,
 )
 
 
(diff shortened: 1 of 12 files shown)


brynpickering:
damn, wrong `master` - two secs.

PR #1997, opened 2026-01-22
Add script path provider (take 2)

brynpickering:
(replaces #1996)
 
This creates a comparable path provider for `scripts` as is currently provided for other snakemake directives. 

The reason for adding it is so that if you use rule inheritance, you always get the correct script location, irrespective of the location of the snakefile that is undertaking the inheritance.

For instance, say you want to run a modified `build_electricity_demand` rule:

```python
use rule build_electricity_demand as test_rule with:
    output:
        resources("electricity_demand_test.csv"),
```

And this rule is defined in `rules/subdir/test.smk`

In the current approach, the script will be searched for at `"../scripts/build_electricity_demand.py"`. It is no longer there, but rather at `"../../scripts/build_electricity_demand.py"`. However, the `script`/`shell`/`run` directive of a rule cannot be modified when using rule inheritance, so you're currently stuck with the incorrect `"../scripts/build_electricity_demand.py"`. We have a real use-case in a soft-fork where this has caused a problem, forcing us to manually copy the whole rule into our snakefile.

The proposed solution sets absolute paths for the scripts which then work in rules inheriting those scripts. 

In theory, you could also update the script path provider to your own custom directory if you had `pypsa-eur` as a submodule but wanted to use your own scripts. I didn't design it for this edge case though. 

## Checklist

### DIFF
--- a/config/config.default.yaml
+++ b/config/config.default.yaml
@@ -125,6 +125,12 @@ electricity:
     contingency: 4000
   max_hours:
     battery: 6
+    "li-ion": 6
+    lfp: 6
+    vanadium: 10
+    lair: 12
+    pair: 24
+    "iron-air": 100
     H2: 168
   extendable_carriers:
     Generator:
--- a/config/plotting.default.yaml
+++ b/config/plotting.default.yaml
@@ -379,6 +379,12 @@ plotting:
     hydro: "Reservoir & Dam"
     battery: "Battery Storage"
     H2: "Hydrogen Storage"
+    li-ion: "Lithium-Ion Storage"
+    lfp: "Lithium-Ion-LFP Storage"
+    vanadium: "Vanadium-Redox-Flow Storage"
+    lair: "Liquid-Air Storage"
+    pair: "Compressed-Air-Adiabatic Storage"
+    iron-air: "Iron-Air Storage"
     lines: "Transmission Lines"
     ror: "Run of River"
     load: "Load Shedding"
@@ -525,6 +531,25 @@ plotting:
     land transport EV: '#baf238'
     land transport demand: '#38baf2'
     EV battery: '#baf238'
+    # all battery variation:
+    li-ion: '#ace37f'
+    li-ion charger: '#ace37f'
+    li-ion discharger: '#ace37f'
+    lfp: '#ace37f'
+    lfp charger: '#ace37f'
+    lfp discharger: '#ace37f'
+    vanadium: '#9B111E'
+    vanadium charger: '#9B111E'
+    vanadium discharger: '#9B111E'
+    lair: '#87CEEB'
+    lair charger: '#87CEEB'
+    lair discharger: '#87CEEB'
+    pair: '#003366'
+    pair charger: '#003366'
+    pair discharger: '#003366'
+    iron-air: '#edba1c'
+    iron-air charger: '#edba1c'
+    iron-air discharger: '#edba1c'
     # hot water storage
     water tanks: '#e69487'
     residential rural water tanks: '#f7b7a3'
--- a/config/schema.json
+++ b/config/schema.json
@@ -2000,6 +2000,42 @@
               "markdownDescription": "Maximum state of charge capacity of the battery in terms of hours at full output capacity `p_nom`. Cf. [PyPSA documentation ](https://pypsa.readthedocs.io/en/latest/components.html#storage-unit).",
               "type": "number"
             },
+            "li-ion": {
+              "default": 6,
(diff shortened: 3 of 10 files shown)


scripts/_helpers.py
```diff
@@ -153,6 +153,27 @@ def path_provider(dir, rdir, shared_resources, exclude_from_shared):
...
+        The root directory of the project (where the script directory is located).
+
+    Returns
+    -------
+    Callable[[str], Path]
+        A function that takes a script name as input and returns the full path to the script.
+    """
+
+    def _get_script_path(script: str) -> Path:
+        return Path("file://") / project_dir / "scripts" / script
```
coroa, line 172: I have no idea how this would behave under windows.

coroa: Why do you need `Path("file://")`? What do you think about `project_dir.resolve()` instead?

brynpickering: I assume `file://<path>` is a snakemake feature that gets parsed internally to `<path>` and since it's then being stored as a `pathlib.Path` it should be Windows compatible... Would the CI tell us either way or do I need to find a willing Windows user to test locally?

lkstrp: > Would the CI tell us either way or do I need to find a willing Windows user to test locally?

Only on Sundays currently 🤪
But you can turn the CI on here https://github.com/PyPSA/pypsa-eur/blob/630d37dd061fda6fff93c3a2c458dcdfdc9dcedd/.github/workflows/test.yaml#L50

Other stuff was not working at some point, but we might have passed that

At least it did not fail because of Windows reasons the last time: https://github.com/PyPSA/pypsa-eur/actions/runs/21106528317/job/60698884537

brynpickering:
@coroa the need is for when the rules are not all stored in a flat directory but that there is some nesting, e.g.:

```
rules/
├─ sub_rules/
│  ├─ subrule_file_1.smk
│  ├─ subrule_file_2.smk
├─ rule_file_1.smk
├─ rule_file_2.smk
scripts/
├─ script_1.py
```

Snakemake resolves the script path relative to the original file in which the rule is defined. If it is defined in `rules/rule_file_1.smk` but then in `rules/sub_rules/subrule_file_1.smk` that rule is inherited and modified, the script path is no longer valid as it is one level of nesting out. 

I'm not sure whether it is a bug per-se. It could be resolved by an upstream fix but I expect it's quite messy; I don't know how snakemake passes the inherited content to the new rule but I expect it doesn't keep much info on the filepath from which that content was derived. You'd need that to derive a relative path between the two rules from which you would get an updated script path.

I can see the benefit of being able to navigate to the script in the editor (I would like that for resources and logs but the path providers for those preclude that). Saying that, on my machine VSCode only recognises about 20% of them as paths that can be followed, so I regularly have to navigate to them separately anyway...

`file://` is used btw so that snakemake recognises it as an absolute path. I tried with standard absolute paths and snakemake parsed them as relative to the snakefile in which the rule has been originally defined.

coroa:
Ok, i did a small trial run at fixing that on the snakemake side and it unfortunately is quite convoluted where the script directive from the to-be-inherited rule is hidden within a callback that is copied over to the new rule w/o an easy way to rebuild that callback. So, i have to admit it is unlikely this will be fixed upstream any time soon.

The problem is that i don't know how those absolute script paths will work with submodule's and so on? Any thoughts?

brynpickering:
> The problem is that i don't know how those absolute script paths will work with submodule's and so on? Any thoughts?

Good point. Let me test it.

brynpickering:
@coroa tested with using pypsa-eur as a submodule with:

```py
configfile: "<path-to-pypsa-eur>/config/config.default.yaml"

module other_workflow:
    snakefile:
        # here, plain paths, URLs and the special markers for code hosting providers (see below) are possible.
        "<path-to-pypsa-eur>/Snakefile"
    config: config

use rule * from other_workflow as other_*
```

And it works as expected.

So, the only remaining issue is that it removes the possibility for clickable links in the IDE (a problem that we would ideally solve for all path providers!).

fneum:
I am happy to delegate the merge decision here to the majority of @lkstrp @brynpickering @coroa. Just make sure:

1. Windows compatibility (you already have that on the radar)
2. `mock_snakemake` compatibility

coroa:
I am fine with going ahead with this, i don't think the ide support can be universally fixed. Windows compatibility is a good point.

brynpickering:
@coroa [this CI check on Windows](https://github.com/PyPSA/pypsa-eur/actions/runs/21630577640/job/62363465499?pr=1997) seems to confirm that it works as expected. It fails, but not for reason of paths.

@yerbol-akhmetov has also tested a few rules on a Windows machine locally with no ill effects.

PR #1991, opened 2026-01-20
feat: add electricitymaps and entsoepy bidding zone data to data layer

daniel-rdt:
## Changes proposed in this Pull Request
This PR proposes to add `electricitymaps` and `entsoepy` bidding zone geospatial data to the data layer as a primary source.

## Checklist

- [ ] I tested my contribution locally and it works as intended.
- [x] Code and workflow changes are sufficiently documented.
- [ ] Changed dependencies are added to `pixi.toml` (using `pixi add <dependency-name>`).
- [ ] Changes in configuration options are added in `config/config.default.yaml`.
- [ ] Changes in configuration options are documented in `doc/configtables/*.csv`.
- [ ] For new data sources or versions, [these instructions](https://pypsa-eur.readthedocs.io/en/latest/data_sources.html) have been followed.
- [x] A release note `doc/release_notes.rst` is added.

PR #2007, opened 2026-01-24
fix: follow up on added bidding zones datasets

lkstrp:
Follow up on #1991

EDIT:
Actually not an issue of PyPSA-Eur, but of the storage plugin and the different compression which cloudflare uses (which proxies the bucket)

Fix in https://github.com/PyPSA/snakemake-storage-plugin-cached-http/pull/3

This PR introduces cache invalidation based on new datasets, which explains why this was not caught in the first place. And also adds the missed path changes


daniel-rdt:
> Follow up on #1991
> This PR introduces cache invalidation based on new datasets, which explains why this was not caught in the first place. And also adds the missed path changes

Good catch, thanks for fixing! @lkstrp

PR #2023, opened 2026-01-29
Fix raw custom costs

lindnemi:
In current master, when using a custom_costs.csv and specifying values in EUR/kW or EUR/GW, the usual preprocessing which translates it to MW is not applied.

This PR fixes that

## Checklist

### DIFF
--- a/scripts/process_cost_data.py
+++ b/scripts/process_cost_data.py
@@ -127,6 +127,10 @@ def prepare_costs(
 
         custom_costs = _convert_to_MW(custom_costs)
 
+        custom_costs = custom_costs.drop("planning_horizon", axis=1).value.unstack(
+            level=1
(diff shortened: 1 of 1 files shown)


doc/release_notes.rst
```diff
@@ -9,6 +9,8 @@ Release Notes
 Upcoming Release
 ================
 
+* Applied scaling of kW and GW to MW for custom_costs as well
```
tgilon:
```suggestion
* Applied scaling of kW and GW to MW for custom_costs as well (https://github.com/PyPSA/pypsa-eur/pull/2023).
```

tgilon (commented):
Make sense, good catch! It would probably be better to create a function than duplicate the code.

PR #1830, opened 2025-09-08
update GEM steel and use GEM cement plant tracker

fneum:
Closes #1774 
Closes #1775 

## Changes proposed in this Pull Request

Update GEM steel plant tracker and use initial release of GEM cement plant tracker.

For use in `build_industrial_distribution_key` rule.

Distribution mostly by clinker capacity. Small values for grinding.

Changes look plausible.

image

image

image

image

## Checklist

- [x] I tested my contribution locally and it works as intended.
- [x] Code and workflow changes are sufficiently documented.
- [x] Changed dependencies are added to `envs/environment.yaml`.
- [x] Changes in configuration options are added in `config/config.default.yaml`.
- [x] Changes in configuration options are documented in `doc/configtables/*.csv`.
- [x] Sources of newly added data are documented in `doc/data_sources.rst`.
- [x] A release note `doc/release_notes.rst` is added.

PR #2031, opened 2026-02-02
follow up on #1830

lkstrp:
Follow up on #1830 

Seems like a little auto merge was moving things to fast :D

@fneum In general, never add a placeholder. Either merge with primary only and change the default in `config.default.yaml` or let me just add the entry and pass Nextcloud Links etc. in the PR


fneum:
Thanks! Yes, that was unexpected :)

PR #2035, opened 2026-02-03
update co2 price time series

fneum:
Closes #1527 

using dataset from Instrat.pl

PR #2036, opened 2026-02-03
Update fossil fuel price time series

fneum:
goes on top of #2035 

smoothed with 6 month moving average (in €/MWh):

image

Here the TTF for comparison (from Ruhnau et al. https://doi.org/10.1038/s41560-023-01260-5)

image

### DIFF
--- a/.github/workflows/push-images.yaml
+++ b/.github/workflows/push-images.yaml
@@ -2,14 +2,12 @@ name: Push images
 
 on:
   push:
-
     branches:
     - master
   workflow_dispatch:
 
 env:
   IMAGE_NAME: pypsa/eur-dev-env
-  BASE_ENV: envs/linux-64.lock.yaml
 
 jobs:
   push-image:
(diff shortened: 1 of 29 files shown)


fneum:
@lkstrp unfortunately the `pydeflate` is not available on conda channels. Is that an issue?

lkstrp:
Yes, since we loose we wanna keep thee conda explicit lock files fully in sync. As discussed [here](https://github.com/PyPSA/pypsa-eur/pull/1886#issuecomment-3522234393). If we would move fully to pixi, that wouldn't be a problem anymore. 

Can we not avoid using pydeflate?

fneum:
> Can we not avoid using pydeflate?

This tool is very useful. It does deflation and currency conversion. Saves a lot of own code for common tasks.

There's also already an issue for conda-forge: https://github.com/jm-rivera/pydeflate/issues/37

fneum:
Functionality validated. RTM.

PR #2056, opened 2026-02-12
Add config schema updater & docs

### LINKED ISSUE #2014: Make pydantic model easily extendable with a configurable plugin option
The new pydantic model to validate the config is a welcome addition. What would be great for downstream projects is to be able to update parts of the model without having to touch too many core PyPSA-Eur files.

My proposal would be to allow a user to define a config updater somewhere (filepath could be referenced in the config, ignoring the validation inception, or be at an expected filepath like `config/update_schema.py`) and then this plugin would take in the base pydantic model and return an updated version of it.

E.g., to add a new section:

```py
from pydantic import create_model, Field, BaseModel
from scripts.lib.validation.config import ConfigSchema

class MyNewConfigSection(BaseModel):
    my_new_field: str = Field("foo")

def config_updater() -> ConfigSchema:
    new_schema = create_model(
        "CustomConfigSchema", 
        __base__=ConfigSchema, 
        new_section=MyNewConfigSection
    )
    return new_schema
```

To also update an existing section:

```py
from pydantic import create_model, Field, BaseModel
from scripts.lib.validation.config import ConfigSchema
from scripts.lib.validation.config.clustering import ClusteringConfig

class MyNewConfigSection(BaseModel):
    my_new_field: str = Field("foo")

def config_updater() -> ConfigSchema:
    cluster_config = create_model(
         "CustomClusteringConfig", 
         __base__=ClusteringConfig, 
         mode=(
             Literal["busmap", "custom_busmap", "administrative", "custom_busshapes", "foobar"],
             Field("busmap",description="...")
         )
    )
    new_schema = create_model(
        "CustomConfigSchema", 
        __base__=ConfigSchema, 
        new_section=MyNewConfigSection,
        clustering=cluster_config
    )
    return new_schema
```

And then in validating the config, it would make that update first:

```py
def validate_config(config: dict) -> ConfigSchema:
    """Validate config dict against schema."""
    config_schema = updater()
    return config_schema(**config)
```

brynpickering:
Closes #2014 

I spent a while thinking about where best to place the importers that would update the base config whilst not interfering too much with other files (avoiding future merge conflicts + import recursion issues).
They can't go in the snakefile as you need access to them to update the config _before_ you call snakemake.
I settled on an otherwise unused space for `__init__.py` in the `scripts.validation` directory. 

- Config schema update base class that can be subclassed to update / add config to the existing schema
- Rearranged validation config scripts to mitigate import recursion errors
- Added docs to explain how to use it (maybe should come first?)

## Checklist


doc/validation_dev.rst
```diff
@@ -168,3 +162,87 @@ ensuring the file path is set when file logging is enabled:
...
+If this code were stored in the script `scripts/_my_config_updates.py` then `scripts.lib.validation.__init__.py` would now include:
+
+.. code-block:: python
+
+    import scripts._my_config_updates
+
+This is sufficient for both updates to be imported.
+Several separate update scripts can exist and be imported into the schema as desired.
+
+.. autoclass:: lib.validation.config._base::ConfigUpdater
```
lkstrp, line 261: I guess the reference can be removed since it will not help anyone reading this

brynpickering:
> If you hide away the config updates for each script in ./scripts, it becomes complicated to understand which entries have been updated already and which came from the main config schema.

The idea is that these updaters could be anywhere. If using submodules you might be pulling in an updater from that submodule. If using a soft-fork you can choose where to put your updater. The source of truth of which updaters are being used will then be listed in `scripts.lib.validators.__init__.py` / `scripts.lib.validators.config_updates.py`

> Chained updates are currently not handled, and just the last update survives.

They are handled just fine. They will be handled in the order they are imported in `__init__.py` (soon to be `config_updates.py`). They will always be applied to the latest state of the config, after applying all previous updates.

> Which means just the import order in config_update/ currently __init__.py matters, which is alphabetically. 

That's true, chained updates are handled fine but fixing the order they appear in `config_updates.py` is necessary. This can be achieved by ruff skipping isort in that file.

> Why not rename it to something like config_updates.py? __init__.py is always annoying to switch to.

Sure. We'd just have to import `config_update.py` in that `__init__.py` to force the import resolution.

doc/validation_dev.rst
brynpickering: It is the base class they will need to subclass, so it seems reasonable to expose the methods they have available to them for that.

lkstrp:
What about
> Any ideas how solve this? It could just raise an error if a key was already updated

An already updated config entry could just be disallowed to get updated again

euronion:
> What about
> 
> > Any ideas how solve this? It could just raise an error if a key was already updated
> 
> An already updated config entry could just be disallowed to get updated again

What would we do in forks of forks of forks in that case?

brynpickering:
@lkstrp I don't see that it's an issue if you update it again. You can get the state of the config item and append to it so that you are merging the two:

```py
class ClusteringConfigUpdater(ConfigUpdater):
    """
    Update an existing config item (in this case, to add `foobar` as an option to the `clustering.mode` key).
    """

    NAME: str = "default.clustering.yaml"  # Currently has no effect

    def update(self) -> type[ConfigSchema]:
        # this will represent the version of clustering schema including any previous changes
        clustering_config = self.base_config().clustering.__class__ 
        current_description = clustering_config.model_fields["mode"].description or ""

        new_description = current_description + " (extra) foobar: new item."
        new_list = Literal[clustering_config.model_fields["mode"].annotation, "foobaz"]
        # You could feasibly add a custom validator at this point, too,
        # using `new_list = typing.Annotated[new_list, pydantic.AfterValidator(func)]`

        clustering_schema = self._apply_updates(
            __base__=clustering_config,
            mode=(new_list, Field("busmap", description=new_description)),
        )
        new_schema = self._apply_updates(
            clustering=(clustering_schema, Field(default_factory=clustering_schema))
        )
        return new_schema
```

As a user you might not want to merge if someone has already changed it, just in case:

```py
class ClusteringConfigUpdater(ConfigUpdater):
    """
    Update an existing config item (in this case, to add `foobar` as an option to the `clustering.mode` key).
    """

    NAME: str = "default.clustering.yaml"  # Currently has no effect

    def update(self) -> type[ConfigSchema]:
        # this will represent the version of clustering schema including any previous changes
        clustering_config = self.base_config().clustering.__class__ 
        if clustering_config != ClusteringConfig:
            raise ValueError(
                "Expected clustering config to be of type ClusteringConfig, but got "
                f"{clustering_config}. This likely means another config updater has already modified the clustering config in an unexpected way."
            )
        current_description = clustering_config.model_fields["mode"].description or ""

        new_description = current_description + " (extra) foobar: new item."
        new_list = Literal[clustering_config.model_fields["mode"].annotation, "foobaz"]
        # You could feasibly add a custom validator at this point, too,
        # using `new_list = typing.Annotated[new_list, pydantic.AfterValidator(func)]`

        clustering_schema = self._apply_updates(
            __base__=clustering_config,
            mode=(new_list, Field("busmap", description=new_description)),
        )
        new_schema = self._apply_updates(
            clustering=(clustering_schema, Field(default_factory=clustering_schema))
        )
        return new_schema
```

Then pydantic will raise an error when building the schema, giving that error message.

So, it is user-configurable.
```

**Comment by brynpickering** [comment]
I've added examples of how that check can be added to a user's `updater` method.

File: scripts/lib/validation/config/_base.py [hunk 2]
```diff
@@ -36,3 +43,71 @@ def values(self) -> Iterator[Any]:
...
+        self.base_config = base_config
+
+    def __init_subclass__(cls):
+        """Register any subclasses so they are automatically captured for use later."""
+        super().__init_subclass__()
+        _registry.append(cls)
+
+    @property
+    @abstractmethod
+    def NAME(self) -> str:
```
lkstrp: If you have plans here, fine to leave it in. But this should be lowercase? I see you wanna define a class attribute constant per subclass, but this is not a constant

brynpickering: @lkstrp I realised if I didn't do something with it now then I never would. The `name` is now used when generating the config (and schema). This allows a parallel default config to be generated for a project, again to avoid conflicts with files that are pulled in from upstream in soft forks.

brynpickering:
@lkstrp requesting re-review as I decided to implement a use for the `name` property and don't want to just sneak it through with your pre-approval.

lkstrp (approved):
No worries, that's great use and aligns file names. And for each soft fork/ schema update you get a new named set of files with `pixi run generate-config` without overwriting the old ones. Love it. I assume there will never be a need for two different schemas within one project/ soft fork and that we want to discourage that?

fneum:
I'll trust you both on this. @brynpickering, you may need to resolve a merge conflict for the merge to go through.

brynpickering:
> I assume there will never be a need for two different schemas within one project/ soft fork and that we want to discourage that?

@lkstrp Yeah, I'd say you want one schema per default config, which this produces. I.e., you'll get `schema.default.<some-name>.json` for your soft fork `config.default.<some-name>.yaml` config. There isn't a case where you would want to allow a separate schema for the same default config.

One issue one might come across is when submodularising, if the submodule is PyPSA-Eur and it runs its own config validation independently to the parent workflow's validation - whose schema gets used? I don't quite know what will happen then (but that's already an unresolved question with the current implementation, I believe)

PR #2070, opened 2026-02-17
rolling horizon: drop cyclicity and allow in solve_network

fneum:


scripts/solve_network.py
```diff
@@ -528,6 +529,13 @@ def prepare_network(
             n, limit_dict=limit_dict, planning_horizons=planning_horizons
         )
 
+    # rolling horizon disables cyclic storage
+    if rolling_horizon:
+        n.storage_units.state_of_charge_cyclic = False
```
tgilon, line 534: @fneum I think this should be `cyclic_state_of_charge`. Am I overlooking something?

FabianHofmann: yes it should be ...

tgilon: Fixed with https://github.com/PyPSA/pypsa-eur/pull/2213

fneum: Ups... my apologies.

PR #2072, opened 2026-02-17
annual nuclear energy availability factors

fneum:


euronion:
Did you get the data from IAEA PRIS? Might be worth noting as a comment in the file.

Also I think you are taking the latest year's values if the desired year is not in the data, not the closest.

fneum:
> Did you get the data from IAEA PRIS? Might be worth noting as a comment in the file.

Yes, documented in `date-repos.rst`

PR #2078, opened 2026-02-18
Add option to specify fixed DC-AC ratio for solar PV

fneum:


lindnemi:
- [ ] The DEA assumes a DC/AC ratio for rooftop solar of 1.05 (for residential) and 1.1 (for commercial). To correctly transform AC costs to DC the ratio should be differentiated from the solar-utility values. There are two options:
A) add a separate solar rooftop profile
B) don't clip the profile directly, but do a carrier specific clipping of the p_max_pu timeseries in `add_electricity`

PR #2086, opened 2026-02-19
Fix estimation for sequestration potentials

toniseibold:
Ensuring that the optimistic sequestration potential is equal or greater than the neutral estimation. In raw data e.g. `data/CO2JRC_OpenFormats/CO2Stop_DataInterrogationSystem/Hydrocarbon_Traps.csv` optimistic assumption is zero in some cases while neutral estimations are > 0.

Removing "conservative estimate Mt" from config since this leads to double counting of the sequestration potential in rule `build_clustered_co2_sequestration_potentials`.

- before: total sequestration potential of 89,982 Mt
- after:  total sequestration potential of 16,227 Mt

## Checklist

- [ ] Code and workflow changes are sufficiently documented. _not applicable_
- [ ] Changed dependencies are added to `pixi.toml` (using `pixi add <dependency-name>`). _not_applicable_
- [ ] Changes in configuration options are documented in `doc/configtables/*.csv`. _not applicable_
- [ ] For new data sources or versions, [these instructions](https://pypsa-eur.readthedocs.io/en/latest/data_sources.html) have been followed. _not applicable_
- [ ] A release note `doc/release_notes.rst` is added. _not applicable_


fneum:
@lisazeyen, if you have a moment, could you say whether that's correct? I think you once mentioned something specific about the accounting of the sequestration potentials.

lisazeyen:
I think we only considered high-density storage so far. But I am not sure if there is some double counting. I am currently on holidays until the 8/3 does this have some time? Then I could look at my notes at home. In my PhD thesis I wrote different numbers (total sequestration 18 Gt: 9 Gt gas,  5Gt oil, 4 Gt aquifer) and I am not sure where the mismatch is coming from, is it including onshore?

fneum:
@lisazeyen, I'm unsure how to judge this. Maybe you have the chance to have a look at your notes in the upcoming days?

scripts/build_co2_sequestration_potentials.py
```diff
@@ -320,4 +320,11 @@ def merge_maps(
 
     gdf = merge_maps(traps_map, storage_map)
 
+    types = [" OIL", " GAS", " aquifer", ""]
```
lisazeyen: It should be either just the high-density storage or the conservative storage potential
```suggestion
    types = [" OIL", " GAS", " aquifer"]
```

lisazeyen:
@fneum @toniseibold sorry for the late reply. 

One should include just one of the two: either the conservative estimation or only consider the high-density storage (gas, oil, aquifer). We actually already discussed this in this [PR](https://github.com/PyPSA/pypsa-eur/pull/1010).

I would recommend to use only high-density storage because this is where most projects are planned. @toniseibold I still wonder why the numbers are slightly different in my PhD thesis compared to what is in the PR description, but might be just that I rounded the numbers.

scripts/build_co2_sequestration_potentials.py
```diff
@@ -320,4 +320,11 @@ def merge_maps(
...
+    for t in types:
+        opt_col = f"optimistic estimate{t} Mt"
+        neu_col = f"neutral estimate{t} Mt"
+
+        gdf[opt_col] = gdf[[opt_col, neu_col]].max(axis=1)
```
fneum: Is this still necessary one @lisazeyen 's suggested changes are implemented? I would avoid that "correction" and work with the actual values in the dataset if possible.

scripts/build_co2_sequestration_potentials.py
fneum: Thanks, @lisazeyen! I agree with the change.

fneum (changes requested):
Sounds good. Thanks @lisazeyen for checking (again)!

@toniseibold, I would avoid the `.max(axis=1)` hotfix if possible. See comment below.

toniseibold:
Thanks for taking a look @lisazeyen.
I think the description of the config was leading me in the wrong direction thinking that `conservative estimate Mt` is a summary of `[conservative estimate GAS Mt, conservative estimate OIL Mt, conservative estimate aquifer Mt]`.

I revert most of my proposed changes and added an explanation in `sector.py`.
What's also worth noting is, that including `conservative estimate Mt` leads to some distinct changes in model results. Especially in myopic runs, it becomes attractive to build DAC in Spain/Italy and sequester CO2 to offset emissions in the rest of Europe.

fneum (approved):

PR #2088, opened 2026-02-19
Heuristically split existing solar capacities 50:50 between rootop and utility scale

lindnemi:
Differences in the myopic test case are minor and occur at only one bus, but are aligned with the purpose of this change.

Before: 
image

After:

image

2030 values change, but 2050 values are little affected

## Checklist


lindnemi:
50:50 is a heuristic value. Since rooftop share between countries differs largely, a spatially resolved split would be more accurate, but there is no data.

For the whole of the EU the current rooftop:utility split is probably 60:40 (according to https://www.solarpowereurope.org/insights/outlooks/eu-solar-market-outlook-2025-2030/detail)

I chose 50:50 because some countries may be well below 60:40 and since rooftop installations are declining. But perhaps 60:40 would be better

fneum (approved):

PR #2099, opened 2026-02-25
feat: setup pypsa-bot github app

lkstrp:
- First version of pypsa-bot github app
- Currently handles pixi tasks, e.g. `generate-config` and `sync-locks` which acts similar to the setup pre-commit
- The pixi tasks are mainly nice to have and could be handled in GitHub actions as well (with auth via app for PRs from forks), but I wanna get started with this simple use case before extending for pypsa-app/ validation run access in PR comments

### DIFF
--- a/doc/release_notes.rst
+++ b/doc/release_notes.rst
@@ -9,6 +9,8 @@ Release Notes
 .. Upcoming Release
 .. =================
 
+* refactor: Use scripts path provider consistently (https://github.com/PyPSA/pypsa-eur/pull/2093).
+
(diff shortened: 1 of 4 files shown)


daniel-rdt:
@lkstrp Do I understand correctly, that this will be extendable to also commit such changes directly to the PR branches (analogue to what e.g. the pre-commit bot is already doing). But this is not yet set-up, correct?

lkstrp:
I hope it is setup already. This PR is already from my fork. I don't think it will work on OET forks because pre-commit can't push there as well because it's an org account. If maintainers have access it should work

PR #2105, opened 2026-02-26
feat: improve load shedding with carrier specific options and load sinks

daniel-rdt:
This PR proposes to improve the previous load shedding implementation by adding the option of modelling carrier specific load shedding with their own shedding costs.

The PR makes use of PyPSA-Eur's new config validation feature and adds new config options underneath `solving:options:load_shedding`:
```
class _LoadSheddingConfig(ConfigModel):
    """Configuration for `solving.options.load_shedding` settings."""

enable: bool = Field(
    False,
    description="Enable load shedding by adding high-cost generators to avoid infeasibilities. Requires either all_carriers: true or at least one entry in carriers.",
)
default_cost: PositiveFloat = Field(
    100000,
    description="The default cost for load-shedding in the unit of the bus carrier (e.g. EUR/MWh for electricity, EUR/t_CO2 for CO2). Must be positive.",
)
all_carriers: bool = Field(
    True,
    description="Switch to apply load shedding to all carriers. Otherwise, load shedding will be applied to listed carriers only.",
)
carriers: dict[str, PositiveFloat] = Field(
    {},
    description="Dictionary of carriers and their specific load shedding cost in the unit of the bus carrier (e.g. EUR/MWh for electricity, EUR/t_CO2 for CO2). If load shedding is enabled for all carriers, the default cost is assumed for non-listed carriers.",
)
```
These options are also validated, so that it is not possible to enable load_shedding without either setting `apply_to_all_carrier: True` or specifying carriers to add load shedding for.

The PR furthermore proposes to add an option that allows adding `load_sinks` (negative load shedding) to the model in the same logic as for regular `load_shedding`. This PR introduces a new set of configuration options analogue to the load shedding implementation.

```
class _LoadSinksConfig(ConfigModel):
    """Configuration for `solving.options.load_sinks` settings."""


enable: bool = Field(
    False,
    description="Add load sinks by adding negative-cost, energy consuming generators to avoid infeasibilities by absorbing excess energy. Requires either all_carriers: true or at least one entry in carriers.",
)
default_cost: PositiveFloat = Field(
    100000,
    description="The default cost for load sinks in the unit of the bus carrier (e.g. EUR/MWh for electricity, EUR/t_CO2 for CO2). Must be positive.",
)
all_carriers: bool = Field(
    False,
    description="Switch to add load sinks for all carriers. Otherwise, load sinks will be added for listed carriers only.",
)
carriers: dict[str, PositiveFloat] = Field(
    {},
    description="Dictionary of carriers and their specific load sink cost in the unit of the bus carrier (e.g. EUR/MWh for electricity, EUR/t_CO2 for CO2). If load sinks are added for all carriers, the default cost is assumed for non-listed carriers.",
)
```

## Checklist

### DIFF
--- a/config/config.default.yaml
+++ b/config/config.default.yaml
@@ -1061,7 +1061,7 @@ industry:
   MWh_CH4_per_tMeOH: 10.25
   MWh_MeOH_per_tMeOH: 5.528
   hotmaps_locate_missing: false
-  reference_year: 2019
+  reference_year: 2023
   oil_refining_emissions: 0.013
 
 # docs in https://pypsa-eur.readthedocs.io/en/latest/configuration.html#costs
--- a/config/schema.default.json
+++ b/config/schema.default.json
@@ -2615,7 +2615,7 @@
           "type": "boolean"
         },
         "reference_year": {
-          "default": 2019,
+          "default": 2023,
           "description": "The year used as the baseline for industrial energy demand and production. Data extracted from `JRC-IDEES 2015 <https://data.jrc.ec.europa.eu/dataset/jrc-10110-10001>`_.",
           "markdownDescription": "The year used as the baseline for industrial energy demand and production. Data extracted from [JRC-IDEES 2015 ](https://data.jrc.ec.europa.eu/dataset/jrc-10110-10001).",
           "type": "integer"
@@ -11745,7 +11745,7 @@
           "type": "boolean"
         },
         "reference_year": {
-          "default": 2019,
+          "default": 2023,
           "description": "The year used as the baseline for industrial energy demand and production. Data extracted from `JRC-IDEES 2015 <https://data.jrc.ec.europa.eu/dataset/jrc-10110-10001>`_.",
           "markdownDescription": "The year used as the baseline for industrial energy demand and production. Data extracted from [JRC-IDEES 2015 ](https://data.jrc.ec.europa.eu/dataset/jrc-10110-10001).",
           "type": "integer"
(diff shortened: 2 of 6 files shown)


config/config.default.yaml
```diff
@@ -1134,7 +1134,16 @@ adjustments:
 solving:
   options:
     clip_p_max_pu: 0.01
-    load_shedding: false
+    load_shedding:
+      enable: false
+      default_price: 100000
+      apply_to_all_carriers: true
+      carriers: {}
```
fneum: Could be shorter `all_carriers`.

config/config.default.yaml
```diff
@@ -1134,7 +1134,16 @@ adjustments:
...
+    load_sinks:
+      enable: false
+      default_price: 100000
```
fneum: I would go with the term `default_cost` here in both cases.

fneum (approved):
Neat! Two micro-suggestions.

config/config.default.yaml
daniel-rdt: Changed it and updated the PR description accordingly

config/config.default.yaml
daniel-rdt: Changed it and updated the PR description accordingly

euronion:
What do you think about moving the load shedding out of `solve_network.py`? It is adding components to the network, which I also found strange to happen inside this (late) step of the network.

daniel-rdt:
> What do you think about moving the load shedding out of `solve_network.py`? It is adding components to the network, which I also found strange to happen inside this (late) step of the network.

Hm, I'm not sure tbh. Somehow, I feel load shedding makes sense as a solving setting and it makes sense to add all solving settings in `solve_network.py`. But I see your point about adding network components at this stage.

fneum:
Yes, could be earlier. But wouldn't change it now.

PR #2113, opened 2026-03-04
Fix corine bool and offshore keyerror

MarcoAnarmo:
Closes #2111 and #2112.

Fixes two detected bugs in the renewable availability/profiles pipeline when using offshore technologies with `corine: false` and non-standard country lists:

- `determine_availability_matrix _MD_UA.py`: Convert `corine `to an empty dict when set to false, preventing `TypeError `on `"key" in corine` checks.
- `build_renewable_profiles.py`: Filter offshore buses against clustered regions index to avoid `KeyError `for buses dropped during clustering.

Used config file: [config.osm_eem.yaml](https://github.com/user-attachments/files/25741190/config.osm_eem.yaml)

## Checklist

### DIFF
--- a/config/config.default.yaml
+++ b/config/config.default.yaml
@@ -1100,6 +1100,7 @@ clustering:
     remove_islands: false
     aggregate_to_tyndp: false
   simplify_network:
+    to_380: true
     to_substations: false
     remove_stubs: true
     remove_stubs_across_borders: false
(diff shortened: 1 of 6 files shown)


fneum:
Could you divide the suggested changes into multiple separate PRs? Then we can decide / discuss one by one. Not all of them are related / dependent on each other.

PR #2131, opened 2026-03-28
fix: remove Dask client.shutdown() in build_renewable_profiles

thogin:

Removes the explicit `client.shutdown()` call at the end of `build_renewable_profiles.py`. This call can fail with a `TimeoutError` when Dask worker processes don't exit within the nanny timeout on shared filesystems (e.g. BeeGFS), converting a completed run into a false Snakemake failure.

Both output files (`.nc` profile and `.geojson` class regions) are written before the shutdown call. The other Dask-using scripts in the repo (`build_hac_features`, `build_line_rating`, `build_daily_heat_demand`) do not call `client.shutdown()` and rely on process exit for cleanup — this change aligns `build_renewable_profiles` with that pattern.

Observed on a SLURM/BeeGFS cluster: 3 false failures out of ~130 runs (2.3%) over 4 weeks, all during the Dask teardown phase after successful computation and output file writing.

## Checklist

### DIFF
--- a/scripts/build_renewable_profiles.py
+++ b/scripts/build_renewable_profiles.py
@@ -326,6 +326,3 @@ if __name__ == "__main__":
         ds["profile"] = ds["profile"].where(ds["profile"] >= min_p_max_pu, 0)
 
     ds.to_netcdf(snakemake.output.profile)
-
-    if client is not None:
-        client.shutdown()


lkstrp (approved):
Thanks for tracking that down @thogin !

coroa:
@thogin This pattern also had a sense, because without it relatively often the main process would stop and the workers then complained noisily that they did not find their nanny anymore.

Without knowing about the PR here, I reintroduced this pattern consistently everywhere in #2137 (but made dask.distributed also the non-default).

And i used now `atexit.register(client.shutdown)` which swallows exceptions by default so your timeouts should only spew out noise.

PR #2144, opened 2026-04-10
feat: split console and file logging levels

FabianHofmann:

Splits the logging level between the console and the rule log file so that parallel atlite-based rules stay readable on the console while the log file still captures full `INFO` output.

`configure_logging` now sets per-handler levels:

- `logging.level` — root log level (unchanged default: `INFO`)
- `logging.console_level` — stream handler (new, default: `WARNING`)
- `logging.file_level` — file handler (new, defaults to `level`)

Users who want the old behaviour set `logging.console_level: INFO`.

Follow-up on the discussion in https://github.com/PyPSA/pypsa-eur/pull/2144#issuecomment-4243259364. The xarray MultiIndex fix has been dropped from this PR and is now tracked upstream in https://github.com/PyPSA/atlite/pull/501.

## Checklist

**Required:**

### DIFF
--- a/config/config.default.yaml
+++ b/config/config.default.yaml
@@ -11,6 +11,8 @@ tutorial: false
 # docs in https://pypsa-eur.readthedocs.io/en/latest/configuration.html#logging
 logging:
   level: INFO
+  console_level: WARNING
+  file_level:
   format: "%(levelname)s:%(name)s:%(message)s"
 
 # docs in https://pypsa-eur.readthedocs.io/en/latest/configuration.html#remote
--- a/config/schema.default.json
+++ b/config/schema.default.json
@@ -2848,7 +2848,7 @@
       "properties": {
         "level": {
           "default": "INFO",
-          "description": "Restrict console outputs to all infos, warning or errors only",
+          "description": "Root log level. Messages below this level are discarded before reaching any handler.",
           "enum": [
             "DEBUG",
             "INFO",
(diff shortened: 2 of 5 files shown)


coroa:
I disagree with removing the log messages. And i would prefer solving the multi index warning at atlite level.

coroa:
Fixed in https://github.com/PyPSA/atlite/pull/501

FabianHofmann:
> I disagree with removing the log messages.

The reason being? With the faster and parallel execution of the atlite-based rules, you can't see which log refers to which conversion. The main time benchmarking logs (which state the tech) we keep. Also snakemake rules have descriptive messages nowadays. I don't see a reason to keep such kind of descriptive logs. If it is about debugging we should change the logger mode at least.

coroa:
I think instead of removing log messages we should change where they are displayed.

Ideally, we would create a logging sink that the snakemake logger plugin pulls up and then instead of showing the log messages on the console only forward them there.

The console would then be replaced with an interactive tui where you see which jobs are running and you can choose to see the logs of them.

As immediate solution, maybe we decide instead that info messages should only go into the log file and only warning messages are printed onto the console. I'd be fine with such a change happening globally in PyPSA-Eur (or locally for that matter)

The only two things missing then to make log files more helpful would be: A bridge that ensures that progressbars also write a minimal version into the log, and a default exception trap that also logs exceptions into the log.

FabianHofmann:
@coroa pr is taking a different approach now, see updated summary above.

fneum:
I'll let you two decide/agree on that. For me it would be important that there's control over how much is shown in console and that log files can store a more complete log. That seems to be the case with the current description.

PR #2147, opened 2026-04-13
Fix DC junctions in simplify_network

bobbyxng:
@brynpickering @euronion  
Same bug present in PyPSA-Eur upstream.

- Fixes overly aggressive DC link simplification in `simplify_network` (literal edge case example for new UK/Shetland HVDC and multi-terminal setup).
- Updates supernodes so only true chain nodes (degree 2, also applied to DC) are collapsed. Preserves DC junction buses/multi-terminals (degree 3+) to avoid dropping branches in topologies like Shetland to onshore branching links.
- Resolves merge conflicts with upstream documentation migration by keeping the release-note entry in `doc/release_notes.md`.

**Original base network topology**  

**Before fix**  
`remove_stubs: true`  

`remove_stubs: false`  

**After fix**  
`remove_stubs: true`  

`remove_stubs: false`  

## Checklist

**Required:**

**If applicable:**


bobbyxng:
Clustered example 50 nodes with fix
image

bobbyxng:
I had to add a new key, named it `allow_ac_dc_mix` for now to make this decision transparent/user configurable. Specifically, now that more DC buses "survive", it can happen that AC and DC buses in the same region (depending on clustering mode) are to be clustered. The question is, do you keep them separate (maintaining individual bus carriers) or allow DC buses to be clustered with AC buses (in this case all bus carriers become AC).
- "Visually" this looks the same, however, the underlying subnetworks and topology, number of lines and links change.
- For example, using `config/test/config.clusters.yaml` (administrative clustering)
  - `allow_ac_dc_mix: false` → 54 buses, 10 subnetworks (AC and DC kept separate)
  - `allow_ac_dc_mix: true` → 52 buses, 8 subnetworks (mixed clusters coerced to AC)

Open question, what to set as default. I believe the closest to replicate old behaviour is to allow mixing of AC and DC, but the technically more correct way would be to set `allow_ac_dc_mix: false`, as you also could have congestion between the AC and DC buses. Any thoughts? @fneum @brynpickering, maybe @martacki (given your clustering expertise :))

scripts/cluster_network.py
```diff
@@ -686,6 +714,9 @@ def update_bus_coordinates(
                 features=features,
             )
 
+        allow_ac_dc_mix = params.cluster_network.get("allow_ac_dc_mix", False)
```
brynpickering: there shouldn't be a need to set a default here any more since our config validation means there is always a value for this config item in any snakemake run. Adding a default here means you have to maintain the default in two places (the validation config and in this script)
```suggestion
        allow_ac_dc_mix = params.cluster_network["allow_ac_dc_mix"]
```

scripts/cluster_network.py
```diff
@@ -431,6 +431,33 @@ def clustering_for_n_clusters(
     return clustering
 
 
+def apply_carrier_mixing_policy(
+    n: pypsa.Network, busmap: pd.Series, allow_ac_dc_mix: bool
+) -> pd.Series:
+    """Apply carrier mixing policy to a busmap before clustering."""
```
brynpickering: Extend docstring to describe params and return

scripts/cluster_network.py
```diff
@@ -586,10 +613,11 @@ def update_bus_coordinates(
     admin_regions["y"] = admin_regions["poi"].y
 
     busmap_df = pd.DataFrame(busmap)
+    busmap_df["admin"] = busmap_df["busmap"].astype(str).str.split("::", n=1).str[0]
```
brynpickering: any possibility that a bus could have `::` in it that you didn't put there?

fneum:
I would suggest `allow_ac_dc_mix: false` as default, since mixing could be heavy simplification of network topology. Don't mind the extra buses.

Not sure the name is expressive enough though. Just looking at the name, one could assume separate synchronous zones (like Balearic Islands, DK etc.) can be clustered / simplified.

scripts/cluster_network.py
bobbyxng: Good point! I am sure that there are no other `::` but I cannot ensure that for other future PRs. So i made the code more robust to use `sjoin_nearest` instead of using substrings.

scripts/cluster_network.py
```diff
@@ -415,11 +415,6 @@ def clustering_for_n_clusters(
     bus_strategies.setdefault("substation_lv", lambda x: bool(x.sum()))
     bus_strategies.setdefault("substation_off", lambda x: bool(x.sum()))
 
-    # TODO Quick Fix for osm-prebuilt-version 0.6
-    for way_i in ["way/140248154", "way/975637991"]:
-        if way_i in n.buses.index:
-            n.buses.loc[way_i, "carrier"] = "AC"
-
```
bobbyxng, line 409: @FabianHofmann Can you double-check that this indeed can be removed? I don't see the need to keep this anymore after the changes in this PR

bobbyxng:
Thanks both @brynpickering and @fneum, I think I addressed all your comments. Regarding the config term, how about `allow_ac_dc_mixing_in_bus_clusters`? This should be more explicit.
Also, I removed an old hotfix in `cluster_network`, maybe @FabianHofmann can confirm that it is indeed okay to remove the hotfix.

fneum (approved):

fneum:

fneum (approved):

PR #2137, opened 2026-04-07
perf: use chunks="auto" in a single thread

coroa:
## Changes proposed in this Pull Request

Alternative to #2135 and #2136 , that simply relies xarray's automatic chunk selection which falls back to the netcdf stored chunks and picks nprocesses=1 in the default xarray backend to clock in similar speed improvements as #2135 .

on master:
```
❯ time pixi run snakemake prepare_sector_networks -c16 -R build_renewable_profiles determine_availability_matrix --until build_renewable_profiles
[...]
________________________________________________________
Executed in   23.60 mins    fish           external
   usr time  160.59 mins    0.13 millis  160.59 mins
   sys time   38.55 mins    1.06 millis   38.55 mins
```

on this perf branch:
```
❯ time pixi run snakemake prepare_sector_networks -c16 -R build_renewable_profiles determine_availability_matrix --until build_renewable_profiles
[...]
________________________________________________________
Executed in  307.62 secs    fish           external
   usr time   20.88 mins  670.00 micros   20.88 mins
   sys time    3.67 mins  108.00 micros    3.67 mins
```

ie. 24min reduces to 5min (out of which 21s are snakemake's dag generation).

## Checklist

**Required:**
- [x] Changes are tested locally and behave as expected.
- [ ] Code and workflow changes are documented.
- [x] A release note entry is added to `doc/release_notes.rst`.

**If applicable:**
- [x] Changes in configuration options are reflected in `scripts/lib/validation`.
- [ ] For new data sources or versions, [these instructions](https://pypsa-eur.readthedocs.io/en/latest/data_sources.html) have been followed.
- [ ] New rules are documented in the appropriate `doc/*.rst` files.

PR #2149, opened 2026-04-14
fix: remove asynchronous model for distributed client

coroa:

Fixup of #2137 , I took `Client(asynchronous=True)` from build_daily_heat_demand and build_temperature_profiles without investigating what it does.

https://distributed.dask.org/en/stable/api.html#client :
```
asynchronous: bool (False by default)
    Set to True if using this client within async/await functions or within Tornado 
    gen.coroutines. Otherwise this should remain False for normal use.
```

We and atlite are not running our own async event loop so this should always be False for us.

## Checklist

**Required:**

**If applicable:**


fneum (approved):
I trust you on that :)

PR #2153, opened 2026-04-17
feat: restructure and improve transmission topology + config for all carriers

### LINKED ISSUE #2151: Restructure and improve pipeline candidate topology (config + methodology)
**Motivation:** I am working on PR that completely changes the way how the candidates for future pipelines are created (first and foremost H2 and CO2, but not limited to the two carriers), as they are currently centered around the AC grid topology (which might have served as a good heuristic/assumption, however does not always reflect best/shortest distance graph-theoretical candidates). Instead graph-theoretical proven methods like Delaunay triangulation and optional Gabriel graph-filtering (subset of Delaunay) yield more "intuitive" future candidates, see comparison:

Current implementation at high local NUTS3 resolution
Image

Delaunay:

Image

For this, I have already moved the pipeline topology construction from `prepare_sector_network`. I would propose to move it into a new `build_carrier_corridors` rule where the user can decide `delaunay` vs `acgrid` (legacy).

Now this made me think whether it makes sense to collect all pipeline related settings in a common spot in the user config, so at one glance we see all enabled or disabled pipelines (or other modes of transport), see draft/example below.

A draft config for this could look like this instead:
```
sectors:
  corridors (or pipelines or other name):
    H2:
      enable: true
        mode: delaunay # delaunay or acgrid
        gabriel_filter: true
        minimum_degree: 3
    CO2:
      enable: true
      mode: delaunay
      delaunay:
        gabriel_filter: true
        minimum_degree: 3
```

- This would be a breaking change and affect potentially the transport infrastructure/links of all carriers, including: H2, CO2, gas, 
- What would be a good key? "corridors", "pipelines"? My thought process behind this was that it is not always pipelines. Or (another use case) potentially in the future, more modes of transport are added.
For gas and the interplay with H2 it might get a bit more complicated already, open for suggestions here.
- Given your previous/current work on network expansion research, what are your thoughts?
@fneum @FabianHofmann @brynpickering @toniseibold @p-glaum @nworbmot @euronion 

Further feedback from others is also very welcome :)

bobbyxng:
Closes https://github.com/PyPSA/pypsa-eur/issues/2151

- Added a dedicated transmission candidate build rule with `carrier` wildcard support
- Implemented graph-theoretical topology for investment candidates in new `build_transmission.py`
  - Delaunay triangulation for candidate edge generation
  - Optional Gabriel filtering
  - Optional minimum degree backfilling, to mitigate under-connectedness of nodes.
  - Minimum topology when minimum degree is set to 1 and Gabriel filtering is enabled. Maximum topology is determined by Gabriel filtering disabled (Delaunay graph is the maximum topology of candidates)
- Carrier-specific transmission settings in config and updated schema
- Added new transmission docs section in `configuration.rst`
- Note that currently, the PR scopes the candidate generation for hydrogen and carbon_dioxide. Will be updated to move all old keys into the new structure

## Motivation
- Current pipeline candidate generation around AC topology can miss intuitive shortest-path graph candidates.
- Following discussion in https://github.com/PyPSA/pypsa-eur/issues/2151, "legacy" topology built on AC grid topology is dropped, switching to new approach entirely
- In the long run, brownfield/planned projects can be added for links of all carrier types (not only AC/DC) using this updated structure

## Checklist

**Required:**
- [X] Changes are tested locally and behave as expected.
- [X] Code and workflow changes are documented.

**If applicable:**
- [X] Changes in configuration options are reflected in `scripts/lib/validation`.
- [X] New rules are documented in the appropriate `doc/*.rst` files.


bobbyxng:
Both new methods yield roughly the same ballpark wrt. number of edges in a 200-node network:
image

Applying the Gabriel filter returns the minimum number of edges obtained through this method. The minimum degree of nodes is set to 1. By changing the degree, the topology can be increased again. So we are always moving in the range `n_edges_gabriel` <= `n` <= `n_edges_delaunay` @fneum 

If you wanna get rid of "stubs", you can simply set the minimum degree to 2 or higher

bobbyxng:
I consolidated old and new keys into the setup below and updated the associated doc entries as well as config examples (please double-check) @fneum 

I was not sure how to handle these settings, which are partly related to the transmission infrastructure in `sector`, including
- `electricity_distribution_grid: true`
- `electricity_distribution_grid_cost_factor: 1.0`
- `electricity_grid_connection: true`
- `transmission_efficiency`
- `H2_retrofit`
- `H2_retrofit_capacity_per_CH4`
- `gas_network_connectivity_upgrade`
- `gas_distribution_grid`
- `gas_distribution_grid_cost_factor: 1.0`

Maybe some of those keys can also be moved to the respective transmission carrier, happy to receive your feedback here @fneum 

```
# docs in https://pypsa-eur.readthedocs.io/en/latest/configuration.html#transmission
transmission:
  electricity:
    enable: true
    base_network: osm
    transmission_limit: vopt
    lines:
      types:
        63.0: "94-AL1/15-ST1A 20.0"
        66.0: "94-AL1/15-ST1A 20.0"
        90.0: "184-AL1/30-ST1A 110.0"
        110.0: "184-AL1/30-ST1A 110.0"
        132.0: "243-AL1/39-ST1A 110.0"
        150.0: "243-AL1/39-ST1A 110.0"
        220.0: "Al/St 240/40 2-bundle 220.0"
        300.0: "Al/St 240/40 3-bundle 300.0"
        330.0: "Al/St 240/40 3-bundle 300.0"
        380.0: "Al/St 240/40 4-bundle 380.0"
        400.0: "Al/St 240/40 4-bundle 380.0"
        500.0: "Al/St 240/40 4-bundle 380.0"
        750.0: "Al/St 560/50 4-bundle 750.0"
      s_max_pu: 0.7
      s_nom_max: .inf
      max_extension: 20000
      length_factor: 1.25
      reconnect_crimea: true
      under_construction: keep
      dynamic_line_rating:
        activate: false
        cutout: default
        correction_factor: 0.95
        max_voltage_difference: false
        max_line_rating: false
    links:
      p_max_pu: 1.0
      p_min_pu: -1.0
      p_nom_max: .inf
      max_extension: 30000
      length_factor: 1.25
      under_construction: keep
    transformers:
      x: 0.1
      s_nom: 2000.0
      type: ""
    projects:
      enable: true
      include:
        tyndp2020: true
        nep: true
        manual: true
      skip:
      - upgraded_lines
      - upgraded_links
      status:
      - under_construction
      - in_permitting
      - confirmed
      new_link_capacity: zero
  hydrogen:
    enable: true
    gabriel_filter:
      enable: true
      min_degree: 1
    length_factor: 1.25
    cost_factor: 1
  carbon_dioxide:
    enable: true
    gabriel_filter:
      enable: true
      min_degree: 1
    length_factor: 1.25
    cost_factor: 1
  methane_gas:
    enable: true
```

bobbyxng:
~~Also renamed "gas" to "methane_gas" for now, open to suggestions, technically we should probably call it `methane` to be in alignment with the other carriers~~ (Edit: renamed it back to `gas`)

rules/build_sector.smk
```diff
@@ -146,6 +146,48 @@ rule cluster_gas_network:
...
+        offshore_shapes=resources("offshore_shapes.geojson"),
+    output:
+        all_edges=resources("transmission/{carrier}_all_edges_{clusters}.geojson"),
+        candidates=resources("transmission/{carrier}_candidates_{clusters}.geojson"),
+    log:
+        logs("build_transmission_topology_{carrier}_{clusters}.log"),
+    benchmark:
+        benchmarks("build_transmission_topology/{carrier}_{clusters}")
+    wildcard_constraints:
+        carrier="hydrogen|carbon_dioxide",
```
brynpickering: Also `methane_gas`?

rules/build_sector.smk
```diff
@@ -146,6 +146,48 @@ rule cluster_gas_network:
         scripts("cluster_gas_network.py")
 
 
+rule build_transmission_topology:
```
brynpickering: Since the only difference between carriers is the suffix/prefix applied in certain data columns, I would suggest creating a delaunay graph per set of common settings first, then apply those prefixes/suffixes separately. That is, if all non-electricity transmission networks define delaunay with Gabriel filter of minimum degree 1, only one graph should be created across all three networks, e.g. `resources("transmission/delaunay_gabriel_min_1.geojson")`.

bobbyxng: Good idea, I have implemented the candidate topology generation based on degree only. The carriers and suffixes are added later in `prepare_sector_network`.

rules/build_sector.smk
bobbyxng: After talking to @fneum renamed it back to `gas`
Also following this change https://github.com/PyPSA/pypsa-eur/pull/2153#discussion_r3131263280 the wildcards are not referring to `carrier` anymore, but to the `min_degree`

bobbyxng:
Added a key that allows to set a maximum offshore (haversine) distance, similar to min_degree. This gives a bit more fine-grained control for offshore connections that the user deems unreasonable upfront

Here's an example with an arbitrary maximum haversine distance of 300 km offshore
image

bobbyxng:
I believe I finalised everything now. All transmission related settings are consolidated in the new `transmission` key.
Note that I deliberately left these keys in `sector` and added a `_cost` suffix, as they do not actually add the `_distribution_grid` but only capital costs to generators.

```
sector:
  electricity_grid_connection_cost: true
  gas_distribution_grid_cost: true
  gas_distribution_grid_cost_factor: 1.0
```

New setup:
```
# docs in https://pypsa-eur.readthedocs.io/en/latest/configuration.html#transmission
transmission:
  electricity:
    enable: true
    base_network: osm
    transmission_limit: vopt
    lines:
      types:
        63.0: "94-AL1/15-ST1A 20.0"
        66.0: "94-AL1/15-ST1A 20.0"
        90.0: "184-AL1/30-ST1A 110.0"
        110.0: "184-AL1/30-ST1A 110.0"
        132.0: "243-AL1/39-ST1A 110.0"
        150.0: "243-AL1/39-ST1A 110.0"
        220.0: "Al/St 240/40 2-bundle 220.0"
        300.0: "Al/St 240/40 3-bundle 300.0"
        330.0: "Al/St 240/40 3-bundle 300.0"
        380.0: "Al/St 240/40 4-bundle 380.0"
        400.0: "Al/St 240/40 4-bundle 380.0"
        500.0: "Al/St 240/40 4-bundle 380.0"
        750.0: "Al/St 560/50 4-bundle 750.0"
      s_max_pu: 0.7
      s_nom_max: .inf
      max_extension: 20000
      length_factor: 1.25
      reconnect_crimea: true
      under_construction: keep
      dynamic_line_rating:
        activate: false
        cutout: default
        correction_factor: 0.95
        max_voltage_difference: false
        max_line_rating: false
    links:
      p_max_pu: 1.0
      p_min_pu: -1.0
      p_nom_max: .inf
      max_extension: 30000
      length_factor: 1.25
      under_construction: keep
      efficiency:
        enable: true
        efficiency_static: 0.98
        efficiency_per_1000km: 0.977
    transformers:
      x: 0.1
      s_nom: 2000.0
      type: ""
    projects:
      enable: true
      include:
        tyndp2020: true
        nep: true
        manual: true
      skip:
      - upgraded_lines
      - upgraded_links
      status:
      - under_construction
      - in_permitting
      - confirmed
      new_link_capacity: zero
  electricity_distribution:
    enable: true
    cost_factor: 1.0
    efficiency:
      enable: true
      efficiency_static: 0.97
  hydrogen:
    enable: true
    gabriel_filter:
      enable: true
      min_degree: 1
    max_offshore_haversine_distance: .inf
    length_factor: 1.25
    cost_factor: 1
    retrofit:
      enable: false
      capacity_per_ch4: 0.6
    efficiency:
      enable: true
      efficiency_per_1000km: 1.0
      compression_per_1000km: 0.018
  carbon_dioxide:
    enable: true
    gabriel_filter:
      enable: true
      min_degree: 1
    max_offshore_haversine_distance: .inf
    length_factor: 1.25
    cost_factor: 1
  gas:
    enable: true
    connectivity_upgrade: 1
    efficiency:
      enable: true
      efficiency_per_1000km: 1.0
      compression_per_1000km: 0.01
```

bobbyxng:
Ready for (final) review @fneum @brynpickering

rules/build_sector.smk
```diff
@@ -146,6 +148,63 @@ rule cluster_gas_network:
...
+        if not gabriel_filter.get("enable", False):
+            continue
+        if "min_degree" not in gabriel_filter:
+            continue
+
+        min_degree = int(gabriel_filter["min_degree"])
+        max_offdist = f"{float(carrier_config.get('max_offshore_haversine_distance', float('inf'))):g}"
+        specs.add((min_degree, max_offdist))
+
+    return sorted(specs, key=lambda spec: (spec[0], float(spec[1])))
```
brynpickering: I don't think this should be needed. If `min_degree` and `max_offdist` are left as wildcards in `build_transmission_topology` then when you define the inputs to `build_sector_network` in `input_transmission_candidates` it will trigger the necessary calls to `build_transmission_topology`. 

Then, in `build_transmission_topology` you just build one topology, based on the requested input.

rules/build_sector.smk
```diff
@@ -146,6 +148,63 @@ rule cluster_gas_network:
...
+            {
+                "min_degree": int(min_degree),
+                "max_offshore_haversine_distance_km": float(max_offdist),
+            }
+            for min_degree, max_offdist in transmission_candidate_specs()
+        ],
+    message:
+        "Building transmission candidates for {wildcards.clusters} clusters"
+    script:
+        scripts("build_transmission_topology.py")
```
brynpickering, line 192: Consider splitting into two (and splitting the script) to first build the full network and then filter it. This separation of concerns should make the respective scripts a bit simpler and means the full set of edges is built only once (minimising overall resource requirements)

```suggestion
rule build_transmission_topology:
    input:
        network=resources("networks/base_s_{clusters}.nc"),
        offshore_shapes=resources("offshore_shapes.geojson"),
    output:
        edges=resources("transmission/candidates_{clusters}_all_edges.geojson"),
    log:
        logs("build_transmission_topology_all_edges_{clusters}.log"),
    benchmark:
        benchmarks("build_transmission_topology_all_edges/{clusters}")
    resources:
        mem_mb=4000,
    params:
        transmission=config_provider("transmission"),
    message:
        "Building transmission candidates for {wildcards.clusters} clusters using all inter-node edges"
    script:
        scripts("build_transmission_topology.py")
        
rule filter_transmission_topology:
    input:
        edges=resources("transmission/candidates_{clusters}_all_edges.geojson"),
    output:
        filtered_edges=resources(
              "transmission/candidates_{clusters}_min_{min_degree}_maxoffdist_{max_offdist}_km.geojson"
            )
    log:
        logs("filter_transmission_topology_{clusters}_{min_degree}_{max_offdist}.log"),
    benchmark:
        benchmarks("filter_transmission_topology/{clusters}_{min_degree}_{max_offdist}")
    resources:
        mem_mb=4000,
    params:
        transmission=config_provider("transmission"),
        candidate_specs=lambda wildcards: {
                "min_degree": int(wildcards.min_degree),
                "max_offshore_haversine_distance_km": float(wildcards.max_offdist),
        },
    message:
        "Building filtered set of transmission candidates for {wildcards.clusters} clusters using Delaunay triangulation with {wildcards.min_degree} minimum node degree target and {wildcards.max_offdist}km maximum offshore haversine distance"
    script:
        scripts("filter_transmission_topology.py")
```

scripts/lib/validation/config/transmission.py
```diff
@@ -0,0 +1,388 @@
...
+    )
+    gabriel_filter: _GabrielFilterConfig = Field(
+        default_factory=_GabrielFilterConfig,
+        description="Gabriel filter configuration.",
+    )
+    max_offshore_haversine_distance: float = Field(
+        float("inf"),
+        gt=0,
+        description="Maximum haversine distance in km for offshore transmission candidates. Candidate edges a total offshore length exceeding this threshold are excluded. Defaults to infinity (no limit).",
+    )
```
brynpickering, line 218: move to within `gabriel_filter`, perhaps by renaming to `filter_edges` and having `gabriel_filter_min_degree` and `max_offshore_haversine_distance` within it (since the max_offshore distance is currently only being triggered when the gabriel filter is also active).

rules/build_sector.smk
```diff
@@ -1562,10 +1621,32 @@ def input_heat_source_power(w):
...
+        key = f"{carrier}_transmission_candidates"
+        if gabriel_cfg.get("enable", False) and "min_degree" in gabriel_cfg:
+            min_degree = int(gabriel_cfg["min_degree"])
+            max_offdist = f"{float(carrier_cfg.get('max_offshore_haversine_distance', float('inf'))):g}"
+            candidates[key] = resources(
+                f"transmission/candidates_{{clusters}}_min_{min_degree}_maxoffdist_{max_offdist}_km.geojson"
+            )
+        else:
+            candidates[key] = resources("transmission/all_edges_{clusters}.geojson")
+    return candidates
```
brynpickering: This is overly cautious given what is being enforced in by the config validation. Transmission must be a dict with all carrier keys defined. min_degree and max_offshore will also be int/float respectively as that is forced by the validator.

```suggestion
    for carrier, carrier_cfg in config["transmission"].items():
        gabriel_cfg = carrier_cfg.get("gabriel_filter")
        if ("gabriel_filter" := carrier_cfg.get("gabriel_filter")) is None or not carrier_cfg["enable"]:
            continue
        key = f"{carrier}_transmission_candidates"
        if gabriel_cfg["enable"]:
            min_degree = gabriel_cfg["min_degree"]
            max_offdist = carrier_cfg["max_offshore_haversine_distance"]
            candidates[key] = resources(
                f"transmission/candidates_{{clusters}}_min_{min_degree}_maxoffdist_{max_offdist}_km.geojson"
            )
        else:
            candidates[key] = resources("transmission/all_edges_{clusters}.geojson")
    return candidates
```

scripts/build_transmission_topology.py
```diff
@@ -0,0 +1,583 @@
...
+            max_offshore_haversine_distance_km,
+        )
+        selected_edges = prepare_candidate_edges(
+            delaunay_graph=delaunay_graph,
+            gabriel_filter_enabled=gabriel_filter_enabled,
+            node_count=len(coords_geo),
+            min_degree=min_degree,
+            max_offshore_haversine_distance_km=max_offshore_haversine_distance_km,
+        )[COLS_EDGES]
+        selected_edges.to_file(output_path)
```
brynpickering: Spin out into its own script with the `all_edges` as the input. Only thing you'll miss is `node_count` but I expect that can be inferred from `bus0` and `bus1` in the delaunay graph gdf

scripts/prepare_sector_network.py
```diff
@@ -2056,39 +2078,53 @@ def add_h2_gas_infrastructure(
...
             p_nom_extendable=True,
-            length=h2_pipes.length.values,
+            length=(
+                h2_pipes["length"].values
+                * cf_transmission["hydrogen"].get("length_factor", 1.0)
+            ),
             capital_cost=costs.at["H2 (g) pipeline", "capital_cost"]
-            * h2_pipes.length.values,
+            * h2_pipes["length"].values
+            * cf_transmission["hydrogen"].get("length_factor", 1.0)
```
brynpickering:
```suggestion
                * cf_transmission["hydrogen"]["length_factor"]
            ),
capital_cost=costs.at["H2 (g) pipeline", "capital_cost"]
            * h2_pipes["length"].values
            * cf_transmission["hydrogen"]["length_factor"]
```
Trust in the config validator to make keys available and to fill with default values where necessary. This will ensure defaults are only ever defined once (in the validator), mitigating the possibility of them going out of sync in future

brynpickering: There are other cases of this in this PR (uses of `.get`) that I haven't left explicit comments on but that can also be updated. Rule of thumb is that if the validator creates entry in the default config YAML, it's safe to assume that the key will be there and will have a default value (even if that's an empty dict/list/whatever).

scripts/simplify_network.py
```diff
@@ -432,7 +432,9 @@ def remove_converters(n: pypsa.Network) -> pypsa.Network:
     Nyears = n.snapshot_weightings.objective.sum() / 8760
     buses_prev, lines_prev, links_prev = len(n.buses), len(n.lines), len(n.links)
 
-    linetype_380 = snakemake.config["lines"]["types"][380]
+    linetype_380 = snakemake.config["transmission"]["electricity"]["lines"]["types"][
```
brynpickering, line 435: This is the opposite of my earlier comment: there's no requirement from the validator for the user to define the 380kV line type. They could replace the line types dict and not include this level. This call will then fail. Since it's a critical level, I would add a field validator to the validator to ensure the `380` key is always defined (perhaps beyond the scope of this PR, open an issue if so).

rules/build_sector.smk
brynpickering: btw, I don't like that the existence of the `gabriel_filter` key implicitly defines that the network will be defined by delaunay triangulation. That is, there's nothing explicit in the config to suggest how the edges will be generated, except for electricity which has `base_network`. Could we perhaps have `base_network` as a config option for every transmission carrier, then limit it to `delaunay` for hydrogen and co2, `osm`(?) for gas. Would leave open the option to extend the base network options in future (e.g. using only existing gas network for h2)

scripts/build_transmission_topology.py
```diff
@@ -0,0 +1,583 @@
...
+    "bus0",
+    "bus1",
+    "length",
+    "gabriel_edge",
+    "geometry",
+    "underwater_fraction",
+]
+
+
+def collect_transmission_topology_settings(
```
brynpickering: This method looks to be unused

scripts/lib/validation/config/transmission.py
brynpickering: Setting `gabriel_filter_min_degree` default to zero will effectively switch it off, so if a user wants to activate it, they can set it to >= 1. Then you have the `enable` key which defines whether filtering should be attempted and then two keys that default to effectively not being applied unless a user changes the values (< inf for haversine, > 0 for gabriel). 


This is arguably overkill, you could even avoid the `enable` key since they'd default to not being applied anyway, but this approach leaves open the option to add further filtering methods in future and switch them on/off as a group as desired.

scripts/build_transmission_topology.py
```diff
@@ -0,0 +1,583 @@
...
+            int(unreachable_mask.sum()),
+            node_count,
+            int(unmet_mask.sum()),
+            node_count,
+        )
+
+    return result
+
+
+def prepare_candidate_edges(
```
brynpickering: make this method specific to applying the haversine distance, then just use `enforce_min_degree` to handle the gabriel filter. No need to pass the `gabriel_filter_enabled` option if my config suggestion is used, since a gabriel filter min_degree of zero is already handled by `enforce_min_degree` as effectively switching off the filtering.

brynpickering (changes requested):
Thanks for the updates @bobbyxng . As you might notice, I'm not really reviewing your implementation. I trust that it works as you expect it to. Instead, I'm concerned with maintainability and user-facing (i.e. config). @fneum might be more qualified to actually dissect the delaunay and filtering methods

scripts/lib/validation/config/transmission.py
bobbyxng: Good point, I have also changed the config now so the user can choose to limit the candidates based on `max_offshore_haversine_distance` without using the gabriel filter. 

Example
```
  carbon_dioxide:
    enable: true
    gabriel_filter_min_degree: 0
    max_offshore_haversine_distance: 700 # km
```

rules/build_sector.smk
bobbyxng: Implemented the first comment.
Regarding the second, had similar thoughts on this, however, without a clear conclusion yet on this:
- @fneum suggested to drop the "legacy" option to generate alternative topologies for H2 and CO2, i.e., from the electricity high voltage grid (there is no really good reason keeping the old approach)
- If we were to include `base_network`, we could either limit the options to `delaunay` for `transmission.hydrogen.base_network` and `transmission.carbon_dioxide.base_network` or also allow `electricity_topology` if we still want to enable the original behaviour of the current master branch
- Regarding using only existing gas network for H2, there is some overlap with using transmission.hydrogen.retrofit, which does not fully mean the same thing as what you are suggesting, might introduce a bit of confusion though.
- I generally like the idea to have `base_network` for each transmission carrier, even if for now, the options would be limited.

scripts/prepare_sector_network.py
bobbyxng: Fair point! Updated everywhere :)

scripts/build_transmission_topology.py
bobbyxng: removed outdated artifact

rules/build_sector.smk
brynpickering: > I generally like the idea to have base_network for each transmission carrier, even if for now, the options would be limited.

Yeah, I'd also be fine with it even if it just has `delaunay` for now (or also the `electricity_topology` for backwards-compatibility, but then you need the logic to pull that topology when requested, so maybe not worth it).

brynpickering: > Regarding using only existing gas network for H2, there is some overlap with using transmission.hydrogen.retrofit, which does not fully mean the same thing as what you are suggesting, might introduce a bit of confusion though.

Yeah, true. It would be neat to be able to define `base_network: gas_topology` and then `retrofit` could be used to define whether it can use the existing `base_network` or has to build from scratch following the same topology. Maybe not something for now.

scripts/simplify_network.py
fneum: Good point, fine with outsourcing this to separate PR. xref #2166

config/examples/config.distribution-grid-experimental.yaml
```diff
@@ -45,10 +45,13 @@ countries:
 
 electricity:
   voltages: [63., 66., 90., 110., 132., 150., 220., 300., 330., 380., 400., 500., 750.]
```
fneum, line 47: Shouldn't the voltages also move into the `transmission.electricity` block?

scripts/build_electricity_transmission_projects.py
```diff
@@ -18,11 +18,11 @@
 Outputs
 -------
 
-- ``transmission_projects/new_lines.csv``: New project lines to be added to the network. This includes new lines and upgraded lines.
-- ``transmission_projects/new_links.csv``: New project links to be added to the network. This includes new links and upgraded links.
-- ``transmission_projects/adjust_lines.csv``: For lines which are upgraded, the decommissioning year of the existing line is adjusted to the build year of the upgraded line.
-- ``transmission_projects/adjust_links.csv``: For links which are upgraded, the decommissioning year of the existing link is adjusted to the build year of the upgraded link.
-- ``transmission_projects/new_buses.csv``: For some links, we have to add new buses (e.g. North Sea Wind Power Hub).
+- ``transmission/electricity_projects/new_lines.csv``: New project lines to be added to the network. This includes new lines and upgraded lines.
```
fneum, line 21: Is this change necessary?

scripts/base_network.py
```diff
@@ -302,12 +302,12 @@ def _reconnect_crimea(lines):
 
 def _set_electrical_parameters_lines_eg(lines, config):
     v_noms = config["electricity"]["voltages"]
-    linetypes = config["lines"]["types"]
+    linetypes = config["transmission"]["electricity"]["lines"]["types"]
```
fneum, line 305: There's some optimisation potential to avoid repeated `config["transmission"]["electricity"]["lines"]` accesses. Could store once in some variable.

scripts/_helpers.py
```diff
@@ -821,8 +821,12 @@ def update_config_from_wildcards(config, w, inplace=True):
 
         _, maxext = find_opt(opts, "linemaxext")
         if maxext is not None:
-            config["lines"]["max_extension"] = maxext * 1e3
-            config["links"]["max_extension"] = maxext * 1e3
+            config["transmission"]["electricity"]["lines"]["max_extension"] = (
```
fneum, line 824: could start with `cf_lines = config["transmission"]["electricity"]["lines"]`.

scripts/lib/validation/config/_schema.py
```diff
@@ -132,6 +127,10 @@ class ConfigSchema(BaseModel):
         default_factory=ElectricityConfig,
         description="Electricity sector configuration.",
     )
+    transmission: TransmissionConfig = Field(
+        default_factory=TransmissionConfig,
+        description="Transmission candidate configuration.",
```
fneum, line 132:
```suggestion
        description="Transmission configuration.",
```

rules/build_sector.smk
```diff
@@ -2,6 +2,8 @@
 #
 # SPDX-License-Identifier: MIT
 
+import math
```
fneum, line 5: Not needed, is it?

rules/build_sector.smk
```diff
@@ -146,6 +148,50 @@ rule cluster_gas_network:
...
+        ),
+    benchmark:
+        benchmarks(
+            "build_transmission_topology/{clusters}_min_{min_degree}_maxoffdist_{max_offdist}"
+        )
+    resources:
+        mem_mb=2000,
+    params:
+        min_degree=lambda w: int(w.min_degree),
+        max_offdist=lambda w: float(w.max_offdist),
```
fneum, line 188: Why as params? Can directly use `snakemake.wildcards.min_degree` in script.

doc/release_notes.rst
```diff
@@ -22,6 +22,26 @@ Release Notes
 
 * feat: Improve the config validation to cover scenario management (https://github.com/PyPSA/pypsa-eur/pull/2155).
 
+* feat: Add new rule :mod:`build_transmission_topology` to generate greenfield pipeline investment candidates
+  (H2, CO2) using Delaunay triangulation with optional Gabriel graph filtering and minimum-degree backfilling,
+  replacing the former AC-topology-based approach. A ``max_offshore_haversine_distance`` setting per carrier
+  limits offshore corridor lengths (https://github.com/PyPSA/pypsa-eur/pull/2153).
+
+* refactor: Rename ``transmission_projects`` to ``transmission.electricity.projects`` and
```
fneum, line 30: This also falls into category breaking.

doc/release_notes.rst
```diff
@@ -22,6 +22,26 @@ Release Notes
 
 * feat: Improve the config validation to cover scenario management (https://github.com/PyPSA/pypsa-eur/pull/2155).
 
+* feat: Add new rule :mod:`build_transmission_topology` to generate greenfield pipeline investment candidates
```
fneum, line 25: This is not fully up to date anymore, is it?

doc/release_notes.rst
```diff
@@ -22,6 +22,26 @@ Release Notes
...
+  limits offshore corridor lengths (https://github.com/PyPSA/pypsa-eur/pull/2153).
+
+* refactor: Rename ``transmission_projects`` to ``transmission.electricity.projects`` and
+  ``gas_distribution_grid`` to ``gas_distribution_grid_cost`` to better reflect their semantics
+  (https://github.com/PyPSA/pypsa-eur/pull/2153). Rename `sector.electricity_grid_connection` to `sector.electricity_grid_connection_cost`
+  and `gas_distribution_grid` to `gas_distribution_grid_cost`.
+
+**Breaking changes**
+
+* Consolidate all carrier-specific transmission settings under a new top-level ``transmission`` config
```
fneum, line 37: This is quite a substantial refactor of the configuration schema. Would a migration script to update old configs make sense? Maybe @lkstrp has an opinion as well.

scripts/build_transmission_topology.py
```diff
@@ -0,0 +1,236 @@
...
+            delaunay_graph["length"] * delaunay_graph["underwater_fraction"]
+        )
+        return delaunay_graph[offshore_length <= max_offdist].copy()
+    return delaunay_graph.copy()
+
+
+if __name__ == "__main__":
+    is_mock_run = "snakemake" not in globals()
+
+    if is_mock_run:
```
fneum, line 196:
```suggestion
    if  "snakemake" not in globals():
```
why not just?

scripts/build_transmission_topology.py
```diff
@@ -0,0 +1,236 @@
...
+            min_degree=1,
+            max_offdist="inf",
+            configfiles=["config/config.200.yaml"],
+        )
+
+    configure_logging(snakemake)
+    set_scenario_config(snakemake)
+
+    min_degree = snakemake.params["min_degree"]
+    max_offdist = snakemake.params["max_offdist"]
```
fneum, line 211:
```suggestion
    min_degree = snakemake.wildcards.min_degree
    max_offdist = snakemake.wildcards.max_offdist
```

scripts/build_transmission_topology.py
```diff
@@ -0,0 +1,236 @@
...
+    "bus0",
+    "bus1",
+    "length",
+    "gabriel_edge",
+    "geometry",
+    "underwater_fraction",
+]
+
+
+def enforce_min_degree(
```
fneum, line 50: Just to double-check: There's no `networkx` implementation of this that you could reuse rather than implementing yourself?

scripts/lib/validation/config/transmission.py
```diff
@@ -0,0 +1,375 @@
+# SPDX-FileCopyrightText: Contributors to PyPSA-Eur <https://github.com/pypsa/pypsa-eur>
+#
+# SPDX-License-Identifier: MIT
+
+"""
+Transmission candidate configuration.
```
fneum, line 6:
```suggestion
Transmission configuration.
```
It's not just candidates.

scripts/prepare_sector_network.py
```diff
@@ -902,32 +914,43 @@ def add_co2_network(n, costs, co2_network_cost_factor=1.0):
     created using the create_network_topology helper function.
     """
     logger.info("Adding CO2 network.")
-    co2_links = create_network_topology(n, "CO2 pipeline ")
+    co2_links = gpd.read_file(carbon_dioxide_transmission_candidates).copy()
+
+    if "name" not in co2_links.columns:
+        co2_links["name"] = co2_links["bus0"] + " -> " + co2_links["bus1"]
```
fneum, line 920: Since you create this file, don't you already know if "name" is in columns?

scripts/prepare_sector_network.py
```diff
@@ -902,32 +914,43 @@ def add_co2_network(n, costs, co2_network_cost_factor=1.0):
...
+
+    if not co2_links.empty and not co2_links.iloc[0]["bus0"].endswith(" co2 stored"):
+        co2_links[["bus0", "bus1"]] = co2_links[["bus0", "bus1"]] + " co2 stored"
```
fneum, line 923: Since you create this file, don't you already know if `co2 stored" is ending of bus names?

scripts/prepare_sector_network.py
```diff
@@ -2056,39 +2078,52 @@ def add_h2_gas_infrastructure(
...
-            n, "H2 pipeline ", carriers=["DC", "gas pipeline"]
-        )
-        h2_buses_loc = n.buses.query("carrier == 'H2'").location  # noqa: F841
-        h2_pipes = h2_pipes.query("bus0 in @h2_buses_loc and bus1 in @h2_buses_loc")
+        h2_pipes = gpd.read_file(hydrogen_transmission_candidates).copy()
+        if "name" not in h2_pipes.columns:
+            h2_pipes["name"] = h2_pipes["bus0"] + " -> " + h2_pipes["bus1"]
+
+        if not h2_pipes.empty and not h2_pipes.iloc[0]["bus0"].endswith(" H2"):
+            h2_pipes[["bus0", "bus1"]] = h2_pipes[["bus0", "bus1"]] + " H2"
```
fneum, line 2103: See above. I think these checks are not necessary.

scripts/prepare_sector_network.py
```diff
@@ -6566,15 +6624,25 @@ def add_import_options(
...
+        ],
+    }
 
-    for k, v in options["transmission_efficiency"].items():
-        if k in options["transmission_efficiency"]["enable"]:
-            lossy_bidirectional_links(n, k, v)
+    for carrier, efficiency_config in transmission_efficiency_map.items():
+        if efficiency_config["enable"]:
+            losses = {k: v for k, v in efficiency_config.items() if k != "enable"}
+            lossy_bidirectional_links(n, carrier, losses)
```
fneum, line 6645: Did you double-check this worked correctly?

fneum (changes requested):
Looks great overall!

PR #2158, opened 2026-04-30
Set IISMethod for Gurobi to 1

euronion:
Small PR to set the default for the [Gurobi IIS Method ](https://docs.gurobi.com/projects/optimizer/en/current/reference/parameters.html#iismethod) to 1. Together with @coroa I observed this method to converge much faster and more reliable than the default `-1` option from Gurobi.

Does not affect regular solving, only the IIS in case of infeasibilities.

## Checklist

**Required:**
- [n/a] A release note entry is added to `doc/release_notes.rst`.

**If applicable:**
- [n/a] For new data sources or versions, [these instructions](https://pypsa-eur.readthedocs.io/en/latest/data_sources.html) have been followed.
- [n/a] New rules are documented in the appropriate `doc/*.rst` files.

### DIFF
--- a/config/config.default.yaml
+++ b/config/config.default.yaml
@@ -1216,6 +1216,7 @@ solving:
       AggFill: 0
       PreDual: 0
       GURO_PAR_BARDENSETHRESH: 200
+      IISMethod: 1
     "gurobi-numeric-focus":
       NumericFocus: 3
       method: 2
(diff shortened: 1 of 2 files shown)


fneum (approved):
> To compute an IIS for an LP, it is sufficient to solve an LP with dimensions similar to the dual of the original model. If the solve time for that LP is excessive, setting the IISMethod parameter to 1 may offer a faster alternative;

https://docs.gurobi.com/projects/optimizer/en/current/reference/parameters.html

Agreed

PR #2162, opened 2026-05-07
docs: migrate to markdown/ mkdocs/ material

### LINKED ISSUE #2059: Update docs to markdown + mkdocs
ReStructuredText is harder to write and maintain than Markdown and mkdocs is effectively at feature parity with Sphinx (if not further ahead as there are some great, actively maintained extension libraries like [Material](https://squidfunk.github.io/mkdocs-material/)).

This would also align this project with PyPSA/PyPSA, which makes cross-repo maintenance a bit easier.

lkstrp:
Closes #2059 

- I still went with Material and MkDocs (same setup as in PyPSA), but once [Zensical](https://squidfunk.github.io/mkdocs-material/blog/2025/11/05/zensical/) is stable we should move there. It currently does not support all features, but it will be close to a drop-in replacement for MkDocs. I cannot find the article right now, but the MkDocs and Material for MkDocs maintainers are not very aligned right now. Another migration will be much easier though, since we only need to touch plugins then. EDIT (https://fpgmaas.com/blog/collapse-of-mkdocs/ Thanks @euronion)
- All features previously using Sphinx are moved over. For the yaml stuff I had to write a macro, and all DAGs are now read from files instead of being created during the build.
- Should we also move this to a new URL? Either `docs.pypsa.org/pypsa-eur` or its own subdomain?
- To review this, the commit after the rename should be checked. There we still get a proper diff: https://github.com/PyPSA/pypsa-eur/commit/5634986115d9428e221d15e0664f5cb1abdf2db7
- If downstream repos changed the documentation, this will be a bit of a nightmare to pull in. It will be easiest to just apply the full downstream diff again on top of here. I think I saw separate docs for e.g. TYNDP @daniel-rdt @tgilon
- @fneum I will do a couple more changes and also go through the full diff again, but you are already free to push whatever on here.

Run this locally with `pixi run update-dags` & `pixi run -e doc mkdocs serve`
Find the build [here](https://pypsa-eur--2162.org.readthedocs.build/en/2162/)

**This should again not be squashed. We lose the git history otherwise.**

## Checklist
- [x] Another check of everything, to find all hallucinations I missed
- [x] Some Design Tweaks, haven't touched that yet.


daniel-rdt:
Thank you @lkstrp for the heads up!

doc/index.md
```diff
@@ -0,0 +1,170 @@
...
+    pages = "1066--1086",
+    year = "2022",
+    doi = "10.1016/j.joule.2022.04.016",
+    eprint = "2109.09563",
+}
+```
+
+If you want to cite a specific PyPSA-Eur version, each release of PyPSA-Eur is stored on Zenodo with a release-specific DOI:
+
+[![DOI](https://zenodo.org/badge/DOI/10.5281/zenodo.3520874.svg)](https://doi.org/10.5281/zenodo.3520874)
```
fneum, line 166: Previously was rendered as badge.

doc/tutorial.md
```diff
@@ -0,0 +1,274 @@
...
+{{ yaml_section(start_at="atlite:", end_before="renewable:", source="test/config.electricity.yaml") }}
+```
+
+We can also decide which weather data source should be used to calculate
+potentials and capacity factor time-series for each carrier. For example, we may
+want to use the ERA-5 dataset for solar and not the default SARAH-3 dataset.
+
+```yaml
+{{ yaml_section(start_at="solar:", end_at="cutout:", source="test/config.electricity.yaml") }}
+```
```
fneum, line 73: This one did not render correctly.

fneum: Also `enable:` section in configuration.

fneum:
Looks great! I would merge that ASAP and then built improvements on top. Needless to say, the documentation of PyPSA-Eur is in a shocking state.

- **YAML blocks:** The YAML selection did not always work correctly (often to long, esp. with shared setting endings or duplicate keys like "enable"). *Must be changed before merge.*

- **Configuration:** There's a mistake in how indentation levels >= 2 are treated. They are shown on the same level as the first. The "renewable" section does not look right (duplication and notes inception). Some unresolved `literalinclude` and `jsonschema` RST syntax. *Must be changed before merge.*

- **Release Notes:** Can we get hyperlinks for PR numbers? *Should be changed before merge.*

- **Rules:** Some unresolved `.. code:: yaml` and `.. tip::` blocks. *Should be changed before merge.*

- **Data Sources:** Links in table are not clickable. *Can be changed before merge.*

- I have ideas for tabs Home, Design, Configuration, Rules, Contributing. To be tackled later.

Everything else, later -- I would say.

scripts/build_cutout.py
```diff
@@ -2,69 +2,44 @@
...
 
-- installed the `Copernicus Climate Data Store <https://cds.climate.copernicus.eu>`_ ``cdsapi`` package  (`install with `pip``) and
-- registered and setup your CDS API key as described `on their website <https://cds.climate.copernicus.eu/api-how-to>`_.
+- installed the [Copernicus Climate Data Store](https://cds.climate.copernicus.eu) `cdsapi` package  (`install with `pip``) and
+- registered and setup your CDS API key as described [on their website](https://cds.climate.copernicus.eu/api-how-to).
 
-.. seealso::
-    For details on the weather data read the `atlite documentation <https://atlite.readthedocs.io/en/latest/>`_.
-    If you need help specifically for creating cutouts `the corresponding section in the atlite documentation <https://atlite.readthedocs.io/en/latest/examples/create_cutout.html>`_ should be helpful.
+**See also:** For details on the weather data read the [atlite documentation](https://atlite.readthedocs.io/en/latest/). If you need help specifically for creating cutouts [the corresponding section in the atlite documentation](https://atlite.readthedocs.io/en/latest/examples/create_cutout.html) should be helpful.
```
brynpickering: Use an [info admonition](https://squidfunk.github.io/mkdocs-material/reference/admonitions/#+type:info) with "See also" as the title

doc/hooks.py
```diff
@@ -0,0 +1,47 @@
...
+#
+# SPDX-License-Identifier: MIT
+
+import os
+import sys
+from unittest.mock import MagicMock
+
+import requests
+
+MOCK_MODULES = [
```
brynpickering: I'm pretty sure you don't need to mock modules with mkdocs

lkstrp:
Thanks both! I think this is ready now

fneum:
Let's go! I'll let you do merge without squash to preserve git history.

Just one thing rendering of YAML at: https://pypsa-eur--2162.org.readthedocs.build/en/2162/configuration/#sector_cf

PR #2163, opened 2026-05-12
Update pixi to use relative exclude-newer

brynpickering:
Fixes the CD hack to inject the date 7 days prior by using recent functionality added to `pixi` to use relative `exclude-newer`.

It is also possible to override this on a per dependency basis, in case there is a vulnerability that would be fixed by using _the_ most recent release.

Requires `pixi` v0.67 or higher. Run `pixi self-update`.

## Checklist

**Required:**

**If applicable:**


lkstrp (approved):
Great !

PR #2164, opened 2026-05-14
Feat: SBOM vulnerability scan

brynpickering:
- Uploads pixi default deps to github dependency list
- Uploads identified vulnerabilities to codeQL vulnerabilities list
- Identifies vulnerabilities on updates to pixi.lock
- Move unnecessary deps from `default` to `dev` env (jupyter introduces `npm` `ip` package vulnerability [cvs-2024-29415](https://github.com/advisories/GHSA-2p57-rm9w-gvfp)).

As is ever the case with these kinds of PRs, I find it impossible to successfully test the workflows locally, even with `act` (docker and osx-arm64 aren't the best of friends). So, I don't know for sure that they work as expected...

What you get:

1. on successful merge of the lockfile updater PR into `master`, the repo dependencies are uploaded to the github API so they are correctly represented in the [dependency graph](https://github.com/PyPSA/pypsa-eur/network/dependencies).
1. on successful merge of the lockfile updater PR into `master`, the latest identified vulnerabilities are uploaded to the repo [CodeQL vulnerability list](https://github.com/PyPSA/pypsa-eur/security/quality). 
1. on any change to `pixi.lock` in a PR to `master`, a comment is added ~~with a table of vulnerabilities identified in the updated env~~ EDIT: now the comment is a _diff_ of the vulnerabilities compared to those present in `master`. 

> [!NOTE]
> PR comment doesn't work correctly in PRs from forks, as [noted by the action maintainer](https://github.com/peter-evans/create-or-update-comment/tree/v4/#action-inputs). We could update this to be triggered on [`pull_request_target`](https://docs.github.com/en/actions/reference/workflows-and-actions/events-that-trigger-workflows#pull_request_target) but there are other concerns then about vulnerabilities introduced by bad actors via the action. The current approach involves a slightly different check if the PR comes from a fork; it won't try to comment but will fail the check if there are any new vulnerabilities present compared to those on `master`.

This is all needed because native github tools (CodeQL & dependabot) don't analyse `pixi.lock` files.

**Required:**

**If applicable:**


.github/workflows/security-scan.yaml
```diff
@@ -0,0 +1,47 @@
+name: Security scan
+on:
+  pull_request:
+    types:
+    - closed
+    branches:
+    - master
+
```
lkstrp, line 12: I think this should run on every masters merge. Or even better with a daily cron job.

PR checks are done with the vulnerability report workflow below. It does not matter if they come from a PR or from the GitHub Actions update workflow.

If we only do security scans with every GitHub Actions update workflow, we would pin dependencies, and then not notice if they are marked as compromised afterwards. We would only notice with the next biweekly update cycle.

```suggestion
 on:
    push:
      branches: [master]
      paths: [pixi.lock]
    schedule:
      - cron: "0 6 * * *"
```

.github/workflows/security-scan.yaml
```diff
@@ -0,0 +1,47 @@
...
+jobs:
+  security-scan:
+    if: ${{ github.event.pull_request.merged == true && github.head_ref == 'update-locked-environment' }}
```
lkstrp:
```suggestion
```

.github/workflows/vulnerability-report.yaml
```diff
@@ -0,0 +1,97 @@
...
+    branches:
+    - master
+    paths:
+    - pixi.lock
+  workflow_dispatch:
+    inputs:
+      pr:
+        description: Pull Request number
+        required: true
+        type: number
```
lkstrp:
```suggestion
```

This will not work and currently still checks the master branch, but posts to PR

.github/workflows/vulnerability-report.yaml
```diff
@@ -0,0 +1,97 @@
...
+          Vulnerability scan of `.pixi/envs/default` based on the updated lockfile.
+
+          <details>
+          <summary>Full scan results</summary>
+
+          ```
+          ${{ steps.vuln-table.outputs.REPORT }}
+          ```
+
+          </details>
```
lkstrp:
```suggestion
```

I guess the action already handles passing '' as comment-id and is creating a new one instead

pixi.toml
lkstrp, line 64:
```suggestion
```
Also move to dev?

pixi.toml
```diff
@@ -63,7 +63,6 @@ gurobi = ">=12.0.3"
 highspy = ">=1.12.0"
 ipython = ">=9.7.0"
```
lkstrp, line 67:
```suggestion
```

Also move to dev?

lkstrp (changes requested):
Loooovely! Sorry I just merged the lock update before I saw this.

> We could update this to be triggered on [pull_request_target](https://docs.github.com/en/actions/reference/workflows-and-actions/events-that-trigger-workflows#pull_request_target) but there are other concerns then about vulnerabilities introduced by bad actors via the action.

We for sure don't wanna use `pull_request_target`. As you say, this comes with tons of other concerns. I guess mainly relying on update lock file, having daily scans on master branch (see comment) should be fine, and manually checking the scan for runs from org PRs (since they fail) is fine

.github/workflows/vulnerability-report.yaml
brynpickering: Ah, yeah. I had hoped it would be a way to force check PRs from forks but you're right, that's not how it would work. I guess the scan does run on fork PRs it just can't post a message to the comment thread. So I could update it to not even try to comment but to also fail if the number of vulnerabilities increases compared to master.

.github/workflows/vulnerability-report.yaml
```diff
@@ -0,0 +1,97 @@
...
+
+    - name: Post vulnerability report comment
+      uses: peter-evans/create-or-update-comment@v4
+      if: steps.find-comment.outputs.comment-id != ''
```
brynpickering:
```suggestion
```

brynpickering:
@lkstrp the vulnerability report check now fails because I've set it to fail on existence of any `critical` vulnerabilities. These come from python v3.13, which we're stuck with until snakemake is able to handle v3.14. I think it's ok to have this check fail whenever there is a critical vulnerability as it can just be checked by the reviewer to see if there's anything suspicious that needs to be addressed.

Could also set the fail bar at another severity level (high? medium?)

.github/workflows/vulnerability-report.yaml
```diff
@@ -0,0 +1,88 @@
...
+        output-format: sarif
+
+    - name: Get SARIF diff
+      run: pixi exec --with sarif-tools sarif diff -o sarif-diff.json master-sarif/*.sarif ${{ steps.vuln-scan-pr.outputs.sarif }}
+
+    - name: Check diff
+      id: check-diff
+      if: github.event.pull_request.head.repo.full_name != github.repository
+      run: |
+        newVuln=${{ fromJson('sarif-diff.json') }}
```
lkstrp: https://docs.github.com/en/actions/reference/workflows-and-actions/expressions#fromjson

I don't think that will work, fromJson requires a json string, not a json file.

.github/workflows/vulnerability-report.yaml
```diff
@@ -0,0 +1,88 @@
...
+        body: |-
+          ## SBOM Vulnerability Report
+
+          Vulnerability scan of `.pixi/envs/default` based on the updated lockfile, compared to vulnerabilities identified on `master` branch.
+
+          <details>
+          <summary>Full scan diff</summary>
+
+          ```json
+          ${{ fromJson('sarif-diff.json') }}
```
lkstrp: https://docs.github.com/en/actions/reference/workflows-and-actions/expressions#fromjson

I don't think that will work, fromJson requires a json string, not a json file.

lkstrp (commented):
> @lkstrp the vulnerability report check now fails because I've set it to fail on existence of any critical vulnerabilities. These come from python v3.13, which we're stuck with until snakemake is able to handle v3.14. I think it's ok to have this check fail whenever there is a critical vulnerability as it can just be checked by the reviewer to see if there's anything suspicious that needs to be addressed.
> Could also set the fail bar at another severity level (high? medium?)

I think the normal default used is high and above, but we can start with critical. Are you sure v3.13 is the reason for the critical vulnerabilities? Those should have been patched?

On a different note, are you sure we can't simplify this? We have currently two workflows, one for a daily security scan on master, uploading an artifact, which is then passed to the second one which checks the diff and reports back as comment. I think GitHub can also handle that diff reporting for us and both workflows can be merged. You just use `github/codeql-action/upload-sarif@v4` again but with a category and set up a branch protection rule. I am not 100% sure if this works, but it would simplify this quite a bit. You would not get the report as a comment, but status checks. Should I just commit what I have in mind or have you checked if GitHub can handle the diff resolution/ reporting back?

.github/workflows/vulnerability-report.yaml
brynpickering: Fair enough! I started with a json string but then switched it to a file for other reasons. Didn't consider the implications properly.

brynpickering:
@lkstrp 

> I think GitHub can also handle that diff reporting for us and both workflows can be merged. You just use github/codeql-action/upload-sarif@v4 again but with a category and set up a branch protection rule. 

Reading [the docs](https://docs.github.com/en/code-security/how-tos/find-and-fix-code-vulnerabilities/integrate-with-existing-tools/uploading-a-sarif-file-to-github#uploading-a-code-scanning-analysis-with-github-actions), I'm not too sure if it would work. `category` looks like it only works on comparing differences within the same commit:

```
category (optional), which assigns a category for results in the SARIF file. This enables you to analyze the same commit in multiple ways and review the results using the code scanning views in GitHub. For example, you can analyze using multiple tools, and in mono-repos, you can analyze different slices of the repository based on the subset of changed files.
```
It could possibly work if you checked out `master` within the PR workflow and then ran the the scan action on that checkout and uploaded that sarif, but perhaps it wouldn't work because by checking out `master` you've changed the commit you're scanning.

> Should I just commit what I have in mind or have you checked if GitHub can handle the diff resolution/ reporting back?

You could also do that, why not! No harm in trying.

lkstrp:
@brynpickering I think it works like a charm:
- Master commits/ daily cron runs/ workflow dispatch: Run [here](https://github.com/lkstrp/pypsa-eur/actions/runs/26823071778/job/79082757396) but don't fail and report to [security tab](https://github.com/lkstrp/pypsa-eur/security/code-scanning)
- PRs with diff which add CVEs [fail](https://github.com/lkstrp/pypsa-eur/pull/37)
- PRs with diff which does not add CVEs [don't fail](https://github.com/lkstrp/pypsa-eur/pull/38)

The problem is fork PRs cannot upload the SARIF, so we don't have access to the diff. Which means we still need the artifacts from master, and can't rely on GitHub Code Scanning here. But it's still possible to still simplify this a bit in a single workflow.

Could you maybe create those PRs in my fork from another fork to test if those work as expected as well? Just use the same branch content

brynpickering:
> Could you maybe create those PRs in my fork from another fork to test if those work as expected as well? Just use the same branch content

Done.

lkstrp:
@brynpickering I reverted your job split, otherwise we need to upload the artifact as well and can't just pass the file. Otherwise I think this is good to go. CI will fail because there is no master baseline yet

brynpickering:
@lkstrp OK, that's fine! If all's good by you, then we can merge this (or wait for a 3rd person to check our combined work :P)

PR #2170, opened 2026-05-18
split build_energy_totals rule into energy and co2

fneum:
**Modularisation:** Split `build_energy_totals` into three rules: `build_co2_totals` (CO2 emissions from EEA and Eurostat, no IDEES dependency) and `build_transformation_output_coke` (coke oven transformation output from Eurostat only) run independently of the remaining :mod:`build_energy_totals` (IDEES-based energy totals, transport data, district heat share, and heating efficiencies).

Verified that outputs remain identical.


fneum:

PR #2173, opened 2026-05-20
Agents/fix availability matrix plotting

euronion:

* The config entry for plotting availability maps for different technologies was not working - fixed. (Default always kicked in, leading to `False` being the only option.)
* Plotting was not available for UA and MD
* Changed the default for plotting by default, because the cost overhead is little

## Checklist

**Required:**

**If applicable:**
- [n/a] For new data sources or versions, [these instructions](https://pypsa-eur.readthedocs.io/en/latest/data_sources.html) have been followed.
- [n/a] New rules are documented in the appropriate `doc/*.rst` files.

### DIFF
--- a/scripts/lib/validation/config/atlite.py
+++ b/scripts/lib/validation/config/atlite.py
@@ -122,8 +122,8 @@ class AtliteConfig(BaseModel):
         description="Whether progressbar for atlite conversion processes should be shown. False saves time.",
     )
     plot_availability_matrix: bool = Field(
-        True,
-        description="Whether to plot the landuse availability matrix.",
+        False,
+        description="Whether to plot the landuse availability matrix. Warning: This requires a significant amount of memory and time and may crash for larger workflows. Use with caution.",
     )
     cutouts: dict[str, _CutoutConfig] = Field(
         default_factory=lambda: {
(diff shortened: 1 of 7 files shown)


scripts/lib/validation/config/atlite.py
```diff
@@ -122,7 +122,7 @@ class AtliteConfig(BaseModel):
         description="Whether progressbar for atlite conversion processes should be shown. False saves time.",
     )
     plot_availability_matrix: bool = Field(
-        False,
+        True,
```
fneum: I would leave the default as `False` because this is a computationally expensive step that not everyone needs. Or can you quantify cost overhead for `config.default.yaml`?. Then I might reconsider.
```suggestion
        False,
```

fneum (changes requested):
Excellent, nice use of `branch()` feature. I would only keep default at `False`.

scripts/lib/validation/config/atlite.py
euronion: Very fair - I wanted to show that the overhead is small, but it turns out it is not and crashed my workflow a couple of times doing so. Reverted to `False` as default :)

fneum (approved):

PR #2178, opened 2026-05-22
fix operational reserve constraints

fneum:
* Operational reserve margin constraints now work for PyPSA v1.0 (dimension name change from "Generator" to "name").
* Load shedding generators are now excluded from reserve margin calculations (the should not).
* Renewable generators are now identified based on the configuration rather than presence of `p_max_pu` data (not a robust selector).
* Fixed for networks without extendable generators (before skipped in such cases).

There are some improvements to be done in the future, such as including links (representing power plants in sector-coupled networks) and storage units (PHS, hydro, batteries) in the operational reserve and generally selecting which carriers can contribute to the reserve. But that's for later.


fneum:

PR #2180, opened 2026-05-26
add highs-hipo solver settings set

fneum:


brynpickering (commented):
Does the current configured version of highs have the HiPO solver available? If not, it may be worth updating the highs version in the `pixi.toml` to:

```toml
highspy = { version = ">=1.13.2.dev0", channel = "conda-forge/label/dev" }
```

fneum:
> Does the current configured version of highs have the HiPO solver available? If not, it may be worth updating the highs version in the `pixi.toml` to:
> 
> ```toml
> highspy = { version = ">=1.13.2.dev0", channel = "conda-forge/label/dev" }
> ```

No, it doesn't. I also played with the version you referenced. I was unsure if it is recommended to use / rely on that version.

PR #2189, opened 2026-06-08
Allow data version CSV / YAML file(s) to be specified in the config

### LINKED ISSUE #2016: Make `data/versions.csv` easily extendable with a configurable option
Discussed with @brynpickering offline:

`data/versions.csv` is currently hard-coded in the `Snakefile`.
This leads to forks needing to modify this file in place or mess with the `Snakefile` or their own modifications and additions.

Following the logic for e.g. `configfiles` in the `Snakefile`, we suggest to provide a list of `data/versions.csv`-like files that overwrite each other's `URL` value for matching `{dataset,version,source}` keys.

Modification to existing `DataConfig`:

```python
class DataConfig(BaseModel):
    """Configuration for `data` settings."""

    version_file: list[str] = Field(
      default_factory=lambda: ["data/versions.csv"],
      description="Files to load data versions from. The files will be combined, with priority given to the files later in the list in case with multiple entries pointing to the same {dataset,version,source} with different URLs"
    )
```

the `config.yaml` would then allow for a field like:

```yaml
version_file:
    - data/versions.csv
    - data/my_versions.csv
    - data/my_forked_fork_versions.csv
```

An additional request came in to convert `data/versions.csv` into a YAML file and consider it a real extension of the configfile. Not sure how I feel about it. The reason for going with a csv file was the easier searching, editing and display on GitHub/locally.

brynpickering:
Closes #2016 

Allows users to completely or partially replace the contents of `data/versions.csv` in the `data.version_files` list. If multiple CSVs are given, they overwrite each other in order they are given.

I've tested the unit tests locally with a partial data file that overrides the base data file. I have successfully run the `base_network` rule with the OSM data retrieval steps.

## Checklist

**Required:**

**If applicable:**


euronion:
The way it is implemented would allow for easy support of `.yaml` files as well, by simply creating a DataFrame from the `yaml` file as well.

Supporting `.yaml` instead of `.csv` was discussed in #2016 - would you be interested in implementing support for it here as well? This would allow forks to decide which file format to use.

brynpickering:
@euronion your wish is my command

euronion:
RTR?

pixi.toml
euronion, line 1: How's that change related to the PR?

scripts/lib/validation/config/data.py
```diff
@@ -31,6 +32,46 @@ class _DataSourceConfig(ConfigModel):
 class DataConfig(BaseModel):
     """Configuration for `data` settings."""
 
+    @field_validator("version_files")
+    @classmethod
+    def check_version_files_are_csv(cls, v: list[FilePath]) -> list[FilePath]:
```
euronion: naming: csv and yaml are being tested for, not only csv

scripts/_helpers.py
```diff
@@ -1088,3 +1088,40 @@ def load_costs(cost_file: str) -> pd.DataFrame:
     """
 
     return pd.read_csv(cost_file, index_col=0)
+
+
+def load_data_versions(file: str, create_cols_from_tags: bool = True) -> pd.DataFrame:
```
euronion:
```suggestion
@lru_cache
def load_data_versions(file: str, create_cols_from_tags: bool = True) -> pd.DataFrame:
```

This is called quite a bit, avoid unnecessary disk access. I'd prefer the caching to happen after `pd.concat(...)` in the calling function, but that would require you to restructure the code a bit. Here should be good enough.

rules/common.smk
```diff
@@ -83,52 +83,44 @@ def config_provider(*keys, default=None):
...
+        data_versions_entry = load_data_versions(path).set_index(
+            ["dataset", "version", "source"]
+        )
+        data_versions_list.append(data_versions_entry)
+    data_versions = pd.concat(data_versions_list)
+    data_versions = (
+        data_versions.loc[~data_versions.index.duplicated(keep="last")]
+        .sort_index()
+        .reset_index()
+    )
```
euronion: These steps were previously cached (see comment also below). Maybe wrap into a dedicated internal function that is cached to avoid dozens of identical file reads + constructions of data frames

rules/common.smk
```diff
@@ -83,52 +83,44 @@ def config_provider(*keys, default=None):
...
     Then the 'data/versions.csv' file is queried to find the matching dataset entry.
 
-    Parameters:
-    name: str
+    Parameters
+    ----------
+    name : str
         The name of the dataset to retrieve version information for.
+    **dataset_config_overrides : str
+        entries to override the dataset config for the given `name`.
```
euronion, line 98: What do we need the overrides for?

rules/common.smk
euronion: Hard-coded file name here no longer correct

scripts/lib/validation/config/data.py
```diff
@@ -31,6 +32,46 @@ class _DataSourceConfig(ConfigModel):
...
+        for path in v:
+            if path.suffix.lower() not in {".csv", ".yaml", ".yml"}:
+                raise ValueError(f"Version file '{path}' must be a CSV or YAML file.")
+            elif not path.exists():
+                raise ValueError(
+                    f"Version file '{path}' must exist and be specified relative to the project root or as an absolute path."
+                )
```
euronion: Isn't this redundant? the `list[FilePath]` in the signature should already cause pydantic to check for the files existing?

euronion (changes requested):
MLGTM! Some minor comments below.

And:  `config` object is still hard coded in `dataset_version` - is the implementation compatible with `config_provider` / config overwrites using the scenario functionality?

rules/common.smk
brynpickering: OSM dataset generation (for upload to zenodo). It was previously using a copy of the method with hardcoded entries for `source` and `version`. I've consolidated it into this one method by adding the kwargs option

pixi.toml
brynpickering: I made updates to the tests and consolidated the use of `--fix`. In the process, it seemed worthwhile to use that `--fix` arg directly in generating the config. It's very much a side note to the PR which I agree isn't actually responding to the main issue. I can spin it out into another PR if you prefer.

brynpickering:
@euronion slight refactor to enable caching (and validating on load, not just in tests - it will cover cases where a user provides their own data file, which we otherwise wouldn't validate prior to merging with the base file).

pixi.toml
euronion: I don't mind it being in this PR. Let's just get @lkstrp to confirm on this particular change, it was his code.

test/test_config_schema.py
```diff
@@ -63,23 +63,27 @@ def _check_file_in_sync(existing_path: Path, generate_func, file_type: str):
         print("=" * 80)
         pytest.fail(
             f"{existing_path} is out of sync with the Pydantic schema. "
-            "Run 'pixi run generate-config' to update. See diff above."
+            "Run `pixi run -e test pytest test/test_config_schema.py --fix` to update. See diff above."
```
euronion: I would keep the old version with `pixi run generate-config`, but fine if you go with the other one.

euronion (changes requested):
I'm not sure how the `yaml` file is supposed to be formed to be valid? How did you test it?

I tried
```yaml
- dataset: aquifer_data
  version: v1.2
  source: archive
  tags: latest supported
  added: 2025-12-02
  note: bl
  url: https://data.pypsa.org/workflows/eur/aquifer_data/v1.2/IHME1500_v12.zip
```

but `snakemake retrieve_aquifer_data_bgr -c1 -n` fails if I use the above entry as `data/versions.yaml` file and add it as follows to the config:
```yaml
data:
  version_files:
  - data/versions.csv
  - data/versions.yaml
 ```

**Comment by brynpickering** [comment]
@euronion that's strange as it works fine when I do exactly the same thing locally:

```sh
Building DAG of jobs...
Job stats:
job                          count
-------------------------  -------
retrieve_aquifer_data_bgr        1
total                            1


[Tue Jun 23 09:11:40 2026]
Retrieving BGR aquifer data
rule retrieve_aquifer_data_bgr:
    input: https://data.pypsa.org/workflows/eur/aquifer_data/v1.2/IHME1500_v12.zip (retrieve from storage)
    output: data/aquifer_data/archive/v1.2/ihme1500_aquif_ec4060_v12_poly.zip, data/aquifer_data/archive/v1.2/IHME1500_v12/shp/ihme1500_aquif_ec4060_v12_poly.shp, data/aquifer_data/archive/v1.2/IHME1500_v12/shp/ihme1500_aquif_ec4060_v12_poly.shx, data/aquifer_data/archive/v1.2/IHME1500_v12/shp/ihme1500_aquif_ec4060_v12_poly.dbf, data/aquifer_data/archive/v1.2/IHME1500_v12/shp/ihme1500_aquif_ec4060_v12_poly.cpg, data/aquifer_data/archive/v1.2/IHME1500_v12/shp/ihme1500_aquif_ec4060_v12_poly.prj, data/aquifer_data/archive/v1.2/IHME1500_v12/shp/ihme1500_aquif_ec4060_v12_poly.sbn, data/aquifer_data/archive/v1.2/IHME1500_v12/shp/ihme1500_aquif_ec4060_v12_poly.sbx
    jobid: 0
    reason: Missing output files: data/aquifer_data/archive/v1.2/IHME1500_v12/shp/ihme1500_aquif_ec4060_v12_poly.sbx, data/aquifer_data/archive/v1.2/IHME1500_v12/shp/ihme1500_aquif_ec4060_v12_poly.cpg, data/aquifer_data/archive/v1.2/ihme1500_aquif_ec4060_v12_poly.zip, data/aquifer_data/archive/v1.2/IHME1500_v12/shp/ihme1500_aquif_ec4060_v12_poly.prj, data/aquifer_data/archive/v1.2/IHME1500_v12/shp/ihme1500_aquif_ec4060_v12_poly.shx, data/aquifer_data/archive/v1.2/IHME1500_v12/shp/ihme1500_aquif_ec4060_v12_poly.sbn, data/aquifer_data/archive/v1.2/IHME1500_v12/shp/ihme1500_aquif_ec4060_v12_poly.shp, data/aquifer_data/archive/v1.2/IHME1500_v12/shp/ihme1500_aquif_ec4060_v12_poly.dbf
    resources: tmpdir=/var/folders/4n/5q9kdtps1334qh_fqx5jl2xw0000gn/T
Job stats:
job                          count
-------------------------  -------
retrieve_aquifer_data_bgr        1
total                            1

Reasons:
    (check individual jobs above for details)
    output files have to be generated:
        retrieve_aquifer_data_bgr
This was a dry-run (flag -n). The order of jobs does not reflect the order of execution.
```

And if I switch out the download link to the one currently set as `primary` in the data/versions.csv, that change makes its way through:

```sh
Building DAG of jobs...
Job stats:
job                          count
-------------------------  -------
retrieve_aquifer_data_bgr        1
total                            1


[Tue Jun 23 09:12:47 2026]
Retrieving BGR aquifer data
rule retrieve_aquifer_data_bgr:
    input: https://download.bgr.de/bgr/grundwasser/IHME1500/v12/shp/IHME1500_v12.zip (retrieve from storage)
    output: data/aquifer_data/archive/v1.2/ihme1500_aquif_ec4060_v12_poly.zip, data/aquifer_data/archive/v1.2/IHME1500_v12/shp/ihme1500_aquif_ec4060_v12_poly.shp, data/aquifer_data/archive/v1.2/IHME1500_v12/shp/ihme1500_aquif_ec4060_v12_poly.shx, data/aquifer_data/archive/v1.2/IHME1500_v12/shp/ihme1500_aquif_ec4060_v12_poly.dbf, data/aquifer_data/archive/v1.2/IHME1500_v12/shp/ihme1500_aquif_ec4060_v12_poly.cpg, data/aquifer_data/archive/v1.2/IHME1500_v12/shp/ihme1500_aquif_ec4060_v12_poly.prj, data/aquifer_data/archive/v1.2/IHME1500_v12/shp/ihme1500_aquif_ec4060_v12_poly.sbn, data/aquifer_data/archive/v1.2/IHME1500_v12/shp/ihme1500_aquif_ec4060_v12_poly.sbx
    jobid: 0
    reason: Missing output files: data/aquifer_data/archive/v1.2/IHME1500_v12/shp/ihme1500_aquif_ec4060_v12_poly.cpg, data/aquifer_data/archive/v1.2/IHME1500_v12/shp/ihme1500_aquif_ec4060_v12_poly.shp, data/aquifer_data/archive/v1.2/IHME1500_v12/shp/ihme1500_aquif_ec4060_v12_poly.dbf, data/aquifer_data/archive/v1.2/ihme1500_aquif_ec4060_v12_poly.zip, data/aquifer_data/archive/v1.2/IHME1500_v12/shp/ihme1500_aquif_ec4060_v12_poly.prj, data/aquifer_data/archive/v1.2/IHME1500_v12/shp/ihme1500_aquif_ec4060_v12_poly.shx, data/aquifer_data/archive/v1.2/IHME1500_v12/shp/ihme1500_aquif_ec4060_v12_poly.sbx, data/aquifer_data/archive/v1.2/IHME1500_v12/shp/ihme1500_aquif_ec4060_v12_poly.sbn
    resources: tmpdir=/var/folders/4n/5q9kdtps1334qh_fqx5jl2xw0000gn/T
Job stats:
job                          count
-------------------------  -------
retrieve_aquifer_data_bgr        1
total                            1

Reasons:
    (check individual jobs above for details)
    output files have to be generated:
        retrieve_aquifer_data_bgr
This was a dry-run (flag -n). The order of jobs does not reflect the order of execution.
```

rules/common.smk
```diff
@@ -83,52 +83,37 @@ def config_provider(*keys, default=None):
...
-    # To use PyPSA-Eur as a snakemake module, the path to the versions.csv file needs to be
-    # registered relative to the current file with Snakemake:
-    fp = workflow.source_path("../data/versions.csv")
-    data_versions = load_data_versions(fp)
+    dataset_config = {**config["data"][name], **dataset_config_overrides}
+
+    data_versions = load_data_versions(
+        *(
+            (
+                Path(workflow.snakefile).parent.parent / path
```
fneum: Why not use the `workflow.source_path()`? I remember there was some reason to do this to work with snakemake modules. Fine if you can confirm this still works.

fneum (commented):
Looks great and useful! Thanks for cleaning up. I had one comment about `workflow.source_path()` and use in snakemake modules.

As for @euronion question:

> is the implementation compatible with config_provider / config overwrites using the scenario functionality?

No, because the `config` is accessed outside a rule definition, so it's not wildcard-aware. (this was the same before).

euronion:
> @euronion that's strange as it works fine when I do exactly the same thing locally:
> [..]

My bad - repo was not clean. Works as expected.
I took the liberty of updating the error message in case the configured version is not found in any of the version files.

rules/common.smk
brynpickering: I can't say I fully understand the difference between them but it seems that `source_path(..)` is very specific for defining `input` paths relative to the current workflow and that it does some [caching magic](https://snakemake.readthedocs.io/en/stable/snakefiles/rules.html#accessing-auxiliary-source-files) so that it doesn't work in this particular context. 

I've generalised it though by pulling in a global variable from the `Snakefile` instead of doing multiple levels of `.parent` to get out of there.

doc/data_sources.md
```diff
@@ -41,37 +41,95 @@ Entries to the `versions.csv` are never deleted and if a dataset was removed or
...
+    url: https://data.pypsa.org/workflows/eur/aquifer_data/v1.2/IHME1500_v12.zip
+```
+
+Priority is given to the definition of a dataset given later in the list.
+Therefore, any datasets sharing the same `[dataset, version, source]` combination will be overwritten in the list.
+Any new entries will be appended to the list.
+
+#### Adding a new version of a dataset
+
+=== "Updating PyPSA/PyPSA-Eur"
```
euronion, line 87: Is the `===` some type of special markdown syntax? Never seen this before

doc/release_notes.md
```diff
@@ -8,6 +8,10 @@
 
 * feat: data version CSV / YAML file can be specified separately or extended by the user in the `data.version_files` config entry ([#2016](https://github.com/PyPSA/pypsa-eur/issues/2016)).
 
+* Fix: Keep unset `p_set` (NaN) as NaN when aggregating components in `cluster_heat_buses`, required for [PyPSA#1703](https://github.com/PyPSA/PyPSA/pull/1703).
+
+* Fix: When clustering heat buses, in myopic optimization, and afterwards viewing the heat energy_balance with n.statistics (and with nice_names=True, which is the default), some assets would still be displayed as belonging to "residential" or "services" sectors, because the nice_names still lingered from the unclustered version. This has been fixed.
+
```
euronion, line 16: Unrelated to PR?

euronion (approved):
PR related things LGTM, two minor comments.

doc/data_sources.md
```diff
@@ -41,37 +41,95 @@ Entries to the `versions.csv` are never deleted and if a dataset was removed or
...
-If you notice that a data source has been updated and want to add the new version to PyPSA-Eur:
+There are two methods for updating the data versions:
 
-1. Add a new row to `data/versions.csv` with the same `dataset` name, the new `version`, `source` set to `primary`, and the `url` pointing to the original data source.
-2. Set appropriate tags (typically `latest supported`).
-3. Update the tags of the previous version (remove `latest`, keep `supported` if still compatible).
-4. Create a pull request with your changes.
-5. Of course, any potential workflow adjustments should be considered and implemented as well.
+1. Add a new row to `data/versions.csv`.
+1. Create a new version file and reference it in the configuration under [`data→version_files`](configuration.md#data_cf).
```
fneum, line 49:
```suggestion
2. Create a new version file and reference it in the configuration under [`data→version_files`](configuration.md#data_cf).
```

fneum (approved):
Thanks for looking into `workflow.source_path()`.

doc/data_sources.md
fneum: It's for creating tabbed content in Mkdocs.

doc/data_sources.md
fneum: Learned something new about markdown regarding automatic numbering.

fneum:
@lkstrp, could you have a quick look at changes to `pixi.toml`? Once confirmed, this is RTM.

pixi.toml
lkstrp: Sorry for the delay! I don't remember why I made that so complicated. I think so this can run without having dev deps installed (or something). But simplicity is king so let's go with dev deps

lkstrp (approved):

PR #2174, opened 2026-05-20
Fix: activate missing losses for "H2 pipeline retrofitted" links

lindnemi:
Bugfix: Activate losses for `H2 pipeline retrofitted` links. Previously these were omitted which caused the model to utilize them much more than the standard `H2 pipeline` links.

## Checklist

**Required:**
- [ ] Changes are tested locally and behave as expected.
- [ ] Code and workflow changes are documented.
- [x] A release note entry is added to `doc/release_notes.rst`.

**If applicable:**
- [ ] Changes in configuration options are reflected in `scripts/lib/validation`.
- [ ] For new data sources or versions, [these instructions](https://pypsa-eur.readthedocs.io/en/latest/data_sources.html) have been followed.
- [ ] New rules are documented in the appropriate `doc/*.rst` files.

PR #2191, opened 2026-06-09
Fix: config link type carriers

nicolezechneraggm:
Bugfix: As introduced for `H2 pipeline retrofitted` in https://github.com/PyPSA/pypsa-eur/pull/2174, the link type carriers `gas pipeline new` and `CO2 pipeline` must be called to correctly address losses. 

If the bidirectional links with (`p_min_pu=-1`) are not added here, they are not split in `lossy_bidirectional_links` in `prepare_sector_network`. They are then allowed negative flows, leading to negative marginal costs in the exporting bus and generating artificial revenue there. 

Additionally, the if-gate in the `lossy_bidirectional_links` function triggers only if `carrier_i.empty`: 
https://github.com/AGGM-AG/pypsa-eur/blob/f750d664f1d5f2afa4c57a3f8b4132ecdd59873f/scripts/prepare_sector_network.py#L5852-L5856
```python 
if (  
    not any((v != 1.0) or (v >= 0) for v in efficiencies.values())  
    or carrier_i.empty  
):  
    return
```
The first part of the condition `(v != 1.0) or (v >= 0)` is always `True`. This may be an oversight that allows the split of bidirectional links regardless of efficiency values, as long as they are given in the config. 

## Checklist

**Required:**

**If applicable:**


fneum:
Thanks, that looks good!

I just would like to skip the split if it's not necessary (e.g. all efficiencies = 1).

Maybe `not any((v != 1.0) or (v >= 0) for v in efficiencies.values())` -> `all(v == 1.0) for v in efficiencies.values())` would be sufficient. Would you implement that?

fneum (approved):

nicolezechneraggm:
Ahoi, 
I can absolutely also fix the if-gate in `prepare_sector_network` too.  
Since I am not the originator of the code, I was unsure if that could cause unintended consequences elsewhere. But since I just checked, and _you_ are the originator, @fneum, I am happy to implement the change.

nicolezechneraggm:
@fneum I added the change to the `lossy_bidirectional_links` function, the PR should be good to go now

nicolezechneraggm:
@fneum Finally managed to fix the tests, so it should be good to go now :)

PR #2199, opened 2026-06-10
don't remove DC buses in sector network

martacki:
Closes # (if applicable).
no issue, but the default configuration (using `config.default.yaml`) is currently failing in `prepare_sector_network` (for the sector-coupeled model)

the latest changes introduced [here](https://github.com/PyPSA/pypsa-eur/commit/8208868e9c91084680227441e93818862f2a7401) lead to an error where during `prepare_sector_network` buses with the carrier `DC` are removed, causing errors later where for example the biomass potentials and other technologies are added.

However, I'm now not sure how to handle the following best: in `prepare_sector_network`, `AC` nodes are bumped to low voltage nodes, `DC` not & are then missing, but are still used for link connections and loads. In this PR as a fast-fix to make the workflow run again I also made them "low voltage", but this might be re-considered. (Didn't asses how much work it would be to leave it at HV so-to-say, but I expect major changes looking at it only briefly)

## Checklist

**Required:**

**If applicable:**
- [x] For new data sources or versions, [these instructions](https://pypsa-eur.readthedocs.io/en/latest/data_sources.html) have been followed.
- [x] New rules are documented in the appropriate `doc/*.md` files.

### DIFF
--- a/.github/workflows/security-scan.yaml
+++ b/.github/workflows/security-scan.yaml
@@ -45,12 +45,15 @@ jobs:
         add-cpes-if-none: true
         output-format: sarif
 
-    # Point alerts to pixi.lock so githubs path scoped PR check registers it
-    - name: Rewrite SARIF locations to pixi.lock
+    # Set uri=pixi.lock for a readable location, and fingerprint each alert by
+    # CVE id so code scanning matches against the master baseline by CVE, not line.
+    - name: Normalize SARIF (pixi.lock location + CVE fingerprint)
       run: |
-        jq '.runs[].results[].locations[].physicalLocation.artifactLocation.uri = "pixi.lock"' \
-          "${{ steps.scan.outputs.sarif }}" > rewritten.sarif
-        mv rewritten.sarif "${{ steps.scan.outputs.sarif }}"
+        jq '.runs[].results[] |= (
+              .locations[].physicalLocation.artifactLocation.uri = "pixi.lock"
+              | .partialFingerprints = {"grypeCve/v1": .ruleId}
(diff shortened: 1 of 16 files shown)


bobbyxng:
@martacki I believe the fix is much simpler, renaming all bus carriers to "AC" after simplification, in the long-term I would suggest to simplify the bus carriers "AC" and "DC" to "electricity" after `simplify_network` (as "AC" before the pull request also referred to both AC and DC/electricity as bus carrier). This does not impact the links still being assigned carrier "DC".

For reference, this was the old line, that I removed a bit over-ambitiously, under the wrong assumption that the downstream sector-coupled workflow would handle the individual bus carriers "AC" and "DC", which is not the case. Existing integration tests unfortunately did not catch the error, sorry for that!

```
# Change carrier type of all added super_nodes to "AC"
n.buses.loc[added_supernodes, "carrier"] = "AC" 
```

I have made a suggestion and tested it succesfully for Europe, 50 nodes and using administrative clustering, NUTS1 level. @fneum

martacki:
> @martacki I believe the fix is much simpler, renaming all bus carriers to "AC" after simplification, in the long-term I would suggest to simplify the bus carriers "AC" and "DC" to "electricity" after `simplify_network` (as "AC" before the pull request also referred to both AC and DC/electricity as bus carrier). This does not impact the links still being assigned carrier "DC".
> 

Not sure this is "much simpler" ;) both works and given that both are simplified networks, there is probably no difference to it in terms of results

martacki:
As I opened the PR I can't approve any changes now @bobbyxng

fneum (approved):

PR #2202, opened 2026-06-12
fix: keep unset p_set as NaN in cluster_heat_buses aggregation

### LINKED ISSUE #2182: cluster_heat_buses silently sets p_set=0.0 on every clustered Generator/Link/Store via NaN-summing aggregation

- [x] I have confirmed this bug exists on the latest release of PyPSA-Eur.
- [x] I have confirmed this bug exists on the current `master` branch of PyPSA-Eur.

### Issue Description

`cluster_heat_buses(n)` in `scripts/prepare_sector_network.py` aggregates the static dataframes of Generator, Link, Store (and others) using:

```python
aggregate_dict = {
    ...
    "p_set": "sum",
    ...
}
df = df.groupby(level=0).agg(aggregate_dict, numeric_only=False)
```

Pandas `Series.sum()` with the default `skipna=True` returns `0.0` for an all-NaN input, not `NaN`. So every clustered Generator/Link/Store ends up with `p_set = 0.0` in `c.static`, even when no source data ever set a `p_set` value on those components.

In PyPSA semantics, `p_set = NaN` means "no setpoint, dispatch is free" while `p_set = 0.0` means "must equal zero." The aggregation rule silently converts the first into the second for every non-Load component flowing through `cluster_heat_buses`.

This was previously masked by an upstream PyPSA bug ([PyPSA #1701](https://github.com/PyPSA/PyPSA/issues/1701), fix in [PyPSA #1703](https://github.com/PyPSA/PyPSA/pull/1703)) where `n.optimize()` silently ignored static `p_set` values. Once #1703 merges, the phantom `0.0`s become enforced constraints, forcing every clustered Generator/Link/Store to zero dispatch and making the sector-coupled `test-sector-overnight` scenario infeasible at HiGHS presolve.

### Reproducible Example

```python
import pandas as pd
import numpy as np

df = pd.DataFrame(
    {"p_set": [np.nan, np.nan, np.nan]},
    index=pd.Index(["g1", "g2", "g3"], name="name"),
)

result = df.groupby(level=0).agg({"p_set": "sum"}, numeric_only=False)
print(result)
#       p_set
# name
# g1      0.0   <- expected NaN; all-NaN group silently summed to 0.0
# g2      0.0
# g3      0.0
```

Inside PyPSA-Eur this happens at `prepare_sector_network.py:5770` for the static dataframe of every component listed at `:5750` (`Bus`, `Carrier`, `Generator`, `Link`, `Load`, `Store`), whenever `options["cluster_heat_buses"]` is enabled.

### Expected Behavior

`p_set` aggregation over an all-NaN group should preserve NaN (semantic: "still unset"), not collapse it to `0.0` (semantic: "forced to zero").

### Suggested fix (open to guidance)

I'm new to this code path, so I'd value a maintainer's view on which direction fits the design intent best. Two options that both restore the correct semantics:

**Option A: drop `p_set` from the aggregation map.** Reads like the simplest move if `p_set` isn't meant to be a real input on Generators/Links/Stores in this code path:

```python
aggregate_dict = {
    ...
    # "p_set": "sum",   <- remove this line
    ...
}
```

**Option B: keep `p_set` in the map but aggregate with `min_count=1`** so NaN-only groups stay NaN. Slightly more defensive if you'd like to keep the map symmetric / future-proof for cases where someone does set `p_set` on a clustered component:

```python
aggregate_dict = {
    ...
    "p_set": lambda s: s.sum(min_count=1),
    ...
}
```

A few clarifying questions I'd appreciate input on:

- Is `p_set` ever expected to be a meaningful input on the non-Load components flowing through `cluster_heat_buses`? If not, Option A seems cleaner.
- If `p_set` could legitimately exist there in some future config, Option B preserves the aggregation path.
- Is there a third approach you'd prefer that I'm missing (for example, a more general "treat all-NaN as NaN" rule applied across the whole aggregation map)?

Happy to open the PR with whichever option you prefer once we agree on the direction.

### Installed Versions

PyPSA-Eur master at `7643299` (latest `[github-actions.ci] Update locked envs`).

lkstrp:
Closes #2182

See https://github.com/PyPSA/PyPSA/pull/1703

### DIFF
--- a/scripts/prepare_sector_network.py
+++ b/scripts/prepare_sector_network.py
@@ -5781,7 +5781,7 @@ def cluster_heat_buses(n):
         # cluster heat nodes
         # static dataframe
         agg = define_clustering(df.columns, aggregate_dict)
-        df = df.groupby(level=0).agg(agg, numeric_only=False)
+        df = df.groupby(level=0).agg(agg)
         # time-varying data
         pnl = c.dynamic
         agg = define_clustering(pd.Index(pnl.keys()), aggregate_dict)
@@ -5790,7 +5790,7 @@ def cluster_heat_buses(n):
             def renamer(s):
                 return s.replace("residential ", "").replace("services ", "")
 
-            pnl[k] = pnl[k].T.groupby(renamer).agg(agg[k], numeric_only=False).T
+            pnl[k] = pnl[k].T.groupby(renamer).agg(agg[k]).T
 
         # remove unclustered assets of service/residential
         to_drop = c.static.index.difference(df.index)
--- a/scripts/prepare_sector_network.py
+++ b/scripts/prepare_sector_network.py
@@ -5695,6 +5695,14 @@ def limit_individual_line_extension(n, maxext):
     n.links.loc[hvdc, "p_nom_max"] = n.links.loc[hvdc, "p_nom"] + maxext
 
 
+def _sum_keep_na(s):
+    """
+    Sum keeping all-NaN groups as NaN instead of collapsing them to 0.
+
+    """
+    return s.sum(min_count=1)
+
+
 aggregate_dict = {
     "p_nom": pd.Series.sum,
     "s_nom": pd.Series.sum,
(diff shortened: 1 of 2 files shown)


scripts/prepare_sector_network.py
```diff
@@ -5695,6 +5695,14 @@ def limit_individual_line_extension(n, maxext):
     n.links.loc[hvdc, "p_nom_max"] = n.links.loc[hvdc, "p_nom"] + maxext
 
 
+def _sum_keep_na(s):
+    """
+    Sum keeping all-NaN groups as NaN instead of collapsing them to 0.
+
+    """
+    return s.sum(min_count=1)
```
FabianHofmann, line 5703: rather in helpers.py?

FabianHofmann (approved):
great! feel free to ignore or tackle the comment. In the long term we should consolidate clustering techniques (but out of scope here)

scripts/prepare_sector_network.py
lkstrp: I guess everywhere else you should rather just pass `s.sum(min_count=1)` directly

fneum (approved):

PR #2211, opened 2026-06-18
fix: resolve plotting crashes due to missing tech colors

### LINKED ISSUE #2108: Missing heat dsm entry in tech_colors causes KeyError in plot_summary.py

- [x] I have confirmed this bug exists on the lastest [release](https://github.com/pypsa/pypsa-eur/releases) of PyPSA-Eur.

### Issue Description

- The `rename_techs()` function strips prefixes (`residential, services, urban, rural, central, decentral`) from technology names, but `tech_colors` only has the original full names - not the aggregated/renamed versions.                                             

- There is no entry for `heat dsm` (the post-rename name), causing the KeyError when `plot_energy()` looks up colors.                                                   

### Reproducible Example

```python
see above
```

### Expected Behavior

see above

### Installed Versions

<details>
Replace this line.
</details>

stan-buren:
Fixes #2108. The plot summary and network mapping scripts crashed when technology color keys were missing from the plotting configuration.

- Add missing 'heat dsm' color entry mapping
- Implement _get_tech_colors() in plot_summary.py to safely lookup colors with warnings
- Apply same safe color lookup and consistency logging to plot_power_network.py and plot_gas_network.py
- Add test suite covering regression and safe fallback lookups

Closes #2108

* **Configuration Fix (`config/plotting.default.yaml`)**: Added the missing mapping `heat dsm: '#ff5c5c'` to `plotting.tech_colors` to prevent crashes when mapping renamed tech categories.
* **Plotting Resilience (`scripts/plot_summary.py`)**: Created a utility `_get_tech_colors` that retrieves colors using `.get()` with a default gray fallback (`#333333`) and triggers warnings when keys are missing. Integrated this helper into cost, energy, and balance plotting functions.
* **Consistent Safe Lookups (`scripts/plot_power_network.py` & `scripts/plot_gas_network.py`)**:
  * Replaced direct index key access on `tech_colors` dicts with safe `.get(..., "#333333")` operations.
  * Added consistency logging in `plot_gas_network.py` warning about missing `"fossil gas"` or `"methanation"` keys in the plotting configuration.
* **Lightweight Testing (`test/test_plot_summary.py`)**: Implemented unit and regression tests confirming color lookup and technology renaming logic. Used a lightweight custom YAML parser to ensure tests remain zero-dependency (allowing `--noconftest` runs without loading heavy GIS/Atlite modules).

## Checklist

**Required:**

**If applicable:**


scripts/plot_gas_network.py
```diff
@@ -121,9 +121,15 @@ def plot_ch4_map(n):
     n.links.bus0 = n.links.bus0.str.replace(" gas", "")
     n.links.bus1 = n.links.bus1.str.replace(" gas", "")
 
+    for key in ["fossil gas", "methanation"]:
+        if key not in tech_colors:
+            logger.warning(
+                f"tech_colors for carrier '{key}' not defined in plotting config."
+            )
```
fneum: Remove. This is more or less what you get from the native KeyError when trying to access non-existent keys. Handling it here adds unnecessary verbosity.

scripts/plot_gas_network.py
```diff
@@ -121,9 +121,15 @@ def plot_ch4_map(n):
...
+
     bus_color = {
-        "fossil gas": tech_colors["fossil gas"],
-        "methanation": tech_colors["methanation"],
+        "fossil gas": tech_colors.get("fossil gas", "#333333"),
+        "methanation": tech_colors.get("methanation", "#333333"),
```
fneum: Please revert. We don't want to work with defaults inside code.

scripts/plot_power_network.py
```diff
@@ -221,7 +221,10 @@ def plot_map(
     )
 
     if with_legend:
-        colors = [tech_colors[c] for c in carriers] + [ac_color, dc_color]
+        colors = [tech_colors.get(c, "#333333") for c in carriers] + [
+            ac_color,
+            dc_color,
+        ]
```
fneum: Please revert. We don't want to work with defaults in code, but the script to fail if keys are not present.

scripts/plot_summary.py
```diff
@@ -61,6 +61,25 @@
...
+    for key in keys:
+        if key in tech_colors:
+            colors.append(tech_colors[key])
+        else:
+            logger.warning(
+                f"tech_colors for carrier '{key}' not defined in plotting config."
+            )
+            colors.append(fallback)
+    return colors
+
```
fneum: I think the check that all indices have defined colors is nice, but please separate it from the color list generation. You can have a `check_tech_colors` function that throws an error when not all colors are defined. Please remove the fallback color handling. As described above, we would want the script to fail in this case.

test/test_plot_summary.py
```diff
@@ -0,0 +1,164 @@
...
+                if indent < 4 and not stripped.startswith("-"):
+                    if ":" in stripped and not stripped.startswith("tech_colors"):
+                        in_tech_colors = False
+                        continue
+                if ":" in stripped:
+                    key, value = stripped.split(":", 1)
+                    key = key.strip()
+                    value = value.strip().strip("'").strip('"')
+                    tech_colors[key] = value
+    return tech_colors
```
fneum: I don't understand why this is necessary. Why not just using yaml.safe_load functionality...

test/test_plot_summary.py
fneum, line 1: Please remove all tests concerned with handling defaults.

fneum (changes requested):
Ok, but we want the scripts to fail if not all colours are defined, not handled by greyscale color.

stan-buren:
Hi @fneum, thanks for the review. 

Makes sense, I reverted all the default fallbacks and warning logs in the network plotting scripts back to the native KeyErrors.

For `plot_summary.py`, I extracted the upfront check into a separate `check_tech_colors` function that raises a KeyError when colors are missing. I also cleaned up the test file to use standard `yaml.safe_load` and removed the default fallback tests.

Let me know if this looks better now!

test/test_plot_summary.py
```diff
@@ -0,0 +1,124 @@
...
+# SPDX-License-Identifier: MIT
+"""
+Tests for scripts/plot_summary.py color matching and tech renaming.
+"""
+
+import pytest
+import yaml
+
+
+def rename_techs(label: str) -> str:
```
fneum: I think it would be better if this is using the original function. Sorry, didn't see that earlier.

test/test_plot_summary.py
```diff
@@ -0,0 +1,124 @@
...
+        if old in label:
+            label = new
+
+    for old, new in rename.items():
+        if old == label:
+            label = new
+    return label
+
+
+def check_tech_colors(tech_colors, keys):
```
fneum: It should use the function you implemented in the rule, not a duplicate of it.

test/test_plot_summary.py
```diff
@@ -0,0 +1,124 @@
...
+    """Standalone copy of check_tech_colors from scripts/plot_summary.py."""
+    missing = [k for k in keys if k not in tech_colors]
+    if missing:
+        raise KeyError(
+            f"The following technology carrier(s) do not have a defined color in the plotting configuration: {missing}"
+        )
+
+
+def test_regression_rename_techs_heat_dsm():
```
fneum: I think this test is fine to remove and already covered by the latter two.

fneum (changes requested):
Almost there, just some finishing touches.

stan-buren:
Thanks @fneum, makes total sense. I reverted the safe fallbacks and warnings in the network scripts, and moved the upfront validation to a standalone `check_tech_colors` function in `plot_summary.py`.

I also cleaned up `test_plot_summary.py` to import `check_tech_colors` directly and deleted the duplicate test cases. The code is indeed much cleaner now.

Let me know if it works.

fneum (approved):
Thanks, now it's great!

PR #2218, opened 2026-06-25
Make `default_target` configurable

brynpickering:
This allows users to leverage the `rulegraph` and `filegraph` rules in their soft forks and to point `pixi run snakemake` to their own "all"-style rule without needing to delete `default_target: True` from the existing `all` rule. 

Technically, it also allows a user to also have different default targets per `run`, so they can call `pixi run snakemake --configfile config/config.<run-with-different-target>.yaml` and it will switch out the default target for that run without the user needing to specify the target in the CLI call. Useful for tests, perhaps? 

## Checklist

**Required:**

**If applicable:**

### DIFF
--- a/Snakefile
+++ b/Snakefile
@@ -59,6 +59,7 @@ resources = path_provider("resources/", RDIR, shared_resources, exclude_from_sha
 scripts = script_path_provider(Path(workflow.snakefile).parent)
 
 RESULTS = "results/" + RDIR
+workflow.default_target = config["run"]["default_target_rule"]
 
 
 localrules:
@@ -104,7 +105,6 @@ if config["foresight"] == "perfect":
 
 
 rule all:
-    default_target: True
     input:
         expand(RESULTS + "graphs/costs.pdf", run=config["run"]["name"]),
         expand(resources("maps/power-network.pdf"), run=config["run"]["name"]),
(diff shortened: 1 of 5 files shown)


euronion:
Interesting feature! So I can also reference rules that are only included later from external `.smk` files, and there are no checks performed on whether the rule exists?
What's the error pattern if the rule doesn't exist?

brynpickering:
> Interesting feature! So I can also reference rules that are only included later from external .smk files, and there are no checks performed on whether the rule exists?
What's the error pattern if the rule doesn't exist?

Asking the right questions. If I put in a rule that doesn't exist, I get a `MissingRuleException`:

```sh
MissingRuleException:
No rule to produce foobar (if you use input functions make sure that they don't raise unexpected exceptions).
```

I could add a custom check at the end of the top-level `Snakefile` to create a more specific error message but I think this might be overkill...

euronion:
A different error message would certainly be more user friendly, albeit maybe overengineered?

Quick and simple: Document the error message that would come out if one uses a rule that does not exist.
More advanced: Your suggestions with a custom check at the end of the Snakefile
Best: Validation with the config file validation (would need to read the Snakefile and parse).

fneum (approved):
Nice mini-feature!

PR #2243, opened 2026-07-28
fix: apply backward fill to country groups

### LINKED ISSUE #2242: Energy totals `fill_missing_years()` fills values cross-country

- [x] I have confirmed this bug exists on the lastest [release](https://github.com/pypsa/pypsa-eur/releases) of PyPSA-Eur.

### Root Cause

The year filling logic in `build_energy_totals.py` applies backward filling to the full dataset instead of country groups: 

https://github.com/PyPSA/pypsa-eur/blob/1fbc646a5a6513017a541d105a35513e71c64b57/scripts/build_energy_totals.py#L506

`.ffill()` is chained after the `gropuby` and applied per country group, but the subsequent `.bfill()` is applied to the joined forward-filling results. This causes `NaN` values to be replaced by the following non-`NaN` value. 

### Effects 

Albania receives the nodal_energy_totals values from Austria 1990. For example, `electricity rail=2.6` TWh in `AL` instead of zero. 

### Suggested Fix

The correct transformation already exists in 

https://github.com/PyPSA/pypsa-eur/blob/1fbc646a5a6513017a541d105a35513e71c64b57/scripts/build_eurostat_balances.py#L62

So the fix becomes 

```Python
fill_values = fill_values.groupby(level="country").transform(lambda x: x.ffill().bfill()) 
```

### Notes

I am not sure if filling with zeros instead of NaN is required. 

### Reproducible Example

```python
import pandas as pd
idx = pd.MultiIndex.from_product(
    [["AL", "AT"], [2019, 2020]], names=["country", "year"]
)
fill_values = pd.Series([np.nan, np.nan, 7.0, 7.0], index=idx)
buggy = fill_values.groupby(level="country").ffill().bfill()  
buggy
Out[3]: 
country  year
AL       2019    7.0
         2020    7.0
AT       2019    7.0
         2020    7.0
dtype: float64

correct = fill_values.groupby(level="country").transform(lambda s: s.ffill().bfill())
correct 
Out[4]: 
country  year
AL       2019    NaN
         2020    NaN
AT       2019    7.0
         2020    7.0
dtype: float64
```

### Expected Behavior

Countries without values have no values after the gap-filling step. 

### Installed Versions

<details>
Replace this line.
</details>

pworschischek-aggmag:
Closes https://github.com/PyPSA/pypsa-eur/issues/2242

Fixes the logic of the gap filling in build energy totals. The backwards filling `.bfill()` method wrongly applied to the full cross-country data set. This caused 1990s values from the next country to fill values for countries without values. 

I found this bug because `electricity rail` values for Albania are very large in PyPSA-AT. Turns out the values are from AT 1990 because that's the next value in the data set. Claude provided the release notes and updated the docs. 

The function now may return NaN in the returned DataFrame object. 

## Checklist

**Required:**
- [x] The description is human-written and any AI-generated content is marked.

**If applicable:**

### DIFF
--- a/doc/release_notes.md
+++ b/doc/release_notes.md
@@ -61,6 +61,8 @@
 
 * chore: Bump the python version to 3.12 ([#2235](https://github.com/PyPSA/pypsa-eur/pull/2235)). This does not increase the real requirement, because 3.12 is already the effective minimum.
 
+* Fix: `fill_missing_years` applied the backward fill across country boundaries, so countries without any Eurostat value silently inherited the values of the next country.
+
 ## PyPSA-Eur v2026.02.0 (18th February 2026)
 
 **Features**
--- a/scripts/build_energy_totals.py
+++ b/scripts/build_energy_totals.py
@@ -486,24 +486,27 @@ def fill_missing_years(fill_values: pd.Series) -> pd.Series:
     ----------
     fill_values : pd.Series
         A pandas Series with a MultiIndex (levels: country and year) representing
-        energy values, where some values may be zero and need to be filled.
+        energy values, where some values may be missing and need to be filled.
 
     Returns
     -------
     pd.Series
-        A pandas Series with zero values replaced by the forward-filled and
+        A pandas Series with missing values replaced by the forward-filled and
         backward-filled values of the corresponding country.
 
     Notes
     -----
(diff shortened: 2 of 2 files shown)


fneum (approved):
Thanks!

PR #2252, opened 2026-08-03
Use geo_boundaries modules as source of spatial shapefiles

brynpickering:
Proof of concept for moving our processes to using [modelblocks](https://www.modelblocks.org/).

I've kept it light touch, leaving the existing `build_shapes` rule to take the consolidated shapefile and explode it out into its respective outputs.

## Comparing results
In the following, I've plotted the current shapes in blue with thin black lines and then overlaid the new shapes with 50% opacity red shapes and black dotted lines. Differences are then evident where the colour changes or there are line mismatches when zooming in.

### NUTS

They are identical for NUTS minus a few river tributaries (e.g. southern NLD / north BEL):

river_overlap

### Non-NUTS
The boundaries are _slightly_ off (e.g. here in BIH):

BIH_mismatch

The geoboundaries open data also comes predominantly from OSM although for BIH it comes [from wikipedia](en.wikipedia.org/wiki/Political_divisions_of_Bosnia_and_Herzegovina), which explains the difference. Kosovo comes from OSM and it has the same region shapes.

### EEZ

Looks the same:
EEZ

## Handling modules
Since this is the first PR using a modelblock, I have made some opinionated decisions on how to handle them:

1. The configs exist separately, each referencing the their own schema in the file. Overriding the default config requires defining items under the `config` key for the module in the module section of the base config. 
  - Maybe this should be called `overrides`?
  - thoughts on how to handle this more gracefully @irmcodebase?
2. The rules exist in a separate module subdirectory. Since it's only one rule per module, it could all sit in a single `modules.smk` file, I'm just unsure how long that file will get if we add lots of modules.
3. I've gone lightest touch on running the module, by letting it get shapes for all countries even when only requesting a subset in the config. 
4. requires using `--use-conda` in the snakemake call to ensure the `geo_boundaries` conda environment is installed. This avoids needing to add their deps to our own.
   - there are plans to make pixi envs possible here which might make things a bit smoother, with a common dependency cache
5. Since the module config can be overridden per pypsa-eur scenario, I've left the `resources` pathvar as `resources`. However, it is mostly download data, which would be better off having in `data`. Currently, we can't split this as two separate pathvars, but I've [upstreamed the request](https://github.com/modelblocks-org/module_geo_boundaries/issues/81).

## Geo-boundaries design decisions

- Pull in all countries and then only filter on `build_shapes`. This add bloat but could only be avoided by doing some config magic to filter out keys under the `geo_boundaries` `scenarios` config key before passing the config over to the module.
- Use geoBoundaries for non-nuts data
- Do not split EEZs using the module's voronoi method.

## Checklist

**Required:**
- [ ] The description is human-written and any AI-generated content is marked.

**If applicable:**


config/modules/geo_boundaries.yaml
```diff
@@ -0,0 +1,57 @@
...
+      BGR: *nuts3
+      CHE: *nuts3
+      CZE: *nuts3
+      DEU: *nuts3
+      DNK: *nuts3
+      EST: *nuts3
+      ESP: *nuts3
+      FIN: *nuts3
+      FRA: *nuts3
+      GBR: *nuts3
```
irm-codebase, line 31: Quick comment on `GBR`. This one will fail if your `nuts` year is > 2016 due to Brexit.

Any year before that will work just fine! I recommend 2016.

rules/modules/geo_boundaries.smk
```diff
@@ -0,0 +1,24 @@
...
+        resources=resources(MODULE_NAME),
+        results=resources(f"{MODULE_NAME}/results"),
+    snakefile:
+        github(
+            "modelblocks-org/module_geo_boundaries",
+            path="workflow/Snakefile",
+            tag=config["modules"][MODULE_NAME]["version"],
+        )
+    config:
+        config["modules"][MODULE_NAME]["config"]
```
irm-codebase: As per the discussion in https://github.com/modelblocks-org/module_geo_boundaries/issues/81

Perhaps PyPSA could create a new directive (`modules(MODULE_NAME)`) that places `logs`, `resources` and `results` in a `modules/{MODULE_NAME}` directory that is invariant per PyPSA scenario name, while keeping `shapes=resources(...)` variant?

This would give you the best of both worlds: stable modules that produce re-usable files across scenarios, and module outputs that are traceable per scenario run.

```suggestion
module geo_boundaries:
    pathvars:
        shapes=resources("shapes/{scenario}.parquet"),
        logs=modules(f"MODULE_NAME/logs"),
        resources=modules(MODULE_NAME),
        results=modules(f"{MODULE_NAME}/results"),
    snakefile:
        github(
            "modelblocks-org/module_geo_boundaries",
            path="workflow/Snakefile",
            tag=config["modules"][MODULE_NAME]["version"],
        )
    config:
        config["modules"][MODULE_NAME]["config"]
```

brynpickering: This would mean that you would never be able to edit the module configuration per run, only add module-level scenarios that you could then request per run. This seems reasonable enough although it requires editing two files to make a run-level modification (create a copy / edit the default module config + edit the run config to point to the new scenario name). How would this work with other modules? Do they all have the concept of `scenarios` for the same purpose?

config/modules/geo_boundaries.yaml
brynpickering: I don't have any issue getting it for 2021. Quick google suggests it is fine in 2021 but not afterwards.

rules/modules/geo_boundaries.smk
irm-codebase: Depends on the case and data at hand! But it's not as complicated as it sounds.

Generally:
- `scenarios` from geo-boundaries become a `shapes` wildcard in subsequent modules.
Assuming the module only has a 'generic' configuration, you won't need to reconfigure it for different shapes. Unless you want to alter behaviour, in which case you just modify the 'generic' configuration that PyPSA-Eur passes to it.
- Some modules have 'scenarios' that fit their scope.
But these are normally applied independently of the `shapes` provided.
  - E.g., [CO2Stop](https://www.modelblocks.org/modules/module_co2stop_cdr/) has configurations settings for `cdr_group` (aquifer, gas, oil) and `scenario` (low, medium, high). Outputs will combine all 3: `<results>/{shapes}/{scenario}/{cdr_group}.parquet`. Think of it as 'applying the `low` scenario for `aquifers` for different `shapes` (LTU, e-highways, EU27, etc).
- Module configuration settings that are 'shape' specific are basically non-existent, so far. The reason is that the user can just handle that on their side with minimal logic.

rules/common.smk
```diff
@@ -208,3 +208,8 @@ def input_cutout(wildcards, cutout_names="default"):
         return [f"{cutouts_path}/{cn}.nc" for cn in cutout_names]
     else:
         return f"{cutouts_path}/{cutout_names}.nc"
+
+
+def module_config(module: str) -> dict:
+    """Get module config from path"""
+    return yaml.safe_load(Path(config["modules"][module]["config_path"]).read_text())
```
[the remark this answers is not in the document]
brynpickering: The import must come from the top-level Snakefile as it's referenced elsewhere in this .smk file (not necessarily a good thing, of course!)

coroa:
I guess there is no good way to check the config schema for the modules already during the initial validate_config sweep?

irm-codebase:
@coroa the module itself always runs that validation when called, so it should catch configuration problems.
`validate_config` is run within the module at startup.
It might not be necessary to add a schema for it on your side, unless you want to validate how that configuration will be "re arranged" by your code... Seems like repeated effort though.

PR #2261, opened 2026-08-11
ci: fix false positives in PR security gate

lkstrp:
The PR gate diffed the PR scan against a SARIF that master produced days earlier. The DB updates daily, so any CVE published in between showed up as new and then broke in PRs. For example, here [attempt 1](https://github.com/PyPSA/pypsa-eur/actions/runs/31472812306/attempts/1) fails, while after a manual rerun on master it succeeds in [attempt 2](https://github.com/PyPSA/pypsa-eur/actions/runs/31472812306).

Now the full master's SBOM is stored as the baseline instead of its scan result (SARIF). PR runs now scan both their own env and the baseline with the same vuln DB and the diff only contains CVEs the PR actually introduces.


.github/workflows/security-scan.yaml
```diff
@@ -95,21 +96,33 @@ jobs:
...
+      id: baseline
+      if: github.event_name == 'pull_request'
+      uses: anchore/scan-action@v7.4.0
+      with:
+        sbom: master-baseline/sbom.spdx.json
+        fail-build: false
+        severity-cutoff: high
+        add-cpes-if-none: true
+        by-cve: true
+        output-format: sarif
```
brynpickering: We probably want all these settings to be the same for the master baseline and for the PR-specific run. You could use a yaml anchor from the previous definition of settings to re-apply them here?

brynpickering (commented):
Makes sense - nice spot!

fneum (approved):

